PR #54, opened 2022-03-29
Redux, thunk, material UI

kimberlythegeek:


kimberlythegeek:
I kept the name for the test-utils file, as it is not a component and should be confused as such. I got this example from https://redux.js.org/usage/writing-tests#components

Issue #68, opened 2022-04-04
add search by short hash


kimberlythegeek:
treeherder API already supports this: https://treeherder.mozilla.org/api/project/try/push/?full=true&count=10&revision=815faae9b5fe

I'm unsure what the results are if multiple revisions match (or the likelihood that this occurs), but I will find out

kimberlythegeek:
fixed by #84

PR #70, opened 2022-04-05
reject thunks that return no results

kimberlythegeek:
the api call to fetch recent revisions or search results is called whenever a value is entered that matches a search hash or email. 

so, if typing "ksereduck@mozilla.com", the api call will be made at `ksereduck@mozilla.c` `ksereduck@mozilla.co`, and `ksereduck@mozilla.com`

I found that the results for the complete email would appear, then disappear, because one of the previous API calls completed after the valid one, and would overwrite the search results with an empty array

### DIFF
--- a/src/reducers/SearchSlice.jsx
+++ b/src/reducers/SearchSlice.jsx
@@ -7,20 +7,21 @@ import {
 } from '../thunks/searchThunk';
 
 const initialState = {
-  recentRevisions: [],
   repository: '',
   searchResults: [],
   searchValue: '',
-  loading: 'idle',
+  errorMessage: null,
 };
 
 const search = createSlice({
   name: 'search',
   initialState,
   reducers: {
-    updateLoadingState(state, action) {
-      state.loading = action.payload;
+    // BEGIN used for testing only
+    updateErrorMessage(state, action) {
+      state.errorMessage = action.payload;
     },
+    // END used for testing only
     updateSearchValue(state, action) {
       state.searchValue = action.payload;
     },
@@ -34,43 +35,31 @@ const search = createSlice({
   extraReducers: (builder) => {
     builder
       // fetchRecentRevisions
-      .addCase(fetchRecentRevisions.pending, (state) => {
-        state.loading = 'pending';
-      })
       .addCase(fetchRecentRevisions.fulfilled, (state, action) => {
-        state.loading = 'succeeded';
-        state.recentRevisions = action.payload;
+        state.searchResults = action.payload;
       })
-      .addCase(fetchRecentRevisions.rejected, (state) => {
-        state.loading = 'failed';
+      .addCase(fetchRecentRevisions.rejected, (state, action) => {
+        state.errorMessage = action.payload;
       })
       // fetchRevisionByID
-      .addCase(fetchRevisionByID.pending, (state) => {
-        state.loading = 'pending';
-      })
       .addCase(fetchRevisionByID.fulfilled, (state, action) => {
-        state.loading = 'succeeded';
         state.searchResults = action.payload;
       })
-      .addCase(fetchRevisionByID.rejected, (state) => {
-        state.loading = 'failed';
+      .addCase(fetchRevisionByID.rejected, (state, action) => {
+        state.errorMessage = action.payload;
       })
       // fetchRevisionsByAuthor
-      .addCase(fetchRevisionsByAuthor.pending, (state) => {
-        state.loading = 'pending';
-      })
       .addCase(fetchRevisionsByAuthor.fulfilled, (state, action) => {
-        state.loading = 'succeeded';
         state.searchResults = action.payload;
       })
-      .addCase(fetchRevisionsByAuthor.rejected, (state) => {
-        state.loading = 'failed';
+      .addCase(fetchRevisionsByAuthor.rejected, (state, action) => {
+        state.errorMessage = action.payload;
       });
   },
 });
 
 export const {
-  updateLoadingState,
+  updateErrorMessage,
   updateSearchValue,
   updateSearchResults,
   updateRepository,
--- a/src/tests/searchThunk.test.js
+++ b/src/tests/searchThunk.test.js
@@ -1,32 +1,9 @@
 import store from '../common/store';
 import {
-  updateLoadingState,
-  updateSearchValue,
-  updateSearchResults,
-  updateRepository,
-} from '../reducers/SearchSlice';
-import {
   fetchRecentRevisions,
   fetchRevisionByID,
   fetchRevisionsByAuthor,
 } from '../thunks/searchThunk';
-
-const unmockedFetch = global.fetch;
-
-beforeEach(() => {
-  global.fetch = jest.fn();
-});
-
-afterEach(() => {
-  global.fetch = unmockedFetch;
-  store.dispatch(updateLoadingState('idle'));
-  store.dispatch(updateSearchValue(''));
-  store.dispatch(updateSearchResults([]));
-  store.dispatch(updateRepository(''));
-  jest.clearAllMocks();
-  jest.resetAllMocks();
-  jest.restoreAllMocks();
-});
 
 describe('fetchRecentRevisions', () => {
   it('should call fetch with correct URL and update recent revisions', async () => {
@@ -40,9 +17,6 @@ describe('fetchRecentRevisions', () => {
     const spyOnDispatch = jest.spyOn(store, 'dispatch');
     const repository = 'brian';
 
-    // Ensure state has been reset in between tests
-    expect(store.getState().search.loading).toBe('idle');
-
     await store.dispatch(fetchRecentRevisions(repository));
 
     expect(spyOnDispatch).toHaveBeenCalled();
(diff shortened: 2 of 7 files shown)


src/reducers/SearchSlice.jsx
```diff
@@ -12,12 +12,20 @@ const initialState = {
   searchResults: [],
   searchValue: '',
   loading: 'idle',
```
julienw, line 14: It would be good to document the expected values and their meaning. In the profiler we're using a type system to help us with that.

It would be good to document which transitions are expected too (eg: idle -> pending, etc).

src/thunks/searchThunk.js
```diff
@@ -4,51 +4,49 @@ const treeherderBaseURL = 'https://treeherder.mozilla.org';
...
+  async (repository, { getState, requestId }) => {
+    const { currentRequestID, loading } = getState().search;
+    if (loading !== 'pending' || requestId !== currentRequestID) {
+      return [];
     }
+    const response = await fetch(
+      `${treeherderBaseURL}/api/project/${repository}/push/`,
+    )
+      .then((res) => res.json())
+      .then((res) => res.results);
```
julienw, line 16: Style comment:
I'd wait this more like this:
```js
const res = await fetch(...)
const json = await res.json();
return res.results;
```

Indeed in a code such as this one, I usually find it more readable to linearize it this way. This is fully equivalent so it's really a style preference.

It may still be necessary to use the Promise API directly when running concurrent tasks, but that's not that common.

src/thunks/searchThunk.js
```diff
@@ -4,51 +4,49 @@ const treeherderBaseURL = 'https://treeherder.mozilla.org';
...
-        .then((res) => res.results);
-    } catch (err) {
-      return rejectWithValue(err.response.data);
+  async ({ repository, search }, { getState, requestId }) => {
+    const { currentRequestID, loading } = getState().search;
+    if (loading !== 'pending' || requestId !== currentRequestID) {
+      return [];
     }
+    const response = await fetch(
+      `${treeherderBaseURL}/api/project/${repository}/push/?revision=${search}`,
```
julienw, line 29: (for another patch)

I think you should either use `encodeURIComponent` or the `URLSearchParams` object to make sure this `search` value isn't malicious.

Also not sure where `treeherderBaseURL` and `repository` come from, but you might need to use `encodeURIComponent` for them too, if they come from some untrusted place.

These are probably not a very big deal in this case because this is just a `get` operation, but better do it to be safe :)

Obviously the same comment applies to the other fetch commands.

src/tests/searchThunk.test.js
```diff
@@ -60,11 +65,22 @@ describe('fetchRecentRevisions', () => {
...
+    expect(store.getState().search.loading).toBe('pending');
+    expect(store.getState().search.currentRequestID).not.toBe(undefined);
+  });
+
+  it('should update state with error message when fetch returns an error', async () => {
     const spyOnFetch = jest
       .spyOn(global, 'fetch')
       .mockImplementation(() =>
-        Promise.reject(new Error('failed to fetch revisions')),
+        Promise.resolve(Promise.reject(new Error('What, ridden on a horse?'))),
```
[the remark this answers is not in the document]
kimberlythegeek: I had some issues getting it to work, so I tried many things to be honest :)

I'll try removing this and see if tests still work as expected

src/reducers/SearchSlice.jsx
```diff
@@ -12,12 +12,20 @@ const initialState = {
   searchResults: [],
```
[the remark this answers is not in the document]
kimberlythegeek: ah, so it is :) thanks for catching this

src/thunks/searchThunk.js
kimberlythegeek: I'm guessing you meant (in the last line) `return json.results;`?

src/thunks/searchThunk.js
kimberlythegeek: the URL is a `const` hardcoded in this same file, the `repository` is coming from the store. it is set when a user selects a repository from the dropdown

src/reducers/SearchSlice.jsx
julienw: This is an old component with very few comments here ahah.

Here is a simple example: https://github.com/firefox-devtools/profiler/blob/95d40cb654241bd0a6bc1b38dca8def5338b1a2a/src/types/actions.js#L131-L138

But the meaning might be obvious in this case so it's really up to you :) Try to put yourself in your own brain reading this in 6 months, and add the information you think could be useful to the future you. You know better than me :-)

src/thunks/searchThunk.js
julienw: Yes! oops :-)

src/thunks/searchThunk.js
julienw: I think it should be done for the `search` at least. This might even bring bugs depending on the characters in the search, such as +, %, or anything that has a special meaning in a URL.

For `repository` my gut feeling is that if you don't ensure that the value is some specific hardcoded value (with an allowlist for example) it should also be escaped but it's likely less important.

src/reducers/SearchSlice.jsx
```diff
@@ -12,12 +12,20 @@ const initialState = {
...
+  error: null,
 };
 
 const search = createSlice({
   name: 'search',
   initialState,
   reducers: {
+    updateCurrentRequestID(state, action) {
+      state.currentRequestID = action.payload;
+    },
```
julienw, line 25: My understanding is that this is used in tests only. At least you should state this in a comment, or possibly in the name of the action itself (eg `updateCurrentRequestIDTestOnly`). The best would be to avoid using test-only functions, but sometimes we can't avoid it :-)

Here is a general guiding principle we try to follow in the profiler: the tests should look like what the users would do.
Here are some more words about that: https://kentcdodds.com/blog/write-tests https://testing-library.com/docs/guiding-principles
This means less unit tests and more integration tests (that is: components with connected redux, and simulating user clicks).
This means stop testing things that can't happen in the real application.
This also means less test-specific code. It's not always possible to avoid that but this is what we're trying to do.
This approach has been quite successful for us, that's why I'm sharing it.

Hope this helps :)

src/thunks/searchThunk.js
kimberlythegeek: currently the `fetch` will only occur if the search value matches either an email or 40 character hash (alphanumeric basically) which I will be extending to include a short hash (12 character alphanumeric string), since I now know that the treeherder API accepts short hashes (#68)

https://github.com/mozilla/perfcompare/blob/master/src/utils/searchViewHelper.js#L14-L23

src/reducers/SearchSlice.jsx
kimberlythegeek: thanks! I appreciate all the feedback, this should be very helpful

src/reducers/SearchSlice.jsx
kimberlythegeek: filed #75

kimberlythegeek:
Filed  #74 #75 #76 based on @julienw 's feedback to address in other patches

kimberlythegeek:
the work here is based on https://redux-toolkit.js.org/api/createAsyncThunk#examples

it says it will handle only one request at a time, but, like Julien, I'm unsure that it truly behaves that way. It appeared to be working as expected, but when running again today (before adding the timeout) I was again having the issue where a rejected promise (no results, incomplete email) was resolving _after_ a fulfilled promise (full email, with results)

the timeout appears to solve this problem. I'm unsure why it doesn't appear to work as expected, but I've followed the examples in the redux-toolkit docs. With sufficient test coverage this shouldn't be an issue. Let me know if you have thoughts/questions on this.

kimberlythegeek:
I was still experiencing issues when using a delay time. There was an edge case that still resulted in a promise with no results resolving after a promise _with_ results, and updating state.search.searchResults to an empty array

I've changed this to use `rejectWithValue`: https://redux-toolkit.js.org/api/createAsyncThunk#handling-thunk-errors

if a promise returns no results, i.e. `ksereduck@mozilla.co`, it rejects with an error message "No results found", and does _not_ update state

If a promise with results completes before a promise with no results, 1) results are added 2) state is not changed when the other promise rejects

@julienw I had no luck when trying to use `abort()`. this requires keeping track of which promise I want to resolve which I want to abort. I found that even when an incomplete email was sent first (or should have been), the complete email always displayed first in my log. So when trying to `abort` any previous call, the invalid request was still the last to appear and resolve (yay asynchronous coding)

however, this also means that, if a user types an email, gets results, and then begins to backspace, the results will _not_ disappear because: 1) the promise returns no results 2) a promise with no results is rejected and does not update state.

The results are always cleared when the search input is cleared, so the results would disappear if the user deleted everything in the input.

there are now two "error" properties in state, I left comments explaining the difference. 

If a promise does reject, it return an error _object_, and the entire object is included in `error`

If a promise resolves with no results, `rejectWithValue` is called with a message, "No results found", and this is the value of `errorMessage` in state.

we will need to check for both of these values when we implement error handling. we can `mapStateToProps` for both of these, so that when either changes value, we can display a relevant error message.

kimberlythegeek:
I forgot when editing this that `createAsyncThunk` will always resolve a promise, so I need to add a try/catch and `resolveWithError`

kimberlythegeek:
I forgot when editing this that `createAsyncThunk` will always resolve a promise, so I need to add a try/catch and `resolveWithError`

julienw:
Removing my review request to clean my review queue before the week-end, as I think we discussed all this on Matrix already, but ping me again whenever you want!

PR #73, opened 2022-04-06
Add router to perf compare

eng-esther:
Add react router to perf compare

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,6 +28,25 @@ module.exports = {
       'error',
       { props: true, ignorePropertyModificationsFor: ['state'] },
     ],
+    'import/order': [
+      'error',
+      {
+        groups: ['builtin', 'external', 'internal'],
+        pathGroups: [
+          {
+            pattern: 'react',
+            group: 'external',
+            position: 'before',
+          },
+        ],
+        pathGroupsExcludedImportTypes: ['react'],
+        'newlines-between': 'always',
+        alphabetize: {
+          order: 'asc',
+          caseInsensitive: true,
+        },
+      },
+    ],
   },
   overrides: [
     {
--- a/src/components/App.jsx
+++ b/src/components/App.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import CssBaseline from '@mui/material/CssBaseline';
+import { BrowserRouter as Router, Route, Routes } from 'react-router-dom';
 
 import SearchView from './Search/SearchView';
 
@@ -8,7 +9,12 @@ function App() {
   return (
     <React.Fragment key="App Component">
       <CssBaseline />
-      <SearchView />
+      <Router>
+        <Routes>
+          <Route path="/" element={<SearchView />} exact />
+          {/*  Your routes go here .. */}
+        </Routes>
+      </Router>
     </React.Fragment>
   );
 }
--- a/src/components/Search/SearchView.jsx
+++ b/src/components/Search/SearchView.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+
 import Box from '@mui/material/Box';
 import Container from '@mui/material/Container';
 import Typography from '@mui/material/Typography';
--- a/src/tests/App.test.jsx
+++ b/src/tests/App.test.jsx
@@ -1,0 +1,19 @@
+/**
+ * @jest-environment jsdom
+ */
+import React from 'react';
+
+import { render, screen } from '@testing-library/react';
+
+import App from '../components/App';
+import * as SearchView from '../components/Search/SearchView';
+
+test('Should render search view on default route', () => {
+  jest
+    .spyOn(SearchView, 'default')
+    .mockImplementation(() => <div>SearchViewMock</div>);
+
+  render(<App />);
+
+  expect(screen.getByText('SearchViewMock')).toBeInTheDocument();
+});
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "react": "^17.0.2",
         "react-dom": "^17.0.2",
         "react-redux": "^7.2.6",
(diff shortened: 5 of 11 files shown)


src/components/App.jsx
```diff
@@ -1,14 +1,18 @@
-import React from 'react';
-
 import CssBaseline from '@mui/material/CssBaseline';
-
+import React from 'react';
```
kimberlythegeek, line 2: This should be at the top. I don't know why the linter didn't catch this, it errors on this in the `App.test.jsx` file

src/components/App.jsx
```diff
@@ -1,14 +1,18 @@
...
+import { BrowserRouter as Router, Route, Routes } from 'react-router-dom';
```
kimberlythegeek, line 3: could you also move this above CssBaseline? I prefer to order imports by hierarchy, React/Redux modules first, then other external modules, then project imports

src/components/Search/SearchView.jsx
```diff
@@ -1,8 +1,7 @@
-import React from 'react';
 import Box from '@mui/material/Box';
 import Container from '@mui/material/Container';
 import Typography from '@mui/material/Typography';
-
+import React from 'react';
```
kimberlythegeek, line 4: same here. it's weird that it's not catching that

src/tests/App.test.jsx
```diff
@@ -0,0 +1,24 @@
+/**
+ * @jest-environment jsdom
+ */
+import { MemoryRouter } from 'react-router';
+import { render, screen } from '@testing-library/react';
+import * as SearchView from '../components/Search/SearchView';
+import App from '../components/App';
+import React from 'react';
+
+beforeEach(() => {
```
kimberlythegeek, line 10: We can remove this as it isn't used in the test

kimberlythegeek (changes requested):
There are linting errors that need to be resolved.

If I run `npm start`, it appears the default url is `localhost:3000/perfcompare` (left over from the create-react-app generation), so I don't see anything displayed. the default should be changed to `/`

After taking a look at this patch, I think that we should separate the header into a reusable component (probably rendered inside `App.jsx`) and then rendering `SearchView` or `CompareView`, etc, using nested routes (see: https://reactrouter.com/docs/en/v6/getting-started/overview#nested-routes)

this could be done in a separate patch, though. thoughts @beatrice-acasandrei @esanuandra ?

src/components/App.jsx
eng-esther: I ran `npm run lint:fix`

eng-esther:
> There are linting errors that need to be resolved.
> 
> If I run `npm start`, it appears the default url is `localhost:3000/perfcompare` (left over from the create-react-app generation), so I don't see anything displayed. the default should be changed to `/`
> 
> After taking a look at this patch, I think that we should separate the header into a reusable component (probably rendered inside `App.jsx`) and then rendering `SearchView` or `CompareView`, etc, using nested routes (see: https://reactrouter.com/docs/en/v6/getting-started/overview#nested-routes)
> 
> this could be done in a separate patch, though. thoughts @beatrice-acasandrei @esanuandra ?

Yeah definitely out of scope for this. ( in terms of the re-usable component )
the router renders the component based on the route change... the header can still be in the app, and can be moved into it's own component.

src/components/App.jsx
eng-esther: probably need to review the linter and the rules..

kimberlythegeek: Yeah I check the rules and tried removing the edits I made, it still didn't catch this. It throws an error for the test file but not this one.

src/components/App.jsx
kimberlythegeek: Not if the linting isn't working as expected. No clear reason why it isn't. You can manually change this and we can create a bug to investigate issues with the linter

esanuandra (approved):

src/components/App.jsx
kimberlythegeek: please move this import

src/components/Search/SearchView.jsx
kimberlythegeek: and this one

beatrice-acasandrei (approved):

.eslintrc.js
```diff
@@ -37,6 +37,25 @@ module.exports = {
         'import/no-extraneous-dependencies': 'off',
         'function-paren-newline': 'off',
         'implicit-arrow-linebreak': 'off',
+        'import/order': [
```
kimberlythegeek, line 40: Nice--thanks for this!

However this should be moved inside the `rules` section. This overrides section only applies to test files.

When I move this to `rules`, I get import errors in some other files also, so feel free to add this rule in a separate patch and fix those errors. (Also I learned that it groups `react-router-dom` with `external` rather than `builtin`, so it says it should come after `CssBaseline` due to alphabetic ordering.)

kimberlythegeek (changes requested):
The commit message should be modified to reflect the changes that occured, it still says "WIP, tests missing"

also the rule you added isn't being applied, it's up to you if you want to add that in a different patch. If you do, please file a corresponding issue.

.eslintrc.js
eng-esther: maybe something is wrong with my setup, cos I dont get any issues when I run `npm run lint:fix` locally, all the errors show on ci level.

eng-esther:
> The commit message should be modified to reflect the changes that occured, it still says "WIP, tests missing"
> 
> also the rule you added isn't being applied, it's up to you if you want to add that in a different patch. If you do, please file a corresponding issue.

done

kimberlythegeek:
> Thanks! Please squash + update the commit message prior to merging

ah, looks like you beat me to it :)

PR #78, opened 2022-04-13
update tests to be more realistic

### LINKED ISSUE #76: write tests to be more realistic
This includes wrapping `dispatch` calls in `act()`, but also avoiding calling dispatch if possible. The tests should use `user-event` to interact with the app the way a user would

kimberlythegeek:
fixes #76

replaced thunk and helper tests with integration tests

instead of directly dispatching actions, these tests interact with the app the way a user would


kimberlythegeek:
blocked by #70

PR #83, opened 2022-04-15
Improve search results UI and hide results when clicking outside input

kimberlythegeek:


kimberlythegeek:
#7

.eslintrc.js
```diff
@@ -60,9 +60,10 @@ module.exports = {
   },
   overrides: [
     {
-      files: ['*test*', 'setupTests.js'],
+      files: ['src/tests/**/*'],
       rules: {
         'react/prop-types': 'off',
+        'import/no-import-module-exports': 'off',
```
kimberlythegeek, line 66: this was failing linting for the added `globalSetup.js` file, which I am using to set the timezone. This ensures snapshots are rendered with the same time zone every time and don't fail tests, since this patch adds date and time to the search results

package.json
```diff
@@ -79,7 +79,8 @@
   },
   "jest": {
     "setupFilesAfterEnv": [
-      "./src/tests/setupTests.js"
-    ]
+      "./src/tests/utils/setupTests.js"
+    ],
+    "globalSetup": "./src/tests/utils/globalSetup.js"
```
kimberlythegeek, line 84: set timezone for tests

src/components/Search/SearchInput.jsx
```diff
@@ -12,47 +12,61 @@ import { connect } from 'react-redux';
...
-  );
+class SearchInput extends Component {
+  componentDidMount() {
+    document.addEventListener('mousedown', handleClickOutsideInput);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('mousedown', handleClickOutsideInput);
+  }
+
```
kimberlythegeek, line 28: had to add an event listener when component mounts to detect when user clicks outside of input or results

src/components/Search/SearchInput.jsx
```diff
@@ -12,47 +12,61 @@ import { connect } from 'react-redux';
...
+  render() {
+    const { repository, searchIsFocused, searchResults } = this.props;
+    return (
+      <Container maxWidth="md">
+        <FormControl sx={{ width: '25%' }}>
+          <InputLabel id="select-repository">repository</InputLabel>
```
kimberlythegeek, line 34: added ID for detecting when this element is clicked (or not)

src/components/Search/SearchInput.jsx
```diff
@@ -12,47 +12,61 @@ import { connect } from 'react-redux';
...
+                  key={item}
+                  onClick={handleChangeDropdown}
+                >
+                  {item}
+                </MenuItem>
+              ))}
+          </Select>
+        </FormControl>
+        <TextField
+          id="search-revision-input"
```
kimberlythegeek, line 54: added ID for detecting when this element is clicked (or not)

src/components/Search/SearchResultsList.jsx
```diff
@@ -1,49 +1,40 @@
...
-                  </ListItem>
-                </ListItemButton>
-              ))}
-          </List>
-        </Box>
-      </Grid>
-    </Grid>
+    searchIsFocused && (
+      <Box
+        id="search-revision-list-box"
```
kimberlythegeek, line 15: added ID for detecting when this element is clicked (or not)

src/components/Search/SearchResultsListItem.jsx
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <ListItemButton key={item.id}>
+      <ListItem className="search-revision-item search-revision" disablePadding>
+        <ListItemIcon className="search-revision-item-icon search-revision">
+          <Checkbox
+            className="search-revision-item-checkbox search-revision"
+            edge="start"
+            tabIndex={-1}
+            disableRipple
+          />
```
kimberlythegeek, line 25: added checkboxes to select multiple revisions. Will need to add validation in the future to prevent selecting more than four.

src/components/Search/SearchResultsListItem.jsx
```diff
@@ -0,0 +1,60 @@
...
+        </ListItemIcon>
+        <ListItemText
+          className="search-revision-item-text search-revision"
+          primary={`${truncateHash(item)} - ${getLastRevision(item).comments} `}
+          secondary={`${item.author} - ${new Date(item.push_timestamp * 1000)}`}
+          sx={{ noWrap: 'true' }}
+          primaryTypographyProps={{ noWrap: true }}
+          secondaryTypographyProps={{ noWrap: true }}
```
kimberlythegeek, line 33: no wrap to prevent text wrapping and showing the entire (sometimes long) commit message

src/tests/utils/globalSetup.js
```diff
@@ -0,0 +1,5 @@
+import 'regenerator-runtime/runtime';
+
+module.exports = async () => {
+  process.env.TZ = 'UTC';
+};
```
kimberlythegeek, line 5: set time zone for tests

src/utils/searchViewHelper.js
```diff
@@ -50,6 +53,22 @@ const searchViewHelper = {
...
+      if (
+        event.target.id === 'search-revision-input' ||
+        event.target.classList.contains('search-revision') ||
+        event.target.parentElement.classList.contains('search-revision')
+      ) {
+        store.dispatch(updateSearchIsFocused(true));
+      } else {
+        store.dispatch(updateSearchIsFocused(false));
+      }
+    } catch (TypeError) {
```
kimberlythegeek, line 68: throws type error if element clicked has no parent element

src/utils/searchViewHelper.js
```diff
@@ -50,6 +53,22 @@ const searchViewHelper = {
...
     }
   },
+
+  handleClickOutsideInput(event) {
+    try {
+      if (
+        event.target.id === 'search-revision-input' ||
+        event.target.classList.contains('search-revision') ||
+        event.target.parentElement.classList.contains('search-revision')
+      ) {
```
kimberlythegeek, line 63: I realized it was not as simple as detecting if the input was clicked or not. If the results were clicked, they would disappear as they are not the input or encapsulated by it.

there may be some edge cases not covered but this was how I was able to get it to behave as expected.

src/utils/searchViewHelper.js
```diff
@@ -31,9 +32,11 @@ const searchViewHelper = {
     // If repository is selected after search value is input, fetch search results
     if (search !== '') {
       searchViewHelper.searchByRevisionOrEmail(repository, search);
+    } else {
+      // Fetch 10 most recent revisions when repository changes
+      store.dispatch(fetchRecentRevisions(repository));
```
kimberlythegeek, line 37: only fetch recent revisions if there is no search value

src/utils/searchViewHelper.js
```diff
@@ -31,9 +32,11 @@ const searchViewHelper = {
...
     }
-    // Fetch 10 most recent revisions when repository changes
-    store.dispatch(fetchRecentRevisions(repository));
+    store.dispatch(updateSearchIsFocused(true));
```
kimberlythegeek, line 39: set input to be focused, so that recent revisions will display when a repository is selected

src/components/Search/SearchResultsListItem.jsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+
+import Checkbox from '@mui/material/Checkbox';
+import ListItem from '@mui/material/ListItem';
+import ListItemButton from '@mui/material/ListItemButton';
+import ListItemIcon from '@mui/material/ListItemIcon';
+import ListItemText from '@mui/material/ListItemText';
+import PropTypes from 'prop-types';
+
```
beatrice-acasandrei, line 9: Does it make sense to move `truncateHash` and `getLastRevision` in the utils folder?

src/utils/searchViewHelper.js
julienw: Another way that could work is to add 3 event handlers:
1. on the document
2. twice the same handler on both the text input and the result list, that would call `e.stopPropagation()`. Then the event wouldn't bubble to the root element and event handler 1 wouldn't run.

Advantages: you don't need to hardcode ids or classnames.

[This article](https://www.freecodecamp.org/news/event-propagation-event-bubbling-event-catching-beginners-guide/#how-event-bubbling-happens-in-react) seems to give a very good explanation about all this event bubbling thing.


If you want to keep it this way, you may find https://developer.mozilla.org/en-US/docs/Web/API/Element/matches interesting as it could possibly simplify your if condition: `if (event.target.matches("#search-revision-input, .search-revision, .search-revision *")`. I'm not so sure this is better though :-)

src/utils/searchViewHelper.js
julienw: It would be better to test for this specific case instead of catching the error. If for some reason you have a `TypeError` anywhere else (for example in the reducers that run when dispatching the action) this would hide it.

`(event.target.parentElement && event.target.parentElement.classList.contains('search-revision'))`

src/utils/searchViewHelper.js
```diff
@@ -51,6 +54,22 @@ const searchViewHelper = {
...
+  handleClickOutsideInput(event) {
+    try {
+      if (
+        event.target.id === 'search-revision-input' ||
+        event.target.classList.contains('search-revision') ||
+        event.target.parentElement.classList.contains('search-revision')
+      ) {
+        store.dispatch(updateSearchIsFocused(true));
+      } else {
+        store.dispatch(updateSearchIsFocused(false));
```
julienw, line 67: I don't have the big picture clear in my head yet, but it seems to me that having this state in redux makes things more difficult. I believe I'd keep this state inside the React component. Indeed this state isn't useful throughout the full app, but only for this specific UI component.

I feel like this comes from the fact that the event handlers are not in the React component, but instead in this separate object 'searchViewHelper` that does direct calls to `store.dispatch`.

To be explicit: I don't really mind having it in Redux for now so that this can move forward. In the future it could be good to think about components as reusable blocks that can also take props from their parent instead of taking all their props from redux.

src/utils/searchViewHelper.js
julienw: With this it is much easier to test the code locally 👍

src/components/Search/SearchInput.jsx
julienw: I feel like you're not using it anymore.

(but I don't mind if you prefer to keep it)

src/components/Search/SearchResultsList.jsx
julienw: Is this id still used?

src/components/Search/SearchResultsListItem.jsx
julienw: I like it!

julienw (commented):
I tested the code locally and this all seems to work fine!

I left a few comments that could make the code better, especially in the long run. Tell me what you think!

.eslintrc.js
kimberlythegeek: In general we decided against one-off disabling like this as it can get confusing if used in too many places. I try to avoid just turning a rule off and instead configure it if possible, but again this is for test files only.

The rule here is disabled for test files only. I did look into `cross-env` from your other comment, we could add that later on and wouldn't need the global setup file anymore.

src/utils/searchViewHelper.js
kimberlythegeek: #89

src/tests/SearchView.test.jsx
```diff
@@ -80,16 +94,115 @@ describe('Search View', () => {
 
     expect(store.getState().search.searchResults).toStrictEqual(testResults);
 
-    await waitFor(() => expect(queryByText('coconut')).toBeInTheDocument());
-    expect(getByText('coconut')).toBeInTheDocument();
-    expect(getByText('spam')).toBeInTheDocument();
+    await waitFor(() =>
```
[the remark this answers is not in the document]
kimberlythegeek: :ok_hand: I resolved that in #85

src/components/Search/SearchInput.jsx
```diff
@@ -12,47 +12,61 @@ import { connect } from 'react-redux';
...
 import SearchResultsList from './SearchResultsList';
 
-const { handleChangeDropdown, handleChangeSearch } = SearchViewHelper;
+const { handleChangeDropdown, handleChangeSearch, handleClickOutsideInput } =
+  SearchViewHelper;
 
-function SearchInput(props) {
-  const { repository, searchResults } = props;
-  const repoList = ['try', 'autoland', 'mozilla-central'];
+const repoList = ['try', 'autoland', 'mozilla-central'];
```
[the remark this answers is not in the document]
kimberlythegeek: this was actually moved from `SearchInput` into a new component `SearchDropDown` in #85 so (for now) it appears in only this one place

We intend to add more repos eventually but we are sticking with just a few for development purposes.

src/utils/searchViewHelper.js
kimberlythegeek: > It would be better to test for this specific case instead of catching the error. If for some reason you have a `TypeError` anywhere else (for example in the reducers that run when dispatching the action) this would hide it.

after using `matches` from your previous comment, it appears I no longer need to check for an error :ok_hand:

src/components/Search/SearchInput.jsx
kimberlythegeek: it is used to identify the label for the dropdown: `labelId="select-repository"`

src/components/Search/SearchResultsList.jsx
kimberlythegeek: nope!

src/utils/searchViewHelper.js
kimberlythegeek: > It would be better to test for this specific case instead of catching the error. If for some reason you have a `TypeError` anywhere else (for example in the reducers that run when dispatching the action) this would hide it.

after using `matches` from your previous comment, it appears I no longer need to check for an error :ok_hand:

src/utils/searchViewHelper.js
kimberlythegeek: so I was considering making these changes now, but as you stated we have separated UI from logic. I don't know what the general consensus on this is, but from my (beginner's) perspective it does seem cleaner. 

I have an easier time of comprehending everything that is going on when components are UI only and logic is handled in a separate file. 

It's worth noting that Perfherder does _not_ separate UI from logic, and this tool intends to replace it (or at least the compare view, for now). There is quite a lot of computation and passing props between components, and it is very hard to follow.

I imagine that is one reason I'm opinionated on the matter :)

julienw: We can discuss this further later :-)

I'd just want to make sure that you know about the `mapDispatchToProps` parameter to the react-redux `connect` method: this makes it very easy to connect the action dispatches with the react components. There's not much passing props going on when working with connected components.

I think my biggest concern is how `store` is like a global that any file can import anywhere and dispatch actions to. Normally the `store` object is only accessed indirectly through `react-redux`'s `connect` function using the `mapDispatchToProps` option or the `dispach` prop. (Or when using the hooks API, with `useDispatch`).
But obviously with these "helper" files it would mean passing the `dispatch` function to all functions that are called, and  understand this would be cumbersome.

Anyway this can be work for later, I think it would be fairly easy to move all code from the "helper" files back to the same files as the react components, so there's nothing to be really worried about here.

.eslintrc.js
julienw: Ah yeah, I see now that it's only for tests, then it's less a concern for me. I'm not used to these `overrides` properties much as it's fairly new in eslint :-)

julienw (approved):

PR #84, opened 2022-04-16
Add ability to search by short hash

kimberlythegeek:
* added ability to search by short hash
* fixed regex to match exactly and only include a-f 0-9
* remove repeated assertions, change test strings to hashes

### DIFF
--- a/src/tests/SearchView.test.jsx
+++ b/src/tests/SearchView.test.jsx
@@ -144,7 +144,7 @@ describe('Search View', () => {
     );
   });
 
-  it('should fetch revisions by ID if searchValue is a 40 character alphanumeric string', async () => {
+  it('should fetch revisions by ID if searchValue is a 12 or 40 character hash', async () => {
     const spyOnFetch = jest.spyOn(global, 'fetch').mockImplementation(() =>
       Promise.resolve(
         Promise.resolve({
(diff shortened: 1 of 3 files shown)


kimberlythegeek:
fixes #68

src/utils/searchViewHelper.js
```diff
@@ -13,11 +13,12 @@ import {
 const searchViewHelper = {
   searchByRevisionOrEmail(repository, search) {
     const emailMatch = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
-    const longHashMatch = /[a-zA-Z0-9]{40}/;
+    const longHashMatch = /\b[a-f0-9]{40}\b/;
```
kimberlythegeek, line 16: realized when testing that it was matching to any alphanumeric string up to 40 characters, instead of exactly 40, and revisions contain only hex digits, so I'm including only letters a-f

src/tests/SearchView.test.jsx
```diff
@@ -362,9 +372,13 @@ describe('Search View', () => {
     const searchInput = getByRole('textbox');
 
     await user.type(searchInput, 'coconut');
+    await user.clear(searchInput);
     await user.type(searchInput, 'spam@eggs');
+    await user.clear(searchInput);
     await user.type(searchInput, 'spamspamspamand@eggs.');
+    await user.clear(searchInput);
     await user.type(searchInput, 'iamalmostlongenoughtobeahashbutnotquite');
+    await user.clear(searchInput);
```
kimberlythegeek, line 381: not clearing search input meant that each value was being appending to searchValue

eng-esther (approved):

PR #85, opened 2022-04-16
Add validation to inputs

kimberlythegeek:
Display error message when

- repository is not selected when searching
- no results are found
- search term does not match a hash or email
- an error occurs.

This required adding a timeout as previously suggested by @julienw , otherwise it would display an error as soon as the user started typing

I replaced the previous error message in state with the helper text, as error message was not being used anywhere.

The tests needed to be updated to handle the timeout, else the tests all failed.


src/components/Search/SearchView.jsx
```diff
@@ -14,7 +16,12 @@ export default function SearchView() {
           PerfCompare
         </Typography>
       </Box>
-      <SearchInput />
+      <Grid container>
+        <Grid item xs={2} />
+        <SearchDropdown />
+        <SearchInput />
+        <Grid item xs={2} />
+      </Grid>
```
kimberlythegeek, line 24: using grid and spacers so that elements will display as expected now that they are in separate components

src/tests/SearchView.test.jsx
```diff
@@ -65,7 +64,7 @@ describe('Search View', () => {
         }),
       ),
     );
-    const user = userEvent.setup();
+    const user = userEvent.setup({ delay: null });
```
kimberlythegeek, line 67: this was added to prevent timeout when using `async/await` in conjuction with `useFakeTimers`

src/utils/searchViewHelper.js
```diff
@@ -40,15 +53,27 @@ const searchViewHelper = {
...
+      }
+    };
+
+    if (_timeout) clearTimeout(_timeout);
 
     // If search input is cleared, clear results
     if (search === '') store.dispatch(updateSearchResults([]));
-
-    if (repository === '') {
-      console.log('Please select a repository');
```
kimberlythegeek, line 48: no more log statement!

beatrice-acasandrei (commented):
If the user types in an email, the error message for repository will be displayed. After choosing the repo the message should go away, but it doesn't. Is this something that needs to be tackled in this patch?

beatrice-acasandrei (commented):
I'm a bit concerned that after typing in an email there's still a call to get the 10 most recent pushes. I've noticed that this call affects the displayed list of revisions. 
Steps to reproduce:
1. Select a repo.
2. Type in an wrong email. The "No results found" message is displayed.
3. Change the repo. The message is still there, but you'll get the most recent revisions displayed also, which can make you think that the results are for that specific email.

kimberlythegeek:
> If the user types in an email, the error message for repository will be displayed. After choosing the repo the message should go away, but it doesn't. Is this something that needs to be tackled in this patch?

ah I didn't catch that, thanks

> I'm a bit concerned that after typing in an email there's still a call to get the 10 most recent pushes when the repo is changed. I've noticed that this call affects the displayed list of revisions. Steps to reproduce:

I think I addressed this in a different patch with an `else` to only fetch recent revisions if there is no search value. I'll add that here

julienw:
Maybe this was discussed in the past, but here are a few comments about the intent for this patch:

> Display error message when
>
> *    repository is not selected when searching

=> I wonder why this state exists. Would it be a good idea to remove the "unselected" option, and instead preselect some sensible default such as `mozilla-central`?

> *   no results are found

=> makes sense, but maybe it shouldn't be an "error", rather just a warning or a simple message

> *   search term does not match a hash or email

=> I expect that in the future we may be able to autocomplete an email, and therefore maybe this error will need to be removed; But that's really for the future so I'm fine with adding an error for this case for now.

> *   an error occurs.

=> makes sense

src/tests/SearchView.test.jsx
```diff
@@ -9,8 +9,14 @@ import userEvent from '@testing-library/user-event';
...
-const { getByLabelText, getByRole, getByText, queryByRole, queryByText } =
-  screen;
+const {
+  findByRole,
+  getByLabelText,
+  getByRole,
+  getByText,
+  queryByRole,
+  queryByText,
+} = screen;
```
julienw, line 19: That's really up to you, but I find it easier to just call `screen.getByText` (and friends) directly in the test code. The main rationale is that when you have to use a different method that's not in the declaration yet, you don't have to add the new one, and possibly remove the old one if it's not used anymore.

src/tests/SearchView.test.jsx
julienw: This comment would be better as a real comment in the code (and in all other places where you use this).

Also, have you considered [advanceTimers](https://testing-library.com/docs/user-event/options#advancetimers) as another option? I'm not saying this is better as I don't know much about the library, so I'll leave the choice up to your judgment.

src/tests/SearchView.test.jsx
```diff
@@ -345,23 +375,37 @@ describe('Search View', () => {
...
       'She turned me into a newt',
     );
   });
 
   it('should not call fetch if searchValue is not a hash or email', async () => {
-    const spyOnFetch = jest.spyOn(global, 'fetch');
-    const user = userEvent.setup();
+    const spyOnFetch = jest.spyOn(global, 'fetch').mockImplementation(() =>
+      Promise.resolve(
+        Promise.resolve({
```
julienw, line 397: I believe you want something such as:
```js
Promise.resolve({
  json: () => Promise.resolve({ results: [] }),
});
```

I see you used the same structure elsewhere, so feel free to fix this in the other locations too.

Let me also explain what `Promise.resolve` does in a few lines. `Promise.resolve(value)` will:
* if `value` isn't a promise (or rather a "thenable", that is an object with a "then" method and some properties, but let's keep it simple) => this returns a resolved promise with the value as the resolved value
* if `value` is a promise => this returns a promise with the same status as the value. This means that:
  * If the input promise is resolved, then the returned promise is resolved with the same resolved value
  * If the input promise is rejected, then the returned promise is rejected with the same rejected value
  * If the input promise is still "pending", then the returned promise is also pending, and will be resolved/rejected whenever the input promise is also resolved/rejected.

This means that constructs like `Promise.resolve(Promise.resolve(something))` or `Promise.resolve(Promise.reject(something))` are redundant.

Therefore `Promise.resolve` is useful when you need to make a promise value from an input value that's not a promise, or that could either be a promise or a non-promise. Also note that `await` does the exact same thing with the value at the right of `await`.

Hope this clears things up :-)

src/tests/SearchView.test.jsx
```diff
@@ -396,27 +449,33 @@ describe('Search View', () => {
...
 
     render(<SearchView />);
 
     const searchInput = getByRole('textbox');
     await user.type(searchInput, 'terryjones@python.com');
 
-    await screen.findByRole('button', { name: 'repository' });
+    await waitFor(() =>
+      expect(queryByText('Please select a repository')).toBeInTheDocument(),
+    );
```
julienw, line 461: I believe `await screen.findByText('Please select a repository')` would be useful here as well.

src/tests/SearchView.test.jsx
```diff
@@ -377,13 +421,22 @@ describe('Search View', () => {
...
       'https://treeherder.mozilla.org/api/project/autoland/push/',
     );
+
+    await waitFor(() =>
+      expect(
+        queryByText(
+          'Search must be a 12- or 40-character hash, or email address',
+        ),
+      ).toBeInTheDocument(),
+    );
```
julienw, line 438: I believe `await screen.findByText('XXX')` would do the same and is more readable too.

src/tests/SearchView.test.jsx
```diff
@@ -438,6 +497,15 @@ describe('Search View', () => {
...
     const searchInput = getByRole('textbox');
     await user.type(searchInput, 'terryjones@python.com');
+    jest.runOnlyPendingTimers();
+
+    expect(spyOnFetch).toHaveBeenCalledWith(
+      'https://treeherder.mozilla.org/api/project/autoland/push/?author=terryjones@python.com',
+    );
+
+    await waitFor(() => expect(queryByText('try')).not.toBeInTheDocument());
+    await waitFor(() => expect(queryByText('coconut')).toBeInTheDocument());
```
julienw, line 507: same remark about using `findByText`

src/tests/SearchView.test.jsx
```diff
@@ -438,6 +497,15 @@ describe('Search View', () => {
 
     const searchInput = getByRole('textbox');
     await user.type(searchInput, 'terryjones@python.com');
+    jest.runOnlyPendingTimers();
+
+    expect(spyOnFetch).toHaveBeenCalledWith(
+      'https://treeherder.mozilla.org/api/project/autoland/push/?author=terryjones@python.com',
+    );
+
+    await waitFor(() => expect(queryByText('try')).not.toBeInTheDocument());
```
julienw, line 506: You could use `waitForElementToBeRemoved` as well, but it's up to you! [more information](https://testing-library.com/docs/guide-disappearance#waiting-for-disappearance)

src/components/Search/SearchDropdown.jsx
```diff
@@ -0,0 +1,86 @@
...
+import MenuItem from '@mui/material/MenuItem';
+import Select from '@mui/material/Select';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+
+import SearchViewHelper from '../../utils/searchViewHelper';
+
+const { handleChangeDropdown } = SearchViewHelper;
+
+function SearchInput(props) {
```
julienw, line 16: A few comments explaining the goal for this component as well as how it's supposed to be used would be useful for a future reader.

src/components/Search/SearchDropdown.jsx
```diff
@@ -0,0 +1,86 @@
...
+  const { dropdownError, dropdownHelperText, repository } = props;
+  const repoList = ['try', 'autoland', 'mozilla-central'];
```
julienw, line 18: I wonder if this list could be useful in a central file, to be reused in other places. This is more a passing-by comment and can wait for a future patch until we need it.

src/components/Search/SearchDropdown.jsx
```diff
@@ -0,0 +1,86 @@
...
+                  id={item}
+                  value={item}
+                  key={item}
+                  onClick={handleChangeDropdown}
+                >
+                  {item}
+                </MenuItem>
+              ))}
+          </Select>
+          <FormHelperText>{dropdownHelperText}</FormHelperText>
```
julienw, line 65: Maybe I'm missing something, but it looks like this line is the only difference (+ the `error` prop). Therefore maybe you could add the condition for these places? This would help to visually see what the differences are between these 2 cases.

For example:
```js
<FormControl error={dropdownError} sx={{ width: '100%' }}>
  <InputLabel>...</InputLabel>
  <Select>...</Select>
  {dropdownHelperText ? <FormHelperText>{dropdownHelperText}</FormHelperText> : null}
</FormControl>
```

This would also make it easier in the future to show some text even when there's no error.

src/components/Search/SearchInput.jsx
```diff
@@ -1,58 +1,47 @@
...
+        /* text field with errors */
+        <TextField
+          error
+          helperText={inputHelperText}
+          label="Search By Revision ID or Author Email"
+          variant="outlined"
+          sx={{ width: '100%' }}
+          onChange={handleChangeSearch}
+        />
+      )}
```
julienw, line 36: The same logic could be applied here, although it's less important here because the JSX is much smaller. So I'll leave up to you here.

src/reducers/SearchSlice.jsx
```diff
@@ -34,6 +47,18 @@ const search = createSlice({
...
+    },
+    updateDropdownHelperText(state, action) {
+      state.dropdownHelperText = action.payload;
+    },
+    updateInputError(state, action) {
+      state.inputError = action.payload;
+    },
+    updateInputHelperText(state, action) {
+      state.inputHelperText = action.payload;
+    },
```
julienw, line 61: It's not 100% clear to me, what the difference between inputError and dropdownError are, it might be a good idea to add comments to explain this here.

src/utils/searchViewHelper.js
```diff
@@ -20,10 +26,19 @@ const searchViewHelper = {
       store.dispatch(fetchRevisionsByAuthor({ repository, search }));
     } else if (longHashMatch.test(search) || shortHashMatch.test(search)) {
       store.dispatch(fetchRevisionByID({ repository, search }));
+    } else {
+      store.dispatch(updateInputError(true));
+      store.dispatch(
+        updateInputHelperText(
+          'Search must be a 12- or 40-character hash, or email address',
+        ),
+      );
```
julienw, line 35: I feel like the same action could be used to change these 2 values, because they're linked.

```js
store.dispatch(setInputInError(errorReason))
```
would set `inputError` to `true` and set the `errorReason` together.
and

```js
store.dispatch(setInputNoError())
```
would set `inputError` to `false` and set the `errorReason` to `''`.

In general it's a good idea to reason about redux actions as user actions, rather as a way to update each individual state properties. That's the reducer's job to update the state from a logical redux action.

Don't worry, this is a quite common error, and it's too late to correct it :-)

src/utils/searchViewHelper.js
```diff
@@ -32,24 +47,35 @@ const searchViewHelper = {
...
-    store.dispatch(fetchRecentRevisions(repository));
   },
 
   handleChangeSearch(event) {
     const search = event.target.value;
     const { repository } = store.getState().search;
     store.dispatch(updateSearchValue(search));
+    store.dispatch(updateSearchResults([]));
+    store.dispatch(updateInputError(false));
+    store.dispatch(updateInputHelperText(''));
```
julienw, line 62: The same comment can be made here. It's probably a good idea to have a dedicated action to update the search value, and the reducers for each of these state values would react to this dedicated action.

Looking at the reducer, this makes it easier to reason about which user actions will change to which values.

In general, seeing several `dispatch` commands like this is a good sign that the actions aren't logical actions.
(Of course this may still happen sometimes, rules can be broken too)

src/utils/searchViewHelper.js
```diff
@@ -3,13 +3,19 @@ import {
...
 } from '../reducers/SearchSlice';
 import {
   fetchRecentRevisions,
   fetchRevisionByID,
   fetchRevisionsByAuthor,
 } from '../thunks/searchThunk';
 
+let _timeout = null;
+
 const searchViewHelper = {
```
julienw, line 19: (not for this patch, more a surprise comment when looking at this code)
I was a bit surprised to see that this helper was touching the `store` directly. Is there a reason why this logic isn't in the react components directly?

src/utils/searchViewHelper.js
```diff
@@ -32,24 +47,35 @@ const searchViewHelper = {
...
 
     // If search input is cleared, clear results
     if (search === '') store.dispatch(updateSearchResults([]));
-
-    if (repository === '') {
-      console.log('Please select a repository');
-    } else {
-      searchViewHelper.searchByRevisionOrEmail(repository, search);
-    }
+    else _timeout = setTimeout(_onTimeout, _idleTime);
```
julienw, line 78: nit: please use brackets for the else block.

I'm surprised that some eslint rule doesn't do that automatically :-)

src/utils/searchViewHelper.js
```diff
@@ -32,24 +47,35 @@ const searchViewHelper = {
...
   handleChangeSearch(event) {
     const search = event.target.value;
     const { repository } = store.getState().search;
     store.dispatch(updateSearchValue(search));
+    store.dispatch(updateSearchResults([]));
+    store.dispatch(updateInputError(false));
+    store.dispatch(updateInputHelperText(''));
+
+    const _idleTime = 500;
+    const _onTimeout = () => {
```
julienw, line 65: (optional) I'm somewhat surprised that you're using the prefix `_` for these 2 variables. Not that I'm opposed, but I was wondering if you were following some conventions for this.

julienw (commented):
Thanks for this change! (and sorry for the delay too!)

I had a quick look but didn't try the code itself. Please tell me if you'd like me to.
Hopefully my comments won't be too bad, happy to discuss this more with you!

src/tests/SearchView.test.jsx
kimberlythegeek: I tested this out, and I think it is similar to `runAllTimers` or `runOnlyPendingTimers`. using these (and `advanceTimers`) doesn't prevent the test from timing out (unless maybe I'm using it improperly)

referenced source: https://github.com/testing-library/user-event/issues/565#issuecomment-800694378

kimberlythegeek:
> > * search term does not match a hash or email
> 
> => I expect that in the future we may be able to autocomplete an email, and therefore maybe this error will need to be removed; But that's really for the future so I'm fine with adding an error for this case for now.

by not matching a hash or email I mean invalid value :)

src/tests/SearchView.test.jsx
kimberlythegeek: I think I just had a headache trying to get the tests to work :sweat_smile: removing the extra resolve did work fine

src/tests/SearchView.test.jsx
kimberlythegeek: I believe I tried that initially. This is the error I get if I replace that line:
```
received value must be an HTMLElement or an SVGElement.
    Received has type:  object
    Received has value: {}
```

kimberlythegeek: if I replace that line here, is it redundant to then `expect` that text `toBeInTheDocument`?

julienw: Yeah I believe so.
You can use `expect(await findByText(...)).toBeInTheDocument()` if you want to be extra explicit :-)
The await is important so that the result of the find is expected.

src/utils/searchViewHelper.js
kimberlythegeek: It was a decision to keep UI and logic separate, to avoid overly complex components (like we currently have in perfherder). Just a preference, I prefer it this way (I believe the others agreed with this also)

src/utils/searchViewHelper.js
kimberlythegeek: feel free to review the updated reducers and make suggestions, I can file a bug and address them in another (smaller) patch :)

src/components/Search/SearchDropdown.jsx
julienw: I guess that now you exported this in a constant file in the other PR, you can import it here after the other PR lands.

I'm fine if you'd rather do it in another small PR too.

src/tests/SearchView.test.jsx
julienw: Thanks for looking at this!

I believe the result is userEvent won't "wait" a small timeout between letters when simulating typing some text in the input. This makes tests a bit more different than what happens in reality, and could hide bugs.
I think it's good to keep it this way for now, there's no need to spend more time on this, but let's just remember that tests are imperfect :-)

julienw (approved):
Thanks, this all looks good to me.
If possible, it would be good that other folks in the team also look at the code before landing, for the sake of sharing the knowledge in the team :-)

PR #86, opened 2022-04-21
add table with selected visions

eng-esther:
Add Perfcompare table, with mock data

Screen Shot 2022-05-05 at 2 41 02 PM
Screen Shot 2022-05-05 at 2 40 54 PM

### DIFF
--- a/src/common/store.ts
+++ b/src/common/store.ts
@@ -2,9 +2,11 @@ import { configureStore } from '@reduxjs/toolkit';
 import { combineReducers } from 'redux';
 
 import search from '../reducers/SearchSlice';
+import selectedRevisions from '../reducers/SelectedRevisions';
 
 const reducer = combineReducers({
   search,
+  selectedRevisions,
 });
 
 export default configureStore({
--- a/src/components/Search/SearchView/SearchView.css
+++ b/src/components/Search/SearchView/SearchView.css
@@ -1,0 +1,11 @@
+.MuiGrid-grid-xs-11 {
+  z-index: 2;
+}
+
+.MuiGrid-grid-xs-12 {
+  position: absolute;
+  z-index: 1;
+
+  top: 200px;
+  width: 70%;
+}
--- a/src/components/Search/SearchView/SearchView.css
+++ b/src/components/Search/SearchView/SearchView.css
@@ -7,5 +7,6 @@
   z-index: 1;
 
   top: 200px;
-  width: 70%;
+  left: 90px;
+  width: 90%;
 }
--- a/src/components/Search/SelectedRevisionsTable/SelectedRevisionsTable.tsx
+++ b/src/components/Search/SelectedRevisionsTable/SelectedRevisionsTable.tsx
@@ -1,0 +1,97 @@
+import * as React from 'react';
+
+import Close from '@mui/icons-material/Close';
+import { TableRow } from '@mui/material';
+import IconButton from '@mui/material/IconButton';
+import Table from '@mui/material/Table';
+import TableBody from '@mui/material/TableBody';
+import TableContainer from '@mui/material/TableContainer';
+import TableHead from '@mui/material/TableHead';
+import { connect, useDispatch } from 'react-redux';
+
+import { deleteRevision } from '../../../reducers/SelectedRevisions';
+import { Revision, State } from '../../../types/state';
+import { TableHeaderDetails } from '../../../types/tableHeaderDetails';
+import StyledTableCell from '../../Shared/StyledTableCell';
+import StyledTableRow from '../../Shared/StyledTableRow';
+import './SelectedRevisionsTable.css';
+
+const tableHeaderDetails: TableHeaderDetails[] = [
+  { id: 1, title: 'Project', alignment: 'left' },
+  { id: 2, title: 'Revision', alignment: 'left' },
+  { id: 3, title: 'Author', alignment: 'left' },
+  { id: 4, title: 'Commit message', alignment: 'center' },
+  { id: 5, title: 'Timestamp', alignment: 'left' },
+];
+
+function SelectedRevisionsTable(props: SelectedRevisionsProps) {
+  const { revisions } = props;
+  const dispatch = useDispatch();
+
+  if (revisions.length === 0) {
+    return <div className="missingRevisions">Add some revisions</div>;
+  }
+  return (
+    <TableContainer className="layout">
+      <Table>
+        <TableHead>
+          <TableRow>
+            <StyledTableCell />
+            {tableHeaderDetails.map((header: TableHeaderDetails) => (
+              <StyledTableCell
+                key={header.id}
+                align={header.alignment}
+                className="title"
+              >
+                {header.title}
+              </StyledTableCell>
+            ))}
+            <StyledTableCell />
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          {revisions.map((row: Revision, index: number) => (
+            <StyledTableRow key={row.id}>
+              <StyledTableCell component="th" scope="row">
+                <div className="cellStyle">{index === 0 ? 'BASE' : 'NEW'}</div>
+              </StyledTableCell>
+              <StyledTableCell component="th" scope="row">
+                {/* need to add the repository name */}
+                {row.repository_id}
+              </StyledTableCell>
+              <StyledTableCell align="left">{row.revision}</StyledTableCell>
+              <StyledTableCell align="left">{row.author}</StyledTableCell>
+              <StyledTableCell align="center">
+                {row.revisions[0].comments}
+              </StyledTableCell>
+              <StyledTableCell align="left">
+                {row.push_timestamp}
+              </StyledTableCell>
+              <StyledTableCell align="right">
+                <IconButton
+                  id="close-button"
+                  onClick={() => dispatch(deleteRevision(row.id))}
+                >
+                  <Close />
+                </IconButton>
+              </StyledTableCell>
+            </StyledTableRow>
+          ))}
+        </TableBody>
+      </Table>
+      <div className="helperText">Maximum 4 revisions.</div>
+    </TableContainer>
+  );
+}
+
+interface SelectedRevisionsProps {
+  revisions: Revision[];
+}
+
+function mapStateToProps(state: State) {
+  return {
+    revisions: state.selectedRevisions.revisions,
+  };
+}
+
+export default connect(mapStateToProps)(SelectedRevisionsTable);
--- a/src/components/Shared/StyledTableCell.tsx
+++ b/src/components/Shared/StyledTableCell.tsx
@@ -1,0 +1,15 @@
+import { styled } from '@mui/material/styles';
+import TableCell, { tableCellClasses } from '@mui/material/TableCell';
+
+const StyledTableCell = styled(TableCell)(() => ({
+  [`&.${tableCellClasses.body}`]: {
+    fontSize: 14,
+  },
+  '&:first-of-type': {
+    width: '50px',
+  },
+  paddingTop: '10px',
+  paddingBottom: '10px',
+}));
+
+export default StyledTableCell;
--- a/src/reducers/SelectedRevisions.tsx
+++ b/src/reducers/SelectedRevisions.tsx
@@ -1,0 +1,62 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+import { SelectedRevisionsState } from '../types/state';
+
+const initialState: SelectedRevisionsState = {
+  revisions: [
+    {
+      id: 1,
+      revision: 'coconut',
+      author: 'john@python.com',
+      push_timestamp: 42,
+      repository_id: 4,
+      revision_count: 0,
+      revisions: [
+        {
+          revision: 'coconut',
+          author: 'john@python.com',
+          comments: 'This is my first message',
+          result_set_id: 0,
+          repository_id: 0,
+        },
+      ],
+    },
+    {
+      id: 2,
+      revision: 'coconut 2',
+      author: 'johncleese@python.com',
+      push_timestamp: 43,
+      repository_id: 4,
+      revision_count: 1,
+      revisions: [
+        {
+          revision: 'coconut 2',
+          author: 'johncleese@python.com',
+          comments: 'This is the second message',
+          result_set_id: 0,
+          repository_id: 0,
+        },
+      ],
+    },
+  ],
+};
+
+const selectedRevisions = createSlice({
+  name: 'selectedRevisions',
+  initialState,
+  reducers: {
+    resetState(state) {
+      state.revisions = initialState.revisions;
+    },
+    deleteRevision(state, action) {
+      return {
+        revisions: state.revisions.filter(
+          (revision) => revision.id !== action.payload,
+        ),
+      };
+    },
+  },
+});
+
+export const { resetState, deleteRevision } = selectedRevisions.actions;
+export default selectedRevisions.reducer;
--- a/src/tests/SelectedRevisionsTable.test.tsx
+++ b/src/tests/SelectedRevisionsTable.test.tsx
@@ -1,0 +1,32 @@
+/**
+ * @jest-environment jsdom
+ */
+import React from 'react';
+
+import SelectedRevisionsTable from '../components/Search/SelectedRevisionsTable/SelectedRevisionsTable';
+import { resetState } from '../reducers/SelectedRevisions';
+import { render, store, screen } from './utils/test-utils';
+
+describe('Search View', () => {
+  afterEach(() => {
+    store.dispatch(resetState());
+  });
+
+  it('should render correctly when there are revisions', () => {
+    render(<SelectedRevisionsTable />);
+    expect(screen.getByText('Maximum 4 revisions.')).toBeInTheDocument();
+  });
+
+  it('should delete revisions after click', async () => {
+    // testing delete functionality
+    render(<SelectedRevisionsTable />);
+    const button = document.querySelectorAll('#close-button');
+
+    await (button[0] as HTMLButtonElement).click();
+    await (button[1] as HTMLButtonElement).click();
+
+    render(<SelectedRevisionsTable />);
+
+    expect(store.getState().selectedRevisions.revisions).toEqual([]);
+  });
+});
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "dependencies": {
         "@emotion/react": "^11.8.2",
         "@emotion/styled": "^11.8.1",
+        "@mui/icons-material": "^5.5.1",
         "@mui/material": "^5.5.1",
         "@reduxjs/toolkit": "^1.7.2",
         "@types/node": "^12.20.48",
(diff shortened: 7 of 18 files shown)


src/components/Search/CompareTable.jsx
```diff
@@ -0,0 +1,151 @@
...
+    <TableContainer
+      component={Paper}
+      style={{
+        boxShadow: 'none',
+        paddingBottom: '50px',
+        display: 'flex',
+        flexDirection: 'column',
+      }}
+    >
+      <Table style={{ minWidth: 700 }} aria-label="customized table">
```
julienw, line 58: `aria-label` is probably not useful in this case, and maybe even harmful. You can remove it.

src/components/Search/CompareTable.jsx
```diff
@@ -0,0 +1,151 @@
...
+            <StyledTableCell align="left" style={{ fontWeight: 600 }}>
+              Timestamp
+            </StyledTableCell>
+            <StyledTableCell />
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          {revisions.map((row) => (
+            <StyledTableRow key={row.id}>
+              <StyledTableCell component="th" scope="row">
```
julienw, line 88: good use of `th` and `scope` 👍

src/components/Search/CompareTable.jsx
```diff
@@ -0,0 +1,151 @@
...
+                <div
+                  style={{
+                    fontWeight: 600,
+                    justifyContent: 'center',
+                    borderRadius: '2.5px',
+                    paddingTop: '5px',
+                    paddingBottom: '5px',
+                    border: '1px solid #d3d3d3',
+                    display: 'flex',
+                    color: '#d3d3d3',
```
julienw, line 98: We need a darker color for accessibility. You can use https://app.contrast-finder.org/?lang=en to find a good one, or maybe https://design.firefox.com/photon/visuals/color.html#grey as well.

julienw (commented):
Thanks, this seems to work fine!
I left a few comments, tell me what you think. Happy to discuss!

Also it would be good to have a few tests. Usually we add them in the same patch as when we add the feature.

src/components/Search/CompareTable.jsx
```diff
@@ -0,0 +1,151 @@
...
+    PropTypes.shape({
+      id: PropTypes.number,
+      projectName: PropTypes.string,
+      revision: PropTypes.string,
+      author: PropTypes.string,
+      commitMessage: PropTypes.string,
+      timeStamp: PropTypes.string,
+    }),
+  ).isRequired,
+};
```
kimberlythegeek, line 143: the structure of this dictionary should match up to the current structure of a revision. See https://github.com/mozilla/perfcompare/blob/master/src/components/Search/SearchResultsListItem.jsx#L38-L56

this is the way the data is returned from the API, and your code should be able to integrate with existing code.

kimberlythegeek (dismissed):

kimberlythegeek:
After viewing the changes (it looks good!) I'm thinking that maybe we should display the table below the search input, and hide it when there are no selected revisions. I think the UI would look cleaner with the table below.

I'd like opinions from others as well @beatrice-acasandrei @esanuandra ?

Also there should be tests to cover any code that was added.

src/components/Search/CompareTable.jsx
```diff
@@ -0,0 +1,151 @@
...
+  paddingBottom: '10px',
+}));
+
+const StyledTableRow = styled(TableRow)(({ theme }) => ({
+  '&:nth-of-type(odd)': {
+    backgroundColor: theme.palette.action.hover,
+  },
+}));
+
+function CompareTable(props) {
```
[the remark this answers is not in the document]
esanuandra: I agree with this.

src/components/Search/CompareTable.jsx
```diff
@@ -0,0 +1,151 @@
...
+  let { revisions } = props;
+
+  const selectedRevisions = [];
+  revisions?.forEach((item) => {
+    if (selectedRevisions.length === 0) {
+      selectedRevisions.push({ ...item, type: 'base' });
+    } else {
+      selectedRevisions.push({ ...item, type: 'new' });
+    }
```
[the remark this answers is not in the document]
eng-esther: yeah so for this, we didnt decide how **base** will be re-assigned. if it was previously deleted.
Indexing it will actually cause more logic, I think once we decide the base re-assignment this will be clearer for me.

src/components/Search/CompareTable.jsx
eng-esther: will remove

src/components/Search/CompareTable.jsx
```diff
@@ -0,0 +1,151 @@
...
+              </StyledTableCell>
+              <StyledTableCell align="left">{row.revision}</StyledTableCell>
+              <StyledTableCell align="left">{row.author}</StyledTableCell>
+              <StyledTableCell align="center">
+                {row.commitMessage}
+              </StyledTableCell>
+              <StyledTableCell align="left">{row.timeStamp}</StyledTableCell>
+              <StyledTableCell align="right">
+                <IconButton
+                  onClick={() => store.dispatch(deleteRevision(row.id))}
```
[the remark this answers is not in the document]
eng-esther: sure
check this code snippet: (using arrows here)
`  <button onClick={() => dispatch({ type: 'increment-counter' })}>`

src/components/Search/SelectedRevisionsTable.jsx
```diff
@@ -0,0 +1,153 @@
+import * as React from 'react';
+
+// import CloseIcon from '@material-ui/icons/Close';
```
esanuandra, line 3: forgot a comment here?

src/components/Search/CompareTable.jsx
kimberlythegeek: The API for performance summary takes a revision ID. the actual comparison is currently done on the front end of Perfherder, but as we've discussed this is not desirable and we want to create a new endpoint for comparison results. I'm assuming this would also take revision IDs? We have not discussed the specifics yet.

The revisions do often have more than one entry. Currently I have helper messages to select the "comment" or commit message from only the last revision in the array (usually the first is the temporary mach commit that includes `try_task_config.json` saying which tests to run)

you make some good points. As I said I already have a helper method to select only the most recent revision, we could later add a helper method to create a new revision item.

kimberlythegeek:
> > Also there should be tests to cover any code that was added.
> 
> yeah we discussed before you left that'll it be better in another path, or would you prefer in this patch?

I prefer writing tests in the same patch to show that any new code is covered.

eng-esther:
> > > Also there should be tests to cover any code that was added.
> > 
> > 
> > yeah we discussed before you left that'll it be better in another path, or would you prefer in this patch?
> 
> I prefer writing tests in the same patch to show that any new code is covered.

cool 👍

src/components/Search/CompareTable.jsx
eng-esther: I'm going to merge the code as is, and will handle in the different patch, the connection between selecting revision and adding to the table havent been done yet, probably best to handle that there.

src/components/Search/SelectedRevisionsTable/SelectedRevisionsTable.jsx
```diff
@@ -0,0 +1,95 @@
...
+import TableContainer from '@mui/material/TableContainer';
+import TableHead from '@mui/material/TableHead';
+import PropTypes from 'prop-types';
+import { connect, useDispatch } from 'react-redux';
+
+import { deleteRevision } from '../../../reducers/SelectedRevisions';
+import { Alignment } from '../../Shared/enums';
+import StyledTableCell from '../../Shared/StyledTableCell';
+import StyledTableRow from '../../Shared/StyledTableRow';
+import './SelectedRevisionsTable.module.css';
```
eng-esther, line 18: css imports not fully functional yet, waiting on probably a PR that has webpack config before I can add the css loaders in the webpack..

src/components/Search/SearchView.jsx
```diff
@@ -34,12 +35,24 @@ export default function SearchView() {
           PerfCompare
         </Typography>
       </Box>
-      <Grid container>
+      <Grid
+        container
+        style={{
+          position: 'absolute',
+          margin: '0 auto',
```
eng-esther, line 42: will move to separate css file once css loader has been added...

kimberlythegeek:
> After viewing the changes (it looks good!) I'm thinking that maybe we should display the table below the search input, and hide it when there are no selected revisions. I think the UI would look cleaner with the table below.
> 
> I'd like opinions from others as well @beatrice-acasandrei @esanuandra ?
> 
> Also there should be tests to cover any code that was added.

After looking at Andra's designs this morning, the "compare" button is below the search input, so perhaps we should leave the revisions above, but I think the table should be hidden when no revisions are selected, or at least remove the headers and display a message "No revisions selected" or something like that

src/components/Search/SelectedRevisionsTable/SelectedRevisionsTable.tsx
```diff
@@ -0,0 +1,106 @@
...
+    PropTypes.shape({
+      id: PropTypes.number,
+      projectName: PropTypes.string,
+      revision: PropTypes.string,
+      author: PropTypes.string,
+      commitMessage: PropTypes.string,
+      timeStamp: PropTypes.string,
+    }),
+  ).isRequired,
+};
```
esanuandra, line 98: Should this be changed with an `interface`?

eng-esther: yeah I think it should be..good catch, thankyou.

julienw (commented):
I'll leave Kimberly review this one, it looks like my initial comments were handled

src/components/Search/SelectedRevisionsTable/SelectedRevisionsTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+  alignment: Alignment;
+}
+
+const tableHeaderDetails: TableHeaderDetails[] = [
+  { title: 'Project', alignment: 'left' },
+  { title: 'Revision', alignment: 'left' },
+  { title: 'Author', alignment: 'left' },
+  { title: 'Commit message', alignment: 'center' },
+  { title: 'Timestamp', alignment: 'left' },
+];
```
kimberlythegeek, line 30: In a followup PR, could you move the interface and array into the shared folder also?

src/reducers/SelectedRevisions.jsx
```diff
@@ -0,0 +1,89 @@
...
+    //   timeStamp: '00:02:00UTC',
+    // },
+    // {
+    //   id: 3,
+    //   projectName: 'raptor 2',
+    //   revision: '35g4bjh54jk345b',
+    //   author: 'author@mozila.com',
+    //   commitMessage: 'Added details for',
+    //   timeStamp: '00:01:00UTC',
+    // },
```
kimberlythegeek, line 64: Be sure to remove the commented lines before merging

src/reducers/SelectedRevisions.jsx
```diff
@@ -0,0 +1,89 @@
...
+};
+
+const compareRevisions = createSlice({
+  name: 'compareRevisions',
+  initialState,
+  reducers: {
+    resetState(state) {
+      state.revisions = initialState.revisions;
+    },
+    addRevision(state, action) {
```
kimberlythegeek, line 75: should we add a test for this that just calls `dispatch` and checks that state has changed?

src/reducers/SelectedRevisions.jsx
```diff
@@ -0,0 +1,89 @@
...
+    //   timeStamp: '00:01:00UTC',
+    // },
+  ],
+};
+
+const compareRevisions = createSlice({
+  name: 'compareRevisions',
+  initialState,
+  reducers: {
+    resetState(state) {
```
kimberlythegeek, line 72: if you add an `afterEach` that calls `resetState` it should show as covered by tests

src/tests/SelectedRevisionsTable.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+              {
+                revision: 'coconut',
+                author: 'john@python.com',
+                comments: 'This is my first message',
+              },
+            ],
+          },
+        ],
+      },
+    } as any);
```
kimberlythegeek, line 53: later on we can update this test to use data from the fixtures file (will be added in #90 )

src/tests/SelectedRevisionsTable.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    render(<SelectedRevisionsTable />);
+    expect(screen.getByText('Maximum 4 revisions.')).toBeInTheDocument();
+
+    // testing delete functionality
+    const button = document.querySelector('#close-button') as HTMLButtonElement;
+
+    await button.click();
+
+    expect(mockDispatch).toHaveBeenCalled();
```
kimberlythegeek, line 63: I'm not sure if this is specific enough. Can you add a check that the deleted item is no longer in the document? example: https://github.com/mozilla/perfcompare/blob/master/src/tests/SearchView.test.tsx#L142-L144

kimberlythegeek (changes requested):
Added a few suggestions for later patches, and updates to tests

kimberlythegeek:
I saw that your latest commit failed tests due to linting errors. If your editor supports it, I'd recommend adding a linter extension to format on save. I use VSCode and it has a prettier extension. Less work for me to do if it's done automatically!

src/tests/SelectedRevisionsTable.test.tsx
eng-esther: sure, would you wanna create a jira item for this?

src/reducers/SelectedRevisions.jsx
eng-esther: removed add revision..

src/components/Shared/StyledTableCell.tsx
```diff
@@ -0,0 +1,15 @@
...
+    fontSize: 14,
+  },
+  '&:first-of-type': {
+    width: '50px',
+  },
+  paddingTop: '10px',
+  paddingBottom: '10px',
+}));
+
+export default StyledTableCell;
```
kimberlythegeek, line 15: I don't agree with creating new components just to add styles. I realize it does this in the docs examples, but we have a theme already, and I'd prefer to add this there.

[example for table styles](https://gist.github.com/kimberlythegeek/2d5806d660330f600b4930f2f675ead4)

and you can set the default body font size to 14 in `typography.js`: `fontSize: 14` below the `fontFamily` attribute

src/components/Search/SearchView/SearchView.css
```diff
@@ -0,0 +1,7 @@
+.gridStyle {
+  position: absolute ;
+  margin: '0 auto';
+  left: 0;
+  right: 0;
+  z-index: 1;
+}
```
kimberlythegeek, line 7: I'm unsure why this style was added, but it changes the rendering of the SearchInput. we can remove this and size the table using `<Grid>` instead

src/components/Search/SelectedRevisionsTable/SelectedRevisionsTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+function SelectedRevisionsTable(props: SelectedRevisionsProps) {
+  const { revisions } = props;
+  const dispatch = useDispatch();
+
+  if (revisions.length === 0) {
+    return <div className="missingRevisions">Add some revisions</div>;
+  }
```
kimberlythegeek, line 33: if there are no revisions we should not render anything.

you can use conditional rendering here: https://github.com/mozilla/perfcompare/blob/master/src/components/Search/SearchView.tsx#L55-L57

src/components/Search/SelectedRevisionsTable/SelectedRevisionsTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+  return (
+    <TableContainer className="layout">
+      <Table style={{ minWidth: 700 }}>
```
kimberlythegeek, line 36: this is a duplicate, but also I don't think we need this, size with Grid instead.

src/components/Search/SelectedRevisionsTable/types.tsx
```diff
@@ -0,0 +1,6 @@
+import { Alignment } from '../../../types/enums';
+
+export interface TableHeaderDetails {
+  title: string;
+  alignment: Alignment;
+}
```
kimberlythegeek, line 6: I think this should go in the `src/types` folder. also this does not need to be a `tsx` file

src/common/store.ts
```diff
@@ -2,9 +2,11 @@ import { configureStore } from '@reduxjs/toolkit';
 import { combineReducers } from 'redux';
 
 import search from '../reducers/SearchSlice';
+import compareTable from '../reducers/SelectedRevisions';
 
 const reducer = combineReducers({
   search,
+  compareTable,
```
kimberlythegeek, line 9: `compareRevisions` and `compareTable` don't really reflect what this table is, and I think it would overlap with names used for the compare view. Please use `selectedRevisions` instead

src/reducers/SelectedRevisions.tsx
```diff
@@ -0,0 +1,54 @@
...
+          revision: 'coconut 2',
+          author: 'johncleese@python.com',
+          comments: 'This is the second message',
+        },
+      ],
+    },
+  ],
+};
+
+const compareRevisions = createSlice({
```
kimberlythegeek, line 36: replace with selectedRevisions instead

src/components/Shared/StyledTableCell.tsx
```diff
@@ -0,0 +1,15 @@
+import { styled } from '@mui/material/styles';
```
kimberlythegeek, line 1: this is deprecated: https://github.com/mui/material-ui/blob/master/packages/mui-material/src/styles/createStyles.js#L8-L9

see below comment for styling table rows.

src/reducers/SelectedRevisions.tsx
```diff
@@ -0,0 +1,54 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+const initialState = {
```
kimberlythegeek, line 3: I'm unsure why the linter didn't warn about this, but there should be a type here (see: https://github.com/mozilla/perfcompare/blob/master/src/types/state.ts#L24-L30). 

I will look into that.

kimberlythegeek (changes requested):

src/components/Shared/StyledTableCell.tsx
eng-esther: https://mozilla-hub.atlassian.net/browse/FXP-2201

src/components/Search/SelectedRevisionsTable/SelectedRevisionsTable.tsx
eng-esther: https://mozilla-hub.atlassian.net/browse/FXP-2202

src/components/Search/SearchView/SearchView.css
```diff
@@ -0,0 +1,11 @@
+.MuiGrid-grid-xs-11 {
+  z-index: 2;
+}
+
+.MuiGrid-grid-xs-12 {
+  position: absolute;
+  z-index: 1;
+
+  top: 200px;
+  width: 70%;
```
kimberlythegeek, line 10: change this width to 90% to better center the table. Everything else looks good :ok_hand:

kimberlythegeek (changes requested):

kimberlythegeek (approved):
Looks good, thanks!

beatrice-acasandrei (approved):

esanuandra (approved):

PR #88, opened 2022-04-22
Convert to TypeScript

kimberlythegeek:


.eslintrc.js
```diff
@@ -18,6 +18,13 @@ module.exports = {
     sourceType: 'module',
   },
   plugins: ['react'],
+  settings: {
+    'import/resolver': {
+      node: {
+        extensions: ['.js', '.jsx', '.ts', '.tsx'],
+      },
+    },
+  },
```
kimberlythegeek, line 27: this was needed to get rid of errors for not including file extension in imports. with or without file extension, I was getting errors and this resolved it

.eslintrc.js
```diff
@@ -53,20 +60,40 @@ module.exports = {
       'error',
       { ImportDeclaration: { multiline: true } },
     ],
-    // require line break after operator
-    'operator-linebreak': ['error', 'after'],
+    'operator-linebreak': 'off',
```
kimberlythegeek, line 63: does not behave properly for ternary operators

.eslintrc.js
```diff
@@ -53,20 +60,40 @@ module.exports = {
...
+    {
+      files: ['*.ts', '*.tsx'],
+      excludedFiles: 'src/tests/**',
+      extends: ['airbnb-typescript'],
+      parserOptions: {
+        project: './tsconfig.json',
+      },
+      rules: {
+        // https://github.com/typescript-eslint/typescript-eslint/issues/1824
+        '@typescript-eslint/indent': 'off',
```
kimberlythegeek, line 96: see github issue

src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { fetchRecentRevisions } from '../../thunks/searchThunk';
+import { Revision, State } from '../../types/state';
+import { handleChangeSearch } from '../../utils/searchViewHelper';
+import SearchResultsList from './SearchResultsList';
+
+class SearchInput extends React.Component<SearchInputProps> {
+  componentDidMount() {
+    const repo = store.getState().search.repository;
+    store.dispatch(fetchRecentRevisions(repo));
+  }
```
kimberlythegeek, line 17: fetch recent revisions on mount now that we have a default repo set. this will display recent revisions when focusing input.

src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  render() {
+    const {
+      focused,
+      handleFocus,
+      handleChildClick,
```
kimberlythegeek, line 23: `focused` `handleFocus` and `handleChildClick` are for detecting if results should be hidden using Julien's suggestion from #89

src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+            label="Search By Revision ID or Author Email"
+            id="search-revision-input"
+            onFocus={handleFocus}
+            variant="outlined"
+            sx={{ width: '100%' }}
+            onChange={handleChangeSearch}
+          />
+        )}
+        {searchResults.length > 0 && focused && (
+          <SearchResultsList handleChildClick={handleChildClick} />
```
kimberlythegeek, line 54: detects when user clicks inside search results so it does not hide results

src/components/Search/SearchInput.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+interface SearchInputProps {
+  handleFocus: (e: React.FocusEvent) => void;
+  handleChildClick: (e: React.MouseEvent) => void;
+  inputError: boolean;
+  inputHelperText: string;
+  focused: boolean;
+  searchResults: Revision[];
+}
```
kimberlythegeek, line 68: using `interface` instead of `PropTypes` for typescript

src/components/Search/SearchView.tsx
```diff
@@ -1,17 +1,37 @@
...
+  const handleFocus = (e: React.FocusEvent) => {
+    e.stopPropagation();
+    setFocused(true);
+  };
+
+  const handleChildClick = (e: React.MouseEvent) => {
+    e.stopPropagation();
+    setFocused(true);
+  };
+
```
kimberlythegeek, line 32: these functions are replacing `handleClickOutsideInput` and using bubbling instead of hard-coded IDs or class names

`stopPropagation` prevents event handlers of parents from running when a child is clicked/focused

src/reducers/SearchSlice.tsx
```diff
@@ -64,23 +65,29 @@ const search = createSlice({
       })
       .addCase(fetchRecentRevisions.rejected, (state, action) => {
         state.inputError = true;
-        state.inputHelperText = action.payload;
+        state.inputHelperText = action.payload
+          ? action.payload
+          : 'An error has occurred';
```
kimberlythegeek, line 70: I had to check for undefined here, or else I would get a type error of `string | undefined`

src/thunks/searchThunk.ts
```diff
@@ -0,0 +1,73 @@
+import { createAsyncThunk } from '@reduxjs/toolkit';
+
+import { treeherderBaseURL } from '../common/constants';
+import type { Repository, Revision } from '../types/state';
+
+export const fetchRecentRevisions = createAsyncThunk<
+  Revision[],
+  Repository['name'],
+  { rejectValue: string }
```
kimberlythegeek, line 9: types for parameter, return value, and reject value

src/thunks/searchThunk.ts
```diff
@@ -0,0 +1,73 @@
...
+>('search/fetchRecentRevisions', async (repository, { rejectWithValue }) => {
+  let response;
+  try {
+    response = await fetch(
+      `${treeherderBaseURL}/api/project/${repository}/push/`,
+    );
+  } catch (err) {
+    return rejectWithValue((err as Error).message);
```
kimberlythegeek, line 17: `(err as Error)` added to avoid type error

src/types/custom.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
+  const content: any;
+  export default content;
+}
```
kimberlythegeek, line 4: added to handle svg files in typescript

src/utils/searchViewHelper.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  // If search input is cleared, clear results
+  if (search === '') {
+    store.dispatch(updateSearchResults([]));
+  } else {
+    // Submit API call 500ms after user stops typing
+    timeout = setTimeout(onTimeout, idleTime);
+  }
+};
+
```
kimberlythegeek, line 78: removed `handleClickOutsideInput`

src/components/Search/SearchView.tsx
kimberlythegeek: I added this here instead of a helper file because I'm using `useState` to set the value of `focused`. I may be able to move this out, but I was running into issue with passing `setFocused` as a parameter due to not setting the type correctly

kimberlythegeek:
I replaced `handleClickOutsideInput` with event handlers as Julien suggested in #89 

I did so in this patch instead of a later one because I was struggling with correctly setting the type of its parameters. I think this comes from the fact that `addEventListener` is a javascript event handler, whereas things like `onClick` and `onFocus` are React handlers. I tried for a good while to resolve this, but this is the only way I got it working.

I also separated the tests into more specific test files, as `SearchView.test` was getting to be a pretty big file.

I realize that for some files converted to typescript, it shows them as "deleted" and "added" instead of modified, making it difficult to compare changes because there is no diff between them. I added comments about what was changed and why to make this easier.

esanuandra (approved):

src/components/Search/SearchInput.tsx
julienw: It would be better to have behavior changes in a separate PR, to make it easier to review.

I believe this PR should be technical only and not bring behavior changes.

src/reducers/SearchSlice.tsx
julienw: That's common when using type systems. Sometimes you _know_ something can't happen, but you still need to check it. I believe this is a small price to pay for the added type safety.

julienw (commented):
I skimmed through the files and this looks reasonable. I also tried locally and everything seems to work.

However in this PR you made 2 changes in one PR: the JS -> TS change and the event handling change. This makes it difficult to look in details.
A more targeted PR means that errors are easier to spot. Here it's nearly impossible to see if the new code is correct, as it's buried in the typescript-related change.

In the end I think that a good PR should be easy to review. We need to minimize the time a reviewer has to spend on a PR to review it. The advantage for the reviewer is that they spend less time on it (and that's *2 *3 if you have 2 or 3 reviewers). The advantage for the author is that the review time is smaller. Another advantage for the author is that it's easier to put the PR in an approved state.

This is a win-win situation.

Therefore if you can do the PR again without the behavior change that would be wonderful.  Otherwise let's move forward with that but please keep this in mind for future work.

Thanks!

.eslintrc.js
kimberlythegeek: Yeah, I have a VSCode prettier extension enabled, which is part of the issue. I get an error from ESLint, then save the file and Prettier auto-formats it. I've found that often ESLint rules conflict with Prettier, and I'm unsure how to give priority to Prettier (do we drop ESLint altogether?) We can discuss next week :)

src/tests/utils/constants.js
```diff
@@ -0,0 +1,32 @@
...
+      {
+        revision: 'spam',
+        author: 'ericidle@python.com',
+        comments: "it's just a flesh wound",
+      },
+    ],
+  },
+];
+
+export default testResults;
```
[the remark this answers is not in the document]
kimberlythegeek: I believe there's a lint rule that throws an error "prefer default exports" if a file contains only one export. This might be another ESLint vs Prettier issue

src/tests/SearchView.test.tsx
```diff
@@ -0,0 +1,172 @@
...
+    expect(
+      screen.queryByText("coconut - you've got no arms left!"),
+    ).toBeInTheDocument();
+    expect(
+      screen.queryByText("spam - it's just a flesh wound"),
+    ).toBeInTheDocument();
+
+    expect(document.body).toMatchSnapshot();
+  });
+});
```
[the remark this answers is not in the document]
kimberlythegeek: I divided the tests into separate files, as this one had gotten to be a very large file. Apologies for my bad habit of making too many changes in one PR :grimacing: I do try, but I'll do a better job of avoiding this in the future.

kimberlythegeek:
so I made a separate *just* renaming js/jsx to ts/tsx, and following commits with actual changes to the files, hoping that it would show a diff for all of those files. Although that first commit does show "renamed" for all of those files, the summary for all changes still shows some js/jsx files as "deleted" and the resulting ts/tsx files as "added" :shrug:

kimberlythegeek:
I removed behavior changes including: creating a constants file for tests, separating tests in to multiple files, separate jest config file, and fetching recent revisions for `try` on mount

.eslintrc.js
julienw: Not sure this is the reason, but the documentation for the eslint plugin says that you should put `plugin:prettier/recommended` last. So maybe this is it?
This should configure `eslint-config-prettier` which disables all stylistic rules, but it needs to come _after_ the others that might enable them.

julienw: To be done in another PR :-)

src/reducers/SearchSlice.tsx
```diff
@@ -28,16 +27,12 @@ const search = createSlice({
     // BEGIN used for testing only
     resetState(state) {
       state.repository = initialState.repository;
-      state.searchIsFocused = initialState.searchIsFocused;
```
julienw, line 31: I see this change is still here in this otherwise technical PR.

src/thunks/searchThunk.ts
julienw: (for another PR) I wonder if this try/catch is useful here -- my expectation is that this would be "rejected with value" with the error automatically if the fetch call throws. Although I see you force the rejectValue to be a string, but maybe this should be an Error? Up to you actually.

(another PR too) Also note the `response.json()` call below may throw as well (if the response's body isn't valid json), so it might make sense to put it inside the try/catch block.

src/components/Search/SearchResultsList.tsx
```diff
@@ -28,39 +29,24 @@ function SearchResultsList(props) {
...
-      revision_count: PropTypes.number,
-      push_timestamp: PropTypes.number.isRequired,
-      repository_id: PropTypes.number.isRequired,
-    }),
-  ).isRequired,
-};
+interface SearchResultsListProps {
+  searchResults: Revision[];
+  handleChildClick: (e: React.MouseEvent) => void;
+}
```
julienw, line 46: yay, so much more convenient

src/utils/searchViewHelper.ts
```diff
@@ -12,79 +13,65 @@ import {
   fetchRevisionByID,
   fetchRevisionsByAuthor,
 } from '../thunks/searchThunk';
+import type { Repository } from '../types/state';
 
-let timeout = null;
+let timeout: NodeJS.Timeout;
```
julienw, line 18: I'm a bit surprised with the `NodeJS` bit -- and at the same time I'm happy that it's not simply a `number` so that its use is better framed. Let's keep it this way for now but we may need to adjust in the future.

src/utils/searchViewHelper.ts
```diff
@@ -12,79 +13,65 @@ import {
...
+  if (search === '') {
+    store.dispatch(updateSearchResults([]));
+  } else {
+    // Submit API call 500ms after user stops typing
+    timeout = setTimeout(onTimeout, idleTime);
+  }
 };
 
-export default searchViewHelper;
+export { handleChangeDropdown, handleChangeSearch };
```
julienw, line 77: optional nit: I believe you could do directly `export const searchByRevisionOrEmail = ...`

src/tests/SearchView.test.tsx
```diff
@@ -66,13 +66,14 @@ describe('Search View', () => {
   });
 
   it('should fetch and display recent results when repository is selected', async () => {
-    const spyOnFetch = jest.spyOn(global, 'fetch').mockImplementation(() =>
+    global.fetch = jest.fn(() =>
```
julienw, line 69: (not for this patch)
I moved our mocking of fetch to the `fetch-mock-jest` util in https://github.com/firefox-devtools/profiler/pull/4020, in case you want to have a look at this.

julienw (approved):
This was easier to review, thanks!

Some behavior changes are still present, but that's ok for this PR. Let's try to make it cleaner next time :-)

It might be worth adding a type checker command in package.json and adding it to the CI so that it's checked automatically. It would recommend to do it in another PR.

webpack.config.js
```diff
@@ -0,0 +1,35 @@
...
+  entry: './src/app.jsx',
+  output: {
+    filename: './bundle.js',
+  },
+  resolve: {
+    // changed from extensions: [".js", ".jsx"]
+    extensions: ['.ts', '.tsx', '.js', '.jsx'],
+  },
+  module: {
+    rules: [
```
eng-esther, line 12: we may need to add css configuration in webpack.config.
https://webpack.js.org/loaders/css-loader/

src/utils/searchViewHelper.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+
+const handleChangeSearch = (
+  event: FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+) => {
+  const search = event.currentTarget.value;
+  const { repository } = store.getState().search;
+  store.dispatch(updateSearchValue(search));
+  store.dispatch(updateSearchResults([]));
+  store.dispatch(clearInputError());
```
eng-esther, line 58: Julien suggested using useDispatch and to probably avoid using store directly, not sure if it applies here.

kimberlythegeek: I can address that in another PR. These function had already been using `store.dispatch` so that part isn't new

PR #92, opened 2022-05-05
moved focus logic to hooks file

kimberlythegeek:
I wasn't sure how to do this before, but I found an example today


src/hooks/focusInput.ts
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+const focusInput = () => {
```
julienw, line 3: nit: my understanding is that there's a convention that hooks use a name that look like `useXXX`. What about `useFocusInput` for this one?

Also it would be good to have a comment explaining what the hook does, with an example of how to use it.

julienw (approved):
I don't know much about hooks yet but this looks good to me. I haven't tried the code locally but I expect you did it :)

Do we have tests that exercize this?

kimberlythegeek:
> I don't know much about hooks yet but this looks good to me. I haven't tried the code locally but I expect you did it :)
> 
> Do we have tests that exercize this?

the test coverage is still 100%--do I need a test specifically for the hook? the checks weren't integrating with CircleCI so it didn't so in the PR, but that should be resolved now.

PR #93, opened 2022-05-05
fetch recent revisions on mount

kimberlythegeek:


kimberlythegeek:
I did this in the App components instead of SearchView because of the changes in #92 

In order to use hooks (`useState`, etc), the component must be a functional component, not a class component. I could have made SearchView a class and used local `state`, but this seemed more complicated for using the hooks.

src/components/App.tsx
```diff
@@ -4,23 +4,32 @@ import CssBaseline from '@mui/material/CssBaseline';
 import { ThemeProvider } from '@mui/material/styles';
 import { BrowserRouter as Router, Route, Routes } from 'react-router-dom';
 
+import store from '../common/store';
 import protocolTheme from '../theme/protocolTheme';
+import { fetchRecentRevisions } from '../thunks/searchThunk';
```
beatrice-acasandrei, line 9: You can now add as param` hide_reviewbot_pushes=true` to endpoint call to exclude results for reviewbot

julienw (changes requested):
I believe it would be better to move the fetch in another new component that renders nothing instead of adding it here.

You can see more information in the bigger component. Happy to discuss this more!

src/components/App.tsx
```diff
@@ -4,23 +4,32 @@ import CssBaseline from '@mui/material/CssBaseline';
...
 import SearchView from './Search/SearchView';
 
-function App() {
-  return (
-    <React.Fragment key="App Component">
```
julienw, line 12: while I'm here, I wonder why there is this key, it doesn't seem useful.
Also `React.Fragment` can be shortened to `<> .. </>`

src/components/App.tsx
```diff
@@ -4,23 +4,32 @@ import CssBaseline from '@mui/material/CssBaseline';
...
-          </Routes>
-        </Router>
-      </ThemeProvider>
-    </React.Fragment>
-  );
+class App extends React.Component {
+  componentDidMount() {
+    const repo = store.getState().search.repository;
+    store.dispatch(fetchRecentRevisions(repo));
+  }
```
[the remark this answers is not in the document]
kimberlythegeek: I was beginning to work on this, but there's a bit more work involved than I expected. I've filed #98 and I'll begin working on that

src/components/App.tsx
kimberlythegeek: #99

unsure why I added the fragment but it isn't needed anymore

beatrice-acasandrei (approved):
LGTM

esanuandra (approved):

PR #100, opened 2022-05-10
fix typescript linting

kimberlythegeek:
Typescript linting isn't correctly warning about types.

### DIFF
--- a/src/types/api.ts
+++ b/src/types/api.ts
@@ -1,0 +1,17 @@
+import type { Repository, Revision } from './state';
+
+type APIFilterParams = {
+  full: boolean;
+  count: number;
+};
+
+type APIPushMeta = {
+  count: number;
+  repository: Repository['name'];
+  filter_params: APIFilterParams;
+};
+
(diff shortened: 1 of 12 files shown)


src/types/api.ts
```diff
@@ -0,0 +1,17 @@
+import type { Repository, Revision } from './state';
+
+type APIFilterParams = {
+  full: boolean;
```
esanuandra, line 4: what does `full` stand for?

beatrice-acasandrei (approved):

src/types/api.ts
beatrice-acasandrei: > what does `full` stand for?

I think it's a query param used to make the call Perfherder's Compare View to get the recent pushes. @kimberlythegeek I've looked at the code and it seems that this param is not needed as per this comment https://github.com/mozilla/treeherder/blob/master/treeherder/webapp/api/push.py#L174

esanuandra (approved):

Issue #10, opened 2022-02-26
Compare View: information on what metrics means


kimberlythegeek:
User feedback has shown most do not know what our metrics represent / are measuring, i.e. "fnbpaint" 
This information should be presented in the UI in a concise way, without having to link to external documentation

cybraia:
Hi, @kimberlythegeek  @Carla-Moz I would like to work on this. Could you give me more details about fubpaint? Let me know where the UI has to be changed. Thanks!

PR #102, opened 2022-05-12
Handle checked revisions

kimberlythegeek:
Added hook to handle checked revisions, and output a message to console when a user tries to select more than four. Added TODO to replace console message with a modal notification.

This PR is based on #100 and #93 and will be blocked until those are merged.

### DIFF
--- a/src/components/Search/SearchResultsList.tsx
+++ b/src/components/Search/SearchResultsList.tsx
@@ -4,12 +4,13 @@ import Box from '@mui/material/Box';
 import List from '@mui/material/List';
 import { connect } from 'react-redux';
 
+import useCheckRevision from '../../hooks/useCheckRevision';
 import type { Revision, State } from '../../types/state';
 import SearchResultsListItem from './SearchResultsListItem';
 
 function SearchResultsList(props: SearchResultsListProps) {
-  const { searchResults, handleChildClick } = props;
-
+  const { searchResults } = props;
+  const { checked, handleToggle } = useCheckRevision();
   return (
     <Box
       sx={{
@@ -32,7 +33,8 @@ function SearchResultsList(props: SearchResultsListProps) {
           <SearchResultsListItem
             key={item.id}
             item={item}
-            handleChildClick={handleChildClick}
+            checked={checked}
+            handleToggle={handleToggle}
           />
         ))}
       </List>
@@ -42,7 +44,6 @@ function SearchResultsList(props: SearchResultsListProps) {
 
 interface SearchResultsListProps {
   searchResults: Revision[];
-  handleChildClick: (e: React.MouseEvent) => void;
 }
 
 function mapStateToProps(state: State) {
--- a/src/components/Search/SearchResultsListItem.tsx
+++ b/src/components/Search/SearchResultsListItem.tsx
@@ -10,10 +10,14 @@ import type { Revision } from '../../types/state';
 import { truncateHash, getLastRevision } from '../../utils/searchResultsHelper';
 
 function SearchResultsListItem(props: SearchResultsListItemProps) {
-  const { item, handleChildClick } = props;
+  const { checked, item, handleToggle } = props;
 
   return (
-    <ListItemButton key={item.id}>
+    <ListItemButton
+      key={item.id}
+      id={item.revision}
+      onClick={(e) => handleToggle(e)}
+    >
       <ListItem className="search-revision-item search-revision" disablePadding>
         <ListItemIcon className="search-revision-item-icon search-revision">
           <Checkbox
@@ -21,7 +25,8 @@ function SearchResultsListItem(props: SearchResultsListItemProps) {
             edge="start"
             tabIndex={-1}
             disableRipple
-            onClick={handleChildClick}
+            data-testid={`checkbox-${item.id}`}
+            checked={checked.indexOf(item.revision) !== -1}
           />
         </ListItemIcon>
         <ListItemText
@@ -33,7 +38,6 @@ function SearchResultsListItem(props: SearchResultsListItemProps) {
           sx={{ noWrap: 'true' }}
           primaryTypographyProps={{ noWrap: true }}
           secondaryTypographyProps={{ noWrap: true }}
-          onClick={handleChildClick}
         />
       </ListItem>
     </ListItemButton>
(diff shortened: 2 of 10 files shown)


esanuandra:
image
Might need to add some background color for the list.
The revisions that are overlapping on the table rows can't be selected.

src/tests/Search/SearchResultsList.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+    await user.click(screen.getByTestId('checkbox-5'));
+
+    expect(
+      screen.getByTestId('checkbox-1').classList.contains('Mui-checked'),
+    ).toBe(true);
+    expect(
+      screen.getByTestId('checkbox-5').classList.contains('Mui-checked'),
+    ).toBe(false);
+
+    expect(spyOnConsole).toHaveBeenCalledWith('maximum four revisions');
```
beatrice-acasandrei, line 105: Instead of displaying and testing the message from the console.log(), you could use (either in this PR or another one) the MaterialUI's [Alert ](https://mui.com/material-ui/react-alert/) component to display this message, which should be fairly easy to add.

src/hooks/useCheckRevision.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    handleChildClick(e);
+
+    if (checked.length < 4 && currentIndex === -1) {
+      newChecked.push(revisionID);
+    } else if (currentIndex !== -1) {
+      newChecked.splice(currentIndex, 1);
+    } else {
+      // TODO: replace with pop up notification
+      console.log('maximum four revisions');
```
eng-esther, line 22: [question] shld this removed ?

kimberlythegeek: This is temporary, I've replaced this with a modal in #113

src/components/Search/SearchView.tsx
```diff
@@ -31,15 +30,14 @@ function SearchView(props: SearchViewProps) {
...
-          <Grid item xs={10} sx={{ zIndex: 2 }}>
-            {searchResults.length > 0 && focused && (
-              <SearchResultsList handleChildClick={handleChildClick} />
-            )}
-          </Grid>
-          <Grid item xs={1} />
+      </Grid>
+      <Grid container>
+        <Grid item xs={1} />
+        <Grid item xs={10} sx={{ zIndex: 2 }}>
```
eng-esther, line 36: should this grid item be #2 instead of #10

src/hooks/useCheckRevision.ts
```diff
@@ -0,0 +1,30 @@
...
+    const currentIndex = checked.indexOf(revisionID);
+    const newChecked = [...checked];
+
+    handleChildClick(e);
+
+    if (checked.length < 4 && currentIndex === -1) {
+      newChecked.push(revisionID);
+    } else if (currentIndex !== -1) {
+      newChecked.splice(currentIndex, 1);
+    } else {
```
kimberlythegeek, line 20: replace with else if length == 4

src/components/Search/SearchResultsList.tsx
```diff
@@ -4,12 +4,13 @@ import Box from '@mui/material/Box';
 import List from '@mui/material/List';
 import { connect } from 'react-redux';
 
+import useCheckRevision from '../../hooks/useCheckRevision';
```
canova, line 7: Not for this PR, and not sure if you've talked about this before. But I feel like this `hooks` directory will be pretty crowded in the future when we keep adding new hooks to the codebase. I wonder if we can make that a bit better by grouping the related hooks together. Like with a similar structure to the `components` directory. What do you think?

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -33,16 +38,16 @@ function SearchResultsListItem(props: SearchResultsListItemProps) {
...
           secondaryTypographyProps={{ noWrap: true }}
-          onClick={handleChildClick}
         />
       </ListItem>
     </ListItemButton>
   );
 }
 
 interface SearchResultsListItemProps {
+  checked: string[];
```
canova, line 48: Hmm, is there a reason to send the whole string array here? Could we only pass a boolean `checked` value for this list item instead? I feel like this could isolate the component more.

src/hooks/useCheckRevision.ts
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+import useFocusInput from './useFocusInput';
+
+const useCheckRevision = () => {
+  const { handleChildClick } = useFocusInput();
+  const [checked, setChecked] = React.useState<string[]>([]);
```
canova, line 7: Probably this is not a huge performance win (since this state array will be very small), but we could also use `Set` object here, assuming order of this array is not important.

canova (approved):
Looks good to me, thanks for working on this!
I have some small questions but they are not blockers.

Also I feel like we should be adding those checked revisions to redux store, but I couldn't see if we are doing this. Is this because it's not implemented yet or do you have other plans?

src/components/Search/SearchResultsList.tsx
```diff
@@ -4,12 +4,13 @@ import Box from '@mui/material/Box';
...
 import type { Revision, State } from '../../types/state';
 import SearchResultsListItem from './SearchResultsListItem';
 
 function SearchResultsList(props: SearchResultsListProps) {
-  const { searchResults, handleChildClick } = props;
-
+  const { searchResults } = props;
+  const { checked, handleToggle } = useCheckRevision();
```
canova, line 13: Ah, also keeping this in a redux state would remove this handling from `SearchResultsList` component since it doesn't really need to know which revisions are checked. We could defer this logic directly to `SearchResultsListItem`.

src/components/Search/SearchResultsList.tsx
kimberlythegeek: that's a good idea :) filed #126

src/components/Search/SearchResultsList.tsx
kimberlythegeek: I've actually begun doing that in the PR to add revisions :ok_hand:

src/components/Search/SearchResultsListItem.tsx
kimberlythegeek: after making the changes in the add revision patch and moving checked revisions to the store, this will be removed.

PR #103, opened 2022-05-12
render nothing for perf compare selected revisions if there are no revisions

eng-esther:
Add conditional rendering if there are no revisions

### DIFF
--- a/src/components/Search/SearchView/SearchView.tsx
+++ b/src/components/Search/SearchView/SearchView.tsx
@@ -6,7 +6,6 @@ import { connect } from 'react-redux';
 
 import useFocusInput from '../../../hooks/useFocusInput';
 import { Revision, State } from '../../../types/state';
-import './SearchView.css';
 import PerfCompareHeader from '../../Shared/PerfCompareHeader';
 import AddRevisionButton from '../AddRevisionButton';
 import SearchDropdown from '../SearchDropdown';
@@ -14,12 +13,13 @@ import SearchInput from '../SearchInput';
 import SearchResultsList from '../SearchResultsList';
 import SearchViewInit from '../SearchViewInit';
 import SelectedRevisionsTable from '../SelectedRevisionsTable/SelectedRevisionsTable';
+import './SearchView.css';
 
 function SearchView(props: SearchViewProps) {
   const { focused, handleParentClick, handleFocus, handleChildClick } =
     useFocusInput();
 
-  const { searchResults } = props;
+  const { searchResults, selectedRevisions } = props;
 
   return (
     <Container maxWidth="lg" onClick={handleParentClick}>
@@ -42,7 +42,7 @@ function SearchView(props: SearchViewProps) {
           <Grid item xs={1} />
         </Grid>
         <Grid item xs={12}>
-          <SelectedRevisionsTable />
+          {selectedRevisions.length > 0 && <SelectedRevisionsTable />}
         </Grid>
       </Grid>
     </Container>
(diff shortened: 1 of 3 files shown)


src/components/Search/SelectedRevisionsTable/SelectedRevisionsTable.tsx
```diff
@@ -24,14 +24,13 @@ const tableHeaderDetails: TableHeaderDetails[] = [
...
+  props: SelectedRevisionsProps,
+): JSX.Element | null {
   const { revisions } = props;
   const dispatch = useDispatch();
 
-  if (revisions.length === 0) {
-    return <div className="missingRevisions">Add some revisions</div>;
-  }
-  return (
+  return revisions.length > 0 ? (
```
kimberlythegeek, line 33: Instead of doing the conditional rendering here, it should be done in SearchView, so that we aren't loading this component at all unless there are selected revisions revisions.
See an example here: https://github.com/mozilla/perfcompare/blob/master/src/components/Search/SearchView/SearchView.tsx#L61-L63

kimberlythegeek (changes requested):

src/components/Search/SelectedRevisionsTable/SelectedRevisionsTable.tsx
eng-esther: right now do we have something to keep track of selected revisions?

kimberlythegeek: If you look at the example provided above, it's doing the same thing. If there are no search results (and input is not focused, that bit isn't relevant here), then the results list does not render or mount at all.

You'd be doing basically the same thing in SearchView using mapStateToProps

src/components/Search/SelectedRevisionsTable/SelectedRevisionsTable.tsx
```diff
@@ -24,14 +24,13 @@ const tableHeaderDetails: TableHeaderDetails[] = [
   { id: 5, title: 'Timestamp', alignment: 'left' },
 ];
 
-function SelectedRevisionsTable(props: SelectedRevisionsProps) {
+function SelectedRevisionsTable(
+  props: SelectedRevisionsProps,
+): JSX.Element | null {
```
kimberlythegeek, line 29: once you've changed the render below, I don't think you'll need a type for return value anymore

kimberlythegeek, line 29: once you've changed the render below, I don't think you'll need a type for return value anymore

src/components/Search/SearchView/SearchView.tsx
```diff
@@ -65,7 +65,7 @@ function SearchView(props: SearchViewProps) {
           <Grid item xs={1} />
         </Grid>
         <Grid item xs={12}>
-          <SelectedRevisionsTable />
+          {searchResults.length > 0 && <SelectedRevisionsTable />}
```
kimberlythegeek, line 68: this should be checking the length of selectedRevisions rather than search results

kimberlythegeek: You'll need to add a prop similar to this for selectedRevisions: https://github.com/mozilla/perfcompare/blob/c73eb07f7413eb873b5d6ad9ca340b86fd00d945/src/components/Search/SearchView/SearchView.tsx#L81

eng-esther: true, it seemed strange at first cos it was still rendering, but makes sense. thanks!

src/components/Search/SearchView/SearchView.tsx
```diff
@@ -7,6 +7,7 @@ import Typography from '@mui/material/Typography';
 import { connect } from 'react-redux';
 
 import useFocusInput from '../../../hooks/useFocusInput';
+import SelectedRevisions from '../../../reducers/SelectedRevisions';
```
kimberlythegeek, line 10: I don't think this is needed.

kimberlythegeek (commented):
other than the unused import and the merge conflicts, looks good and works as expected :ok_hand:

kimberlythegeek (approved):
looks good, thanks!

PR #106, opened 2022-05-12
Initial Results View

kimberlythegeek:
based on #105


kimberlythegeek:
renamed all instances of "compare" to "results" since we seem to be using Results to describe this view. I can't rename the branch without closing the PR and opening a new one, but I don't think this matters.

esanuandra (approved):

PR #109, opened 2022-05-12
Remove CSS files and styled components

kimberlythegeek:
The StyledTableRow and StyledTableCell are unnecessary components. While we may use styled components in the future for more complex styles, I was able to easily add these styles to the theme. 

Any other styles were moved inline to the `sx` property. We can address how best to improve this in the future, but now I am still learning how develop custom themes in MUI.

The tableHeaderDetails type was unneccessarily complex to hold titles for the table header, holding IDs as well as the alignment of the cell. 

As of now we are still ironing out the details of our UX/UI design, and we should not focus too much on the CSS, rather the functionality. I replaced this with an enum type to hold just the names of the headers.


esanuandra:
image
image
The suggestions list closes only when we click on these two blue areas. Is this the behavior we want?

src/components/Shared/SelectedRevisionsTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+                sx={{
+                  '.cellStyle': {
+                    fontWeight: 600,
+                    borderRadius: '2.5px',
+                    padding: '5px',
+                    display: 'flex',
+                    color: '#737373',
+                    width: '50px',
+                    border: 'solid 1px #737373',
+                  },
```
esanuandra, line 59: I think we could move this into the `componets.js` MuiTableCell right after the `paddingBottom` at line 15. Inline CSS is not a recommended approach.

kimberlythegeek: thanks!

kimberlythegeek:
> The suggestions list closes only when we click on these two blue areas. Is this the behavior we want?

I've been seeing issues with this also. It also hides (for me) when I click on the header area, but there are spots where it doesn't close when clicked.

I don't think this is a bug introduced in this patch, I see this in master also. I'll file a bug for this. #116

esanuandra:
Looks fine for now. The conflicts need to be fixed.

eng-esther (approved):

Issue #98, opened 2022-05-06
create hooks for useDispatch and useSelector


kimberlythegeek:
replace usage of store.dispatch with useDispatch and store.getState with useSelector

kimberlythegeek:
fixed by #112

PR #112, opened 2022-05-13
Use hooks to access store

kimberlythegeek:
This will be followed up with replacing any calls to `store.dispatch` with `useDispatch`. I've begun working on this several times but I'm still figuring out how to convert everything--hope to get some insight from Julien on this soon.


kimberlythegeek:
Also should note that the tests added here are still using `store.dispatch` to set state.

I'm going to consult with Julien (hopefully next week) to ask about using this in tests. The tests are becoming increasingly slow as more are added. My assumption is that this is fine to use in tests. I would like to remove all of the repeated code, and have each test only testing a specific functionality.

Perhaps there is a way to do this with hooks, but so far I have not found one for redux/redux-toolkit.

kimberlythegeek:
fixes #98

julienw (approved):
I assume there's no behavior change, in that case it looks good to me!
This PR is easy to review, thanks for that :-)

I believe the next step would be to export a store creating function instead of the store object itself, and create an empty store for each test, so that tests are always fully isolated. Happy to discuss about it more later!
I was thinking that as a first step, if you don't want to change existing tests, you could export both `store` and a `createStore()` function, so that new tests would use the creator, but old tests would still work with the object directly. That's up to you really :-)

kimberlythegeek:
> I assume there's no behavior change, in that case it looks good to me! This PR is easy to review, thanks for that :-)
> 
> I believe the next step would be to export a store creating function instead of the store object itself, and create an empty store for each test, so that tests are always fully isolated. Happy to discuss about it more later! I was thinking that as a first step, if you don't want to change existing tests, you could export both `store` and a `createStore()` function, so that new tests would use the creator, but old tests would still work with the object directly. That's up to you really :-)

Thanks! I'll file an issue for this. #115

PR #113, opened 2022-05-13
create reusable modal alert

kimberlythegeek:
Added dismissable alert to Header component. This replaces the `console.log` statement when a user tries to select more than 4 revisions. Currently, because of the way the focus hook works, dismissing the notification will also hide the search results. I'd like to replace this with a timeout, or at least not hide the results somehow

### DIFF
--- a/src/components/Shared/HeaderAlert.tsx
+++ b/src/components/Shared/HeaderAlert.tsx
@@ -1,0 +1,34 @@
+import Alert from '@mui/material/Alert';
+import AlertTitle from '@mui/material/AlertTitle';
+import { connect } from 'react-redux';
+
+import useAlert from '../../hooks/useAlert';
+import type { AlertType, State } from '../../types/state';
+
+function HeaderAlert(props: HeaderAlertProps) {
+  const { alert } = props;
+  const { dispatchClearAlert } = useAlert();
+  return (
+    <Alert
+      severity={alert.severity}
+      onClose={(e: React.SyntheticEvent) => {
+        dispatchClearAlert(e);
+      }}
+    >
+      {alert.title && <AlertTitle>{alert.title}</AlertTitle>}
+      {alert.message}
+    </Alert>
+  );
+}
+
+interface HeaderAlertProps {
+  alert: AlertType;
+}
+
+function mapStateToProps(state: State) {
+  return {
+    alert: state.alert.alert,
+  };
+}
+
+export default connect(mapStateToProps)(HeaderAlert);
--- a/src/hooks/useAlert.ts
+++ b/src/hooks/useAlert.ts
@@ -1,0 +1,26 @@
+import { useDispatch } from 'react-redux';
+
+import type { AppDispatch } from '../common/store';
+import { clearAlert, setAlert } from '../reducers/AlertSlice';
+import type { AlertType } from '../types/state';
+
+const useAlert = () => {
+  const dispatch: AppDispatch = useDispatch();
+
+  const dispatchClearAlert = (e: React.SyntheticEvent) => {
+    e.stopPropagation();
+    dispatch(clearAlert());
+  };
+
+  const dispatchSetAlert = (
+    message: AlertType['message'],
+    severity: AlertType['severity'],
+    title: AlertType['title'] = undefined,
+  ) => {
+    dispatch(setAlert({ message, title, severity }));
+  };
+
+  return { dispatchClearAlert, dispatchSetAlert };
+};
+
+export default useAlert;
--- a/src/hooks/useCheckRevision.ts
+++ b/src/hooks/useCheckRevision.ts
@@ -1,28 +1,29 @@
 import React from 'react';
 
+import useAlert from './useAlert';
 import useFocusInput from './useFocusInput';
 
 const useCheckRevision = () => {
+  const { dispatchSetAlert } = useAlert();
   const { handleChildClick } = useFocusInput();
   const [checked, setChecked] = React.useState<string[]>([]);
 
   const handleToggle = (e: React.MouseEvent) => {
     const revisionID = (e.currentTarget as HTMLElement).id;
-    const currentIndex = checked.indexOf(revisionID);
+    const isChecked = checked.includes(revisionID);
     const newChecked = [...checked];
 
-    handleChildClick(e);
-
-    if (checked.length < 4) {
-      // if item is not already checked, add to checked
-      if (currentIndex === -1) newChecked.push(revisionID);
+    // if item is not already checked, add to checked
+    if (checked.length < 4 && !isChecked) {
+      newChecked.push(revisionID);
+    } else if (isChecked) {
       // if item is already checked, remove from checked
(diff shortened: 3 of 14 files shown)


kimberlythegeek:
blocked by #102

src/hooks/useCheckRevision.ts
```diff
@@ -0,0 +1,31 @@
...
+    const newChecked = [...checked];
+
+    handleChildClick(e);
+
+    if (checked.length < 4 && currentIndex === -1) {
+      newChecked.push(revisionID);
+    } else if (currentIndex !== -1) {
+      newChecked.splice(currentIndex, 1);
+    } else {
+      useSetAlert('Maximum 4 Revisions', 'warning');
```
eng-esther, line 23: is this alert timed? or it will remain on the page until it has been addressed?

eng-esther, line 23: is this alert timed? or it will remain on the page until it has been addressed?

kimberlythegeek: as it is now, it will stay until dismissed.

kimberlythegeek:
> After the alert appears I can't uncheck the checked revisions from the list.

ah--thanks for catching that. I modified the if/else statement and apparently did it wrong :) should be fixed now. Also added a test case to catch this.

src/hooks/useAlert.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  const useSetAlert = (
+    message: AlertType['message'],
+    severity: AlertType['severity'],
+    title: AlertType['title'] = undefined,
+  ) => {
+    dispatch(setAlert({ message, title, severity }));
+  };
+
+  return { useClearAlert, useSetAlert };
```
julienw, line 22: I find the names a bit confusing: they look like they are hooks themselves, but they actually dispatch an action.
I would use simply `clearAlert` and `setAlert`, or more explicit `dispatchClearAlert` and `dispatchSetAlert`. (but I prefer the first option).

src/hooks/useAlert.ts
```diff
@@ -0,0 +1,25 @@
+import { useDispatch } from 'react-redux';
+
+import type { AppDispatch } from '../common/store';
+import { clearAlert, setAlert } from '../reducers/AlertSlice';
+import type { AlertType } from '../types/state';
+
+const useAlert = () => {
```
julienw, line 7: optional comment: is this hook really useful? I feel like it's adding more code than it removes.

src/hooks/useCheckRevision.ts
```diff
@@ -11,18 +13,17 @@ const useCheckRevision = () => {
     const currentIndex = checked.indexOf(revisionID);
```
julienw, line 13: optional: It would be clearer to have a boolean if you dont use the index itself. You can use `const isAlreadyChecked = checked.includes(revisionID)` for example.

src/hooks/useCheckRevision.ts
```diff
@@ -1,8 +1,10 @@
 import React from 'react';
 
+import useAlert from './useAlert';
 import useFocusInput from './useFocusInput';
 
 const useCheckRevision = () => {
+  const { useSetAlert } = useAlert();
   const { handleChildClick } = useFocusInput();
   const [checked, setChecked] = React.useState<string[]>([]);
```
julienw, line 9: optional: I wonder if using a `Set<string>` could be a better fit, seeing the operations you're doing here.

src/tests/PerfCompareHeader.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  it('should hide alert when close button is clicked', async () => {
+    // set delay to null to prevent test time-out due to useFakeTimers
+    const user = userEvent.setup({ delay: null });
+    title = undefined;
+
+    store.dispatch(setAlert({ message, severity, title }));
+    render(<PerfCompareHeader />);
+
+    expect(screen.getByText('Maximum 4 Revisions')).toBeInTheDocument();
+    expect(screen).toMatchSnapshot();
```
julienw, line 28: Usually I use `document.body` to match.
Looking at the new snapshot, I think that using `screen` doesn't work as you expect :-)

src/components/Shared/HeaderAlert.tsx
```diff
@@ -0,0 +1,34 @@
...
+  const { useClearAlert } = useAlert();
+  return (
+    <Alert
+      severity={alert.severity}
+      onClose={() => {
+        useClearAlert();
+      }}
+    >
+      {alert.title && <AlertTitle>{alert.title}</AlertTitle>}
+      {alert.message}
```
julienw, line 19: What happens if you have `isAlert === true` but both title and message are empty? Should we display some default message?

src/components/Shared/HeaderAlert.tsx
```diff
@@ -0,0 +1,34 @@
...
+import useAlert from '../../hooks/useAlert';
+import type { AlertType, State } from '../../types/state';
+
+function HeaderAlert(props: HeaderAlertProps) {
+  const { alert } = props;
+  const { useClearAlert } = useAlert();
+  return (
+    <Alert
+      severity={alert.severity}
+      onClose={() => {
```
julienw, line 14: It would probably be a good idea to call `e.stopPropagation()` on the event passed to the callback, so that the close operation doesn't trigger other things (eg: so that it doesn't close the select box)

julienw (approved):
I left a few comments, some of them are optional too, so tell me what you think!

src/components/Shared/HeaderAlert.tsx
kimberlythegeek: `title` is the only parameter with a default value, so if you try to call `useSetAlert` without a message, it will throw an error.

I could add a default message though, as there may be a use case in the future where an unexpected error occurs.

src/components/Shared/HeaderAlert.tsx
kimberlythegeek: thanks!

src/hooks/useAlert.ts
kimberlythegeek: I guess I don't have a strong argument for it now, but ideally in the future we will be able to display more than one alert at a time, and this hook would be used to handle those actions.

src/hooks/useAlert.ts
kimberlythegeek: I can't use `clearAlert` and `setAlert` as those are the names of the reducer actions, but I can change it to `dispatch` instead of `set`

src/hooks/useCheckRevision.ts
kimberlythegeek: I'll leave this as is for now, and file an issue to consider using Sets instead. it seems like it's worthwhile when handling many items, but this array will never have more than 4 elements. at least for while :) #137

julienw: Yeah, to be clear, this wasn't a performance-oriented reason, more a simplification of the code, where you wouldn't need to check for existence before pushing :-)

PR #117, opened 2022-05-25
Replace "Add Revisions" text with icon

kimberlythegeek:


eng-esther (approved):
Not sure if this already added but might need happy path code coverage of if icon clicked the right call back is called..

PR #118, opened 2022-05-26
clean up linting file and fix errors

kimberlythegeek:
I noticed that the linter was not always warning about missing types. When I first updated the eslint config to lint typescript files, I wasn't sure how to also correctly lint javascript files, so I added overrides for typescript files. This config is much cleaner and should correctly catch typescript errors.


.eslintignore
```diff
@@ -0,0 +1,3 @@
+.eslintrc.js
+src/tests/utils/test-utils.jsx
```
kimberlythegeek, line 2: I've tried making this a tsx file more than once, but haven't yet found the solution to do so. This file will likely not change, so instead of adding several rules to ignore, I've ignored the file.

kimberlythegeek: failing rules were:
```
'children' is missing in props validation               react/prop-types
Unsafe assignment of an `any` value        @typescript-eslint/no-unsafe-assignment
Unsafe argument of type `any` assigned to a parameter of type `ReactElement<any, string | JSXElementConstructor<any>>`  @typescript-eslint/no-unsafe-argument
'@testing-library/react' should be listed in the project's dependencies, not devDependencies import/no-extraneous-dependencies
 ```

File: .eslintignore [hunk 2]
```diff
@@ -0,0 +1,3 @@
...
+src/tests/utils/fileTransformer.js
```
kimberlythegeek, line 3: same here. failing rules for these two files were: 
```
Require statement not part of import statement                          @typescript-eslint/no-var-requires
Unsafe argument of type `any` assigned to a parameter of type `string`  @typescript-eslint/no-unsafe-argument
```

src/components/Search/SearchViewInit.tsx
```diff
@@ -1,21 +1,18 @@
...
 
-import { store } from '../../common/store';
 import { fetchRecentRevisions } from '../../thunks/searchThunk';
 import { Repository, State } from '../../types/state';
 
 // component to fetch recent revisions when search view is loaded
-class SearchViewInit extends React.Component<SearchViewInitProps> {
-  componentDidMount() {
-    const { repository } = this.props;
-    store.dispatch(fetchRecentRevisions(repository));
```
kimberlythegeek, line 13: I was getting an error for 'promises must be awaited...etc', but the promise is awaited in `fetchRecentRevisions` with a `catch` block to catch errors.

src/components/Search/SearchViewInit.tsx
```diff
@@ -1,21 +1,18 @@
...
-  }
-
-  render() {
-    return null;
-  }
+function SearchViewInit(props: SearchViewInitProps) {
+  const dispatch = useDispatch();
+  const { repository } = props;
+  React.useEffect(() => {
```
kimberlythegeek, line 12: converting this to a functional component and using `useEffect` instead of `componentDidMount` seemed to resolve this. this should have the same effect, only it is called after render instead of before: https://reactjs.org/docs/hooks-reference.html#useeffect

src/tests/Search/SearchView.test.tsx
```diff
@@ -91,6 +91,9 @@ describe('Search View', () => {
     await screen.findByText(
       'Search must be a 12- or 40-character hash, or email address',
     );
+
+    // fetch should only be called on initial load
+    expect(spyOnFetch).toHaveBeenCalledTimes(1);
```
kimberlythegeek, line 96: received an error that this test had no assertions, so I added this one.

src/tests/Search/SearchView.test.tsx
```diff
@@ -183,27 +186,6 @@ describe('Search View', () => {
...
-    await user.type(searchInput, 'abcdef1234567890abcdef1234567890abcdef12');
-    jest.runOnlyPendingTimers();
-
-    await screen.findByText('An error has occurred');
-    expect(store.getState().search.inputError).toBe(true);
-    expect(store.getState().search.inputHelperText).toBe(
-      'An error has occurred',
-    );
-  });
-
```
kimberlythegeek, line 206: duplicate test caught by linter

src/tests/SelectedRevisionsTable.test.tsx
```diff
@@ -16,16 +18,18 @@ describe('Search View', () => {
...
     const dispatchSpy = jest.spyOn(store, 'dispatch');
+    const user = userEvent.setup({ delay: null });
 
-    const { rerender } = render(<SelectedRevisionsTable />);
+    // const { rerender } = render(<SelectedRevisionsTable />);
+    render(<SelectedRevisionsTable />);
     const button = document.querySelectorAll('#close-button');
 
-    await (button[0] as HTMLButtonElement).click();
-    await (button[1] as HTMLButtonElement).click();
```
kimberlythegeek, line 24: received this error: `Unexpected `await` of a non-Promise (non-"Thenable") value  @typescript-eslint/await-thenable` so I replaced it with userEvent

src/tests/utils/test-utils.jsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import { render as rtlRender } from '@testing-library/react';
 import { Provider } from 'react-redux';
 
-import { store } from '../../common/store.ts';
```
kimberlythegeek, line 6: lint error 'Unexpected use of file extension'

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "allowJs": true, // allow a partial TypeScript and JavaScript codebase
     "esModuleInterop": true,
-    "jsx": "react", // use typescript to transpile jsx to js
+    "jsx": "react-jsx", // use typescript to transpile jsx to js
```
kimberlythegeek, line 5: added to resolve `'React' was imported but never used`

src/tests/SelectedRevisionsTable.test.tsx
```diff
@@ -16,17 +18,16 @@ describe('Search View', () => {
 
   it('should delete revisions after click and not show revisions table if no revisions', async () => {
     const dispatchSpy = jest.spyOn(store, 'dispatch');
+    const user = userEvent.setup({ delay: null });
 
-    const { rerender } = render(<SelectedRevisionsTable />);
```
kimberlythegeek, line 20: I was unsure what the purpose of this was, so I tried removing it and the test still worked as expected. If there's justification for keeping this I can revert this change.

canova:
@kimberlythegeek I wasn't able to run the `lint` command successfully with this PR. I get this error when I run `npm run lint`:
```
> perfcompare@0.1.0 lint
> eslint src/**/*.{js,jsx,ts,tsx}


Oops! Something went wrong! :(

ESLint: 8.12.0

You are linting "src/**/*.jsx", but all of the files matching the glob pattern "src/**/*.jsx" are ignored.

If you don't want to lint these files, remove the pattern "src/**/*.jsx" from the list of arguments passed to ESLint.

If you do want to lint these files, try the following solutions:

* Check your .eslintignore file, or the eslintIgnore property in package.json, to ensure that the files are not configured to be ignored.
* Explicitly list the files from this glob that you'd like to lint on the command-line, rather than providing a glob as an argument.

```

I think this is because we have only one jsx file which is the ignored `src/tests/utils/test-utils.jsx`. Since it can't find another jsx file anymore, it's throwing this error. Easiest solution would be to remove the jsx from `eslint src/**/*.{js,jsx,ts,tsx}` command. It might be a problem if we intend to add jsx files in the future, but I doubt it.

kimberlythegeek:
@canova weird--I don't see that error. I wonder if it's a platform issue. I'll try running on my mac also, but I will go ahead and remove this.

canova (approved):
Nice, great that we fixed some linting errors now! Looks good to me! I've added two small comments and one follow-up. Let me know what you think!

I've also created #124 for converting `test-utils.jsx` to tsx and we have #125 filed now for removing `fileTransformer.js` file. With them, I believe we won't need the `.eslintignore` file anymore.

.eslintrc.js
```diff
@@ -1,32 +1,34 @@
 module.exports = {
-  env: {
-    browser: true,
-    es2021: true,
-    jest: true,
-  },
-  extends: [
-    'eslint:recommended',
```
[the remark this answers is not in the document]
kimberlythegeek: is it needed if we have this? `'plugin:@typescript-eslint/recommended'`

.eslintignore
```diff
@@ -0,0 +1,3 @@
+.eslintrc.js
```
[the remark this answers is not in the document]
kimberlythegeek: I don't get any linting errors, but my VS Code formatter does highlight this as error, which is kind of annoying. My default formatter is Prettier. any idea how I can resolve this?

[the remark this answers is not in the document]
canova: Intereting, I haven't checked the vscode formatter, but indeed I see the error now.

Ok, I've investigated a bit more and it looks like all the .js files are not working with linter anymore and not only `eslintrc.js`. (for example the same error exists in `webpack.config.js` and other js files. They are not caught with the`lint` script because it's only running the eslint inside `src` directory: `eslint src/**/*.{js,ts,tsx}`.

So we should figure out a generic way to fix this instead of adding them to the gitignore file.
I've figured out a working solution but it involves a few changes:

```diff
diff --git a/.eslintrc.js b/.eslintrc.js
index ac7ebf0..fb9bbd6 100644
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,21 +1,20 @@
 module.exports = {
   root: true,
+  "env": {
+    "browser": true,
+    "node": true
+  },
   parser: '@typescript-eslint/parser',
   parserOptions: {
-    tsconfigRootDir: __dirname,
-    project: ['./tsconfig.json'],
     ecmaFeatures: {
       jsx: true,
     },
   },
   plugins: ['@typescript-eslint', 'import', 'jest', 'react'],
   extends: [
-    'plugin:@typescript-eslint/recommended',
-    'plugin:@typescript-eslint/recommended-requiring-type-checking',
     'plugin:react/recommended',
     'plugin:react/jsx-runtime',
     'plugin:jest/recommended',
-    'airbnb-typescript',
     'prettier',
   ],
   settings: {
@@ -52,6 +51,24 @@ module.exports = {
     'react/jsx-uses-react': 'error',
   },
   overrides: [
+    {
+      files: ['*.ts', '*.tsx'],
+      // We should extend TypeScript plugins here instead of extending them
+      // outside the `overrides`.
+      extends: [
+        'plugin:@typescript-eslint/recommended',
+        'plugin:@typescript-eslint/recommended-requiring-type-checking',
+        'airbnb-typescript',
+      ],
+      parserOptions: {
+        tsconfigRootDir: __dirname,
+        project: ['./tsconfig.json'],
+      },
+      rules: {
+        // https://github.com/typescript-eslint/typescript-eslint/issues/1824
+        '@typescript-eslint/indent': 'off',
+      },
+    },
     {
       // typescript test files
       files: ['src/tests/**/*.{ts,tsx}'],
```
These changes are what makes it work for me with the js files as well. There are a few things to unpack though:
1. We need to add `env` object here because we need to tell eslint that this code runs on both web **and** node. The "node" field here is critical because otherwise `module.exports` in this file will give error thinking that there is no `modules` global variable (which only exists on node).
2. We need to create another override here for ts and tsx files unfortunatly. I know this makes the code a bit more crowded, but this is actually needed to isolate ts and tsx file specific rules. We are also moving the typescript related `extends` here.
3. Most importantly `project: ['./tsconfig.json'],` field should be inside the override as well. This is telling linter that only typescript files are subject to tsconfig.json rules. Otherwise eslint will try to assume other .js files are also typescript.
4. Lastly we have `'@typescript-eslint/indent': 'off',` rule here because of a known eslint bug :( This was there before actually. We didn't need it in the new version because we didn't have .ts specific rules, but that problem is back again.

.eslintrc.js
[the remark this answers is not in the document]
canova: These are `eslint:recommended` rules: https://github.com/eslint/eslint/blob/main/conf/eslint-recommended.js
And these are `plugin:@typescript-eslint/recommended` rules: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/eslint-recommended.ts

`eslint:recommended` looks more generic and it looks like it's enabling more rules.

Also note the comment at the top of typescript one:
> - disables rules from eslint:recommended which are already handled by TypeScript.

So it overrides some rules of `eslint:recommended`. The order of them seems important too. We should specify the `eslint:recommended` first and then `plugin:@typescript-eslint/recommended`.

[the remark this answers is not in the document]
canova: Ah, also documentation mentiones that they need to be used together: https://github.com/typescript-eslint/typescript-eslint/blob/main/docs/linting/CONFIGS.md#eslint-recommended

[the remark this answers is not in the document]
julienw: I don't see `estlint:recommended` has been added back as nazim suggested, it would be good to do it in a follow-up :-)

.eslintrc.js
```diff
@@ -1,32 +1,33 @@
...
-    sourceType: 'module',
   },
-  plugins: ['react'],
+  plugins: ['@typescript-eslint', 'import', 'jest', 'react'],
+  extends: [
+    'plugin:react/recommended',
+    'plugin:react/jsx-runtime',
+    'plugin:jest/recommended',
+    'prettier',
+  ],
```
julienw, line 19: Shouldn't the `jest` entries in `plugins` and `extends` be added for tests only?

.eslintrc.js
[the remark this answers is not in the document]
kimberlythegeek: > I don't see `estlint:recommended` has been added back as nazim suggested, it would be good to do it in a follow-up :-)

ah, thanks for catching that. I thought I had added it.

PR #120, opened 2022-05-26
move selected revisions table to fix overlap

kimberlythegeek:
I was sure I had styled the components properly to overlap the search results over the selected revisions and avoid pushing it down, but after merging I realized it was not properly styled, perhaps I made an error while rebasing, I'm not sure.

I could revert my previous PR, or move the table back above the input. I know it was my suggestion to move it below to begin with, but I didn't realize it would be so much work to overlay the results.

Another issue with properly overlapping is that it requires setting `position: absolute`, and this makes it difficult to keep the table centered and the desired width.


eng-esther:
is the table now at the top and the drop down at the bottom?

beatrice-acasandrei (approved):

esanuandra (approved):

Issue #110, opened 2022-05-13
Improve UI of Selected Revisions


kimberlythegeek:
Improve the UI of selected revisions table to more closely match Mozilla's protocol theme: https://protocol.mozilla.org/

(edit) this is less of a priority than the points listed below

- [ ] truncate hash
- [ ] truncate commit message
- [ ] limit table width so there is no side scrolling
- [ ] map repository ID to names
- [ ] convert timestamp to date

kimberlythegeek:
If the commit message is very long, it changes the height of the row, so this should be truncated, as well as the hash (similar to what was done in the search results).

It should be a fixed width so there is no horizontal scrolling. the timestamps should be converted to dates. the repository IDs should be matched to repository names.

kimberlythegeek:
fixed by #145

PR #121, opened 2022-05-26
implement add revision

kimberlythegeek:


kimberlythegeek:
Also it should be noted that the appearance of the table needs to be fixed. If the commit message is very long, it changes the height of the row, so this should be truncated, as well as the hash (similar to what was done in the search results).

It should be a fixed width so there is no horizontal scrolling. the timestamps should be converted to dates. the repository IDs should be matched to repository names. I've filed #110 for this and will add my comments there

julienw (changes requested):
Thanks, it's very good to see how the project moves forward!

I'm marking this PR as "request changes" for 2 main reasons:
* a performance reason due to the fact that all list items will be rerendered for each new check
* an architecture reason because I think some code should be moved to reducers instead of hooks

But happy to discuss about this!

src/reducers/SelectedRevisions.tsx
```diff
@@ -48,6 +13,9 @@ const selectedRevisions = createSlice({
     resetState(state) {
       state.revisions = initialState.revisions;
     },
+    setSelectedRevisions(state, action: PayloadAction<Revision[]>) {
+      state.revisions = action.payload;
+    },
```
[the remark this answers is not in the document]
eng-esther: we can try, but I think this is fine the way it is.

src/hooks/useSelectRevision.ts
```diff
@@ -0,0 +1,49 @@
...
+      } else if (currentIndex !== -1) {
+        // TODO: replace with pop up notification
+        console.log('revision already selected');
+      } else {
+        // TODO: replace with pop up notification
+        console.log('maximum four revisions');
+      }
+    });
+    dispatch(setSelectedRevisions(newSelected));
+    dispatch(clearCheckedRevisions());
```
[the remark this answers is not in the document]
kimberlythegeek: I would not be able to dispatch `setAlert` if I moved this code to the reducer

src/types/state.ts
```diff
@@ -42,6 +42,12 @@ export type SearchState = {
   inputHelperText: string;
 };
 
+// contains the indices of currently checked revisions
+// in searchResults state
+export type CheckedRevisionsState = {
+  revisions: number[];
```
eng-esther, line 48: should this be 
```suggestion
  revisions: Revision[];
```

src/tests/SelectedRevisionsTable.test.tsx
```diff
@@ -29,4 +60,68 @@ describe('Search View', () => {
...
+    const { testData } = getTestData();
+    global.fetch = jest.fn(() =>
+      Promise.resolve({
+        json: () => ({
+          results: testData,
+        }),
+      }),
+    ) as jest.Mock;
+    // set delay to null to prevent test time-out due to useFakeTimers
+    const user = userEvent.setup({ delay: null });
```
eng-esther, line 74: [nitpick] I noticed  this block occurs in 3 it blocks.
would it make sense to have a before each shared by all three?

src/types/state.ts
kimberlythegeek: This state holds the index of a revision in searchResults. I can rename this to revisionIndices to make this more clear.

eng-esther (approved):

src/tests/SelectedRevisionsTable.test.tsx
julienw: The testing library folks advise againse using the `beforeEach` thingie because they obscure what a test is doing.

However it might make sense to extract them to a common function named for example `setup` with common setup code.

kimberlythegeek:
@julienw 
> I feel bad that nearly everything I suggested didn't work out in the end, sorry about that. Thanks a ton for trying all of them.

no worries! I appreciate the feedback and enjoy learning

> Some problems I found while testing, but can be handled in follow-ups: 1. image
> 
> Probably a z-index to adjust?
> 
>     2. Also we don't see the warning once we're scrolled down (this happen when 4 revisions are already added for example). Maybe we need it as `position: fixed`. (that might fix the first issue too).

I've filed #139 to fix this UI, and also found that MaterialUI does have a timed alert component called SnackBar, which I will include in that patch

>     3. I see "4" in the "Project" column after I added revisions, I'm not sure what this means.

Currently we only show the repository_id, but I will be addressing this in #110

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -3,23 +3,29 @@ import ListItem from '@mui/material/ListItem';
...
   truncateHash,
   getLatestCommitMessage,
 } from '../../utils/searchResultsHelper';
 
 function SearchResultsListItem(props: SearchResultsListItemProps) {
-  const { checked, item, handleToggle } = props;
+  const { index, item, handleToggle } = props;
+  const isChecked: boolean = useSelector((state: RootState) =>
+    state.checkedRevisions.revisions.includes(index),
+  );
```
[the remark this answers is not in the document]
kimberlythegeek: #140

PR #123, opened 2022-05-27
add button to enable nav into results page

eng-esther:
Screen Shot 2022-05-26 at 8 40 01 PM


julienw:
Hey @eng-esther, I see this PR has conflicting files now. Could you please rebase it first, before I can review it? I believe you could do these steps:
1. squash everything to only 1 single commit. Ideally when there are several commits for one PR, this is so that each commit introduce one single small independant change, and all commits together implement the more complex feature. The goal is to make a complex PR easier to review by splitting it in independant smaller code changes. (Another strategy is splitting a big PR in smaller PRs, but that's not always possible).
  Here I see that your commits are not independant changes, but rather the later commits fix mistakes introduced by previous commits. That's why I think it makes more sense to squash everything together in this case.
2. rebase from this single commit, which should probably be easier.

Sorry I couldn't review earlier, because I was away for a few days.

eng-esther:
messy rebase

PR #124, opened 2022-05-27
Convert the test-utils.jsx file to a tsx file

canova:
With this change we remove the last jsx file from the codebase. We also had to remove the `jsx` extension from the glob pattern inside `lint` and `lint:fix` commands. This is becasue `eslint` will no longer allow us to run the command with that extension since we don't have any files with that anymore.

Tested locally and `npm run start-dev` and `npm run lint` passes successfully.

Also it would be good to look at the PR commit by commit. First commit only does the renaming of the file. Second commit adds the types. Wanted to seperate them so it would be clear which lines have changed.


src/tests/utils/test-utils.tsx
```diff
@@ -5,8 +5,8 @@ import { Provider } from 'react-redux';
 
 import { store } from '../../common/store.ts';
 
-function render(ui) {
-  function Wrapper({ children }) {
+function render(ui: React.ReactElement) {
```
canova, line 8: This type is coming from:
https://github.com/testing-library/react-testing-library/blob/f176285e4e92754751b708e1b1adf1f38edea6a8/types/index.d.ts#L97

src/tests/utils/test-utils.tsx
```diff
@@ -5,8 +5,8 @@ import { Provider } from 'react-redux';
...
+  function Wrapper({ children }: { children: React.ReactElement }) {
```
canova, line 9: This type is coming from: 
https://github.com/testing-library/react-testing-library/blob/f176285e4e92754751b708e1b1adf1f38edea6a8/types/index.d.ts#L135

kimberlythegeek (approved):
Awesome! thank you! I checked this against the linting pr #118 and only found one error that already existed in a test file, I can address that in my patch.

canova:
Hey @kimberlythegeek, I've just rebased the PR on top of the latest master that included PR #118. I think I can land it now. But you also said that there was an error before with a test. Is this resolved now? I actually checked this PR by rebasing on top of yours before and I didn't see any error back then. Now I'm a bit wary of landing it just in case if I don't see the error and you do. Do you think it's okay to land it now?

kimberlythegeek:
> Hey @kimberlythegeek, I've just rebased the PR on top of the latest master that included PR #118. I think I can land it now. But you also said that there was an error before with a test. Is this resolved now? I actually checked this PR by rebasing on top of yours before and I didn't see any error back then. Now I'm a bit wary of landing it just in case if I don't see the error and you do. Do you think it's okay to land it now?

Yep! go for it--I resolved it in the linting PR

canova:
Thanks!

PR #130, opened 2022-06-01
update lint config to have no default parser

kimberlythegeek:


.eslintrc.js
```diff
@@ -12,6 +12,7 @@ module.exports = {
   },
   plugins: ['@typescript-eslint', 'import', 'jest', 'react'],
   extends: [
+    'eslint:recommended',
     'plugin:react/recommended',
     'plugin:react/jsx-runtime',
     'plugin:jest/recommended',
```
canova, line 18: We still have the `plugin:jest/recommended` extension and `jest` plugin in the global scope. Can we remove them as well?

julienw:
I removed my r request because I see Nazim is on it already :-)

eng-esther (approved):

PR #131, opened 2022-06-01
Rework Search Results Helper Methods

kimberlythegeek:


eng-esther (approved):
will some basic test be added in another patch?

kimberlythegeek:
@eng-esther 
> will some basic test be added in another patch?

thanks for mentioning this--add a couple of tests, and found that it was actually return an 11 character string because `slice` does not include the last index

PR #132, opened 2022-06-01
Remove `eject` script and change `lint:fix` to use `lint` directly in package.json

canova:
As discussed, we don't really need the `eject` script in the package.json file for now.
I've also changed the `lint:fix` to use `lint` directly instead. Reusing the other command makes it easier to keep them in sync when we need to change these commands.


kimberlythegeek:
Thanks @canova !

PR #135, opened 2022-06-01
use swc-jest in place of babel and ts-jest

### LINKED ISSUE #134: investigate improving build and test run times
Julien suggested using babel for building to improve build times and test run times. We may be running type-checking several times and causing these slow build times.

kimberlythegeek:
Using this library significantly improves test run time. fixes #134


julienw (approved):
This is pretty cool :-)

This works well for me too, down to 3 seconds from 11.

PR #143, opened 2022-06-08
Create new store for each test

### LINKED ISSUE #115: export store creating function for tests
from Julien:
> I believe the next step would be to export a store creating function instead of the store object itself, and create an empty store for each test, so that tests are always fully isolated. Happy to discuss about it more later! I was thinking that as a first step, if you don't want to change existing tests, you could export both `store` and a `createStore()` function, so that new tests would use the creator, but old tests would still work with the object directly. That's up to you really :-)

kimberlythegeek:
fixes #115

added function to create new store for each test.

because we are using a helper function to create a new render and wrap UI components with a provider, I also had to add functions to create a new render and renderWithRouter for each test.

### DIFF
--- a/src/tests/utils/test-utils.tsx
+++ b/src/tests/utils/test-utils.tsx
@@ -6,14 +6,13 @@ import { SnackbarProvider } from 'notistack';
 import { Provider } from 'react-redux';
 import { Router } from 'react-router-dom';
 
-import { store } from '../../common/store';
+import type { Store } from '../../common/store';
 import SnackbarCloseButton from '../../components/Shared/SnackbarCloseButton';
 
-function render(ui: React.ReactElement) {
-  function Wrapper({ children }: { children: React.ReactElement }) {
-    return (
-      <Provider store={store}>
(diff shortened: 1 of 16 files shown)


beatrice-acasandrei (approved):
lgtm

src/tests/utils/test-utils.tsx
```diff
@@ -6,48 +6,53 @@ import { SnackbarProvider } from 'notistack';
 import { Provider } from 'react-redux';
 import { Router } from 'react-router-dom';
 
-import { store } from '../../common/store';
+import { Store } from '../../common/store';
```
julienw, line 9: I believe that it's better to use `import type` when importing types only, so that we're more guaranteed that it's not kept after transpiling. I believe this is more a problem with babel than with tsc.
see https://www.typescriptlang.org/docs/handbook/modules.html#importing-types for more information

julienw (approved):
lgtm, thanks :) I believe that moving forward things will be easier to work with with this simplification

PR #149, opened 2022-06-10
Add state and reducer for results table

kimberlythegeek:
blocked by #148


src/components/Results/ResultsTable.tsx
```diff
@@ -55,33 +36,31 @@ function ResultsTable() {
...
+              newRuns,
+            } = result;
+            return (
+              <TableRow key={index}>
+                <TableCell>{platformName}</TableCell>
+                <TableCell>{testName}</TableCell>
+                <TableCell>{baseValue}</TableCell>
+                <TableCell>{newValue}</TableCell>
+                <TableCell>{delta}%</TableCell>
+                <TableCell>{confidenceText}</TableCell>
```
eng-esther, line 57: [suggestion] do you want to use a map to iterativelly generate the cells?

kimberlythegeek: I considered this, but was unsure what we will be doing in the future, so mapping may not work. example, for delta, when we have `is_improvement` and `is_regression`, we'll likely use conditional rendering to display either a check icon or x icon.

src/common/store.ts
```diff
@@ -2,12 +2,14 @@ import { configureStore } from '@reduxjs/toolkit';
...
 
 import checkedRevisions from '../reducers/CheckedRevisions';
+import results from '../reducers/ResultsSlice';
 import search from '../reducers/SearchSlice';
 import selectedRevisions from '../reducers/SelectedRevisions';
 
 const reducer = combineReducers({
-  search,
   checkedRevisions,
+  results,
```
eng-esther, line 11: [question] would we have other types of results?
maybe this should be renamed to compareResults ? 🤷‍♀️

eng-esther (approved):

src/reducers/CompareResultsSlice.ts
```diff
@@ -0,0 +1,64 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+// TODO: replace with thunk that accesses endpoint
+// import { fetchCompareResults } from '../thunks/fetchCompareResults';
+import type { CompareResultsState } from '../types/state';
+
+const initialState: CompareResultsState = [
```
julienw, line 7: (optional) A comment stating that they are test values could be useful

src/types/state.ts
```diff
@@ -39,6 +52,8 @@ export type SelectedRevisionsState = {
   revisions: Revision[];
 };
 
+export type CompareResultsState = CompareResultsItem[];
```
julienw, line 55: Don't you need to add `compareResults` below to the `State` too?

Looking at https://github.com/mozilla/perfcompare/pull/149/files#diff-c1cf636d61418b4439ce697bd7f4249145469dacd55453c8832c045f1a99f5cbR9-R14 it seems that some other properties are missing too. Is that expected that TypeScript doesn't error about that?

kimberlythegeek: I originally was using this `State` type, but at some point added `RootState` that infers the type: https://github.com/mozilla/perfcompare/blob/master/src/common/store.ts#L21

I will file an issue to remove `State` and replace with `RootState`

my guess on why it isn't erroring is that the only place `State` is used, it is not accesing the missing state types: https://github.com/mozilla/perfcompare/pull/149/files#diff-f4aeacfb95826ff4c36316d81101147c633ca4a49655531c49c6dff7d4767970R33

PR #154, opened 2022-06-15
add types for results to match API response

kimberlythegeek:


src/types/enums.ts
```diff
@@ -15,3 +15,232 @@ export type CompareResultsTableHeaders =
   | 'Total Runs';
 
 export type ConfidenceText = 'high' | 'med' | 'low';
```
eng-esther, line 17: [suggestion] I see you use med here but not updated the mock data to use medium.

src/types/enums.ts
```diff
@@ -15,3 +15,232 @@ export type CompareResultsTableHeaders =
...
+
+export type MeasurementUnit =
+  | 'W'
```
eng-esther, line 20: [question] do we need the | before the first measurement type?

src/types/enums.ts
eng-esther: will need to confirm from Beatrice if it's `med` or `medium`

src/common/constants.ts
```diff
@@ -13,4 +16,235 @@ export const repoMapping: Dictionary<Repository['name']> = {
...
+  10: 'raptor',
+  11: 'js-bench',
+  12: 'devtools',
+  13: 'browsertime',
+  15: 'mozperftest',
+  16: 'fxrecord',
+};
+
+export const platformMap: Record<Platform, string> = {
+  linux32: 'Linux',
```
eng-esther, line 33: [question] should linux 32 be a string?

eng-esther (approved):

src/types/enums.ts
kimberlythegeek: 👌🏻 I think in perfherder it shows 'med' but yes we need to confirm this.

esanuandra (approved):

PR #158, opened 2022-06-16
rename compareResults test

kimberlythegeek:


kimberlythegeek:
added this in previous PR #154

PR #163, opened 2022-06-18
add dark mode

kimberlythegeek:
`//TODO: change color of OS icons based on mode`

added dark mode with toggle


src/theme/protocolTheme.ts
```diff
@@ -0,0 +1,29 @@
...
+// https://protocol.mozilla.org/
+import { useMemo, useState } from 'react';
+
+import { createTheme, Theme } from '@mui/material/styles';
+
+import components from './components';
+import typography from './typography';
+
+const useProtocolTheme = () => {
+  const [mode, setMode] = useState<'light' | 'dark'>('light');
```
kimberlythegeek, line 11: this seemed like the easiest/most logical place to put this hook. the example in the docs I believe creates a new component and wraps the `App` component. I think this approach is cleaner.

src/theme/typography.ts
```diff
@@ -1,8 +1,9 @@
+import type { TypographyOptions } from '@mui/material/styles/createTypography';
```
kimberlythegeek, line 1: was receiving type errors when using `typography.js` after converting `protocolTheme` to a typescript file

beatrice-acasandrei (approved):

PR #166, opened 2022-06-22
Add parameters to url routing

eng-esther:
I’m not too certain how many url params there would be, query params would be a cleaner approach.

React router does not require us to explicitly pass in URL params in the <Route> definition because the path wont be matched: https://learnwithparam.com/blog/how-to-handle-query-params-in-react-router/

I decided to not do much there but add a react hook to retrieve query params


kimberlythegeek:
I think this will work for us. Will you be adding the parameters to the compare button in this patch also?

after thinking about how we will use the parameters, I left some comments/suggestions on Beatrice's PR for the endpoint:

> do we intend to handle multiple new revisions in one API call, or make an API call for each new revision?
> 
> If we make one API call for all of them, we may want to shorten these names to base_rev and new_rev or new_rev_1, etc. We don't need to figure this out now, but it's something to consider.

> after thinking on this more, I would think one API call would be preferable. so we can validate if base_rev and base_repo are none, then check for new_revs and name them based on length, new_rev_1, new_rev_2, etc.
> 
> apparently it's possible to send multiple values for a single parameter: https://stackoverflow.com/questions/6243051/how-to-pass-an-array-within-a-query-string
> 
> I think this would help future-proof the code in the event we allow more than four in the future (i.e. for a comparison graph of several different revisions)
> 
> we could also replace the repository with its ID to shorten the URL further. base_repo and new_repos

https://github.com/mozilla/treeherder/pull/7439#discussion_r904077340

kimberlythegeek (approved):

PR #181, opened 2022-06-30
Add parameters to url routing

eng-esther:
Sample url query params currently generated.
Need to confirm with kim how to generate the short code version of revisions

http://localhost:3000/compare-results?revs=0396329b7d2e5e15a0841736625cc89ceadb4cf4,451c34fdf54bf79c705e058531314baaac7e49e7,dc608e935c5901a6404e9c58a5f73084b396f318,9ab74fb00e731757852c19495351b9f15357c5d4&repos=4,4,4,4

after truncating the url looks more like this:
http://localhost:3000/compare-results?revs=65e579f52525,229fa0fcb904,401c2d7e80fb,c4f1589aaaf5&repos=1,1,1,1

### DIFF
--- a/src/components/Search/SearchView.tsx
+++ b/src/components/Search/SearchView.tsx
@@ -9,6 +9,7 @@ import { useNavigate } from 'react-router-dom';
 
 import useFocusInput from '../../hooks/useFocusInput';
 import { Revision, State } from '../../types/state';
+import { truncateHash } from '../../utils/helpers';
 import PerfCompareHeader from '../Shared/PerfCompareHeader';
 import SelectedRevisionsTable from '../Shared/SelectedRevisionsTable';
 import AddRevisionButton from './AddRevisionButton';
@@ -22,8 +23,14 @@ function SearchView(props: SearchViewProps) {
     useFocusInput();
 
   const navigate = useNavigate();
-  const goToCompareResultsPage = () => {
-    navigate('/compare-results', { replace: false });
+
+  const goToCompareResultsPage = (selectedRevisions: Revision[]) => {
+    const revisions = selectedRevisions.map(rev => truncateHash(rev.revision));
+    const repos = selectedRevisions.map(rev => rev.repository_id);
+    navigate({
+      pathname: '/compare-results', 
+      search: `?revs=${revisions.join(',')}&repos=${repos.join(',')}`,
+    });
   };
 
   const { searchResults, selectedRevisions } = props;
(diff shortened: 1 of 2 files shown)


src/components/Search/SearchView.tsx
```diff
@@ -22,8 +22,15 @@ function SearchView(props: SearchViewProps) {
...
   const navigate = useNavigate();
-  const goToCompareResultsPage = () => {
-    navigate('/compare-results', { replace: false });
+
+  const goToCompareResultsPage = (selectedRevisions: Revision[]) => {
+    const revisions = [...selectedRevisions.map(rev => rev.revision)];
+    const repos = [...selectedRevisions.map(rev => rev.repository_id)];
+    navigate({
+      pathname: '/compare-results', 
+      // how to get short form of revision id
```
eng-esther, line 31: @kimberlythegeek  were you able to figure out how to get short form of the revisions

kimberlythegeek: as I understand it, the short hash is just the first 12 characters. you can use `truncateHash` from the helper function file for this

src/components/Search/SearchView.tsx
```diff
@@ -22,8 +23,14 @@ function SearchView(props: SearchViewProps) {
     useFocusInput();
 
   const navigate = useNavigate();
-  const goToCompareResultsPage = () => {
-    navigate('/compare-results', { replace: false });
+
+  const goToCompareResultsPage = (selectedRevisions: Revision[]) => {
+    const revisions = [...selectedRevisions.map(rev => truncateHash(rev.revision))];
+    const repos = [...selectedRevisions.map(rev => rev.repository_id)];
```
julienw, line 29: nit: you don't need to do the spread operation in `[...XXX.map()]` because `map` always returns a new array.

src/components/Search/SearchView.tsx
```diff
@@ -22,8 +23,14 @@ function SearchView(props: SearchViewProps) {
...
+    navigate({
+      pathname: '/compare-results', 
+      search: `?revs=${revisions.toString()}&repos=${repos.toString()}`,
```
julienw, line 32: optional: it may be clearer to the reader if you use the `join` operation: `revisions.join(',')`

julienw (commented):
just a few quick comments, but I'll leave it up to kimberly for the approval :-)

src/components/Search/SearchView.tsx
kimberlythegeek: If this is not needed I'd prefer that it be removed

src/components/Search/SearchView.tsx
kimberlythegeek: I don't have a preference either way. although I didn't know that `array.toString` would automatically comma delimit it--I learned something :)

eng-esther: done

esanuandra (dismissed):

PR #182, opened 2022-07-01
implement edit revison

kimberlythegeek:


kimberlythegeek:
So I know there is likely still work to be done on this patch. In general, I like the way it looks (with some changes to be made), and I think it functions well. However I don't like the approach taken here (the if statements for `view == 'search' and `view == 'compare-results`)

I went back and forth several times while working on this. I had originally planned on doing much of the 'per theme styling' with just CSS, but realized that wasn't possible or feasible in several cases.

I started writing out my reasoning along with the issues I was encountering, but it became very long and I realized it would be better to discuss over a zoom call :) 

Overall I'm not happy with the implementation, I think it is messy and cluttered. I welcome criticism and suggestions on how best to achieve the desired goals here.

src/components/Shared/SelectedRevisionsTableRow.tsx
```diff
@@ -15,50 +18,69 @@ import {
...
+          )}
+          {view == 'compare-results' && (
+            <IconButton
+              id={`edit-revision-button-${index}`}
+              aria-label={`edit-revision-${row.id}`}
+              onClick={() => setIsOpen(true)}
+            >
+              <EditIcon />
+            </IconButton>
+          )}
```
julienw, line 68: same idea here: instead of the `view`, could pass the icon as a parameter, as well as the onClick event handler. And in that case the `popover` would be in the component above.
This works if src/components/Shared/SelectedRevisionsTable.tsx is duplicated, I believe this would be OK because it's a small component.

Generally I believe that as engineers we sometimes have a tendancy to not duplicate, as we were often told that duplication is bad. But it can still be the best strategy. Better engineers than myself say that 2 uses isn't a duplication yet, duplication happens with 3 or more. So I think I'd do this: duplicating SelectedRevisionsTable for the 2 use cases.

src/components/CompareResults/SearchResultsItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+import TableRow from '@mui/material/TableRow';
+
+import type { Revision } from '../../types/state';
+import {
+  truncateHash,
+  getLatestCommitMessage,
+  formatDate,
+} from '../../utils/helpers';
+
+export default function SearchResultsItem(props: SearchResultsItemProps) {
```
julienw, line 14: why are we showing these results differently than in the first page? Can't we reuse the same styling?

PR #187, opened 2022-07-07
fix issue where add revision sometimes did not work

kimberlythegeek:
the previous implementation dispatched two separate actions when adding revisions:
```
dispatch(setSelectedRevisions(newSelected));
dispatch(clearCheckedRevisions());
```

sometimes `clearCheckedRevisions` would occur before `setSelectedRevisions`, and no revision(s) would be added when clicking the add button.

The simplest solution seemed to be to combine `checkedRevisions` and `selectedRevisions` into a single reducer `revisions`, and resetting the state for `checkedRevisions` anytime `setSelectedRevisions` is called.

The first commit includes these changes, while the second commit is just updating all references to checked and selected revisions.


kimberlythegeek:
closing in favor #188 , the solution was simpler than I thought.

PR #191, opened 2022-07-12
Update to React 18

kimberlythegeek:
this update required creating typed hooks `useAppDispatch` and `useAppSelector`, see: https://redux-toolkit.js.org/tutorials/typescript#define-typed-hooks

additional packages were also updated


package.json
```diff
@@ -4,25 +4,24 @@
...
-    "@emotion/react": "^11.8.2",
-    "@emotion/styled": "^11.8.1",
-    "@mui/icons-material": "^5.5.1",
-    "@mui/material": "^5.5.1",
-    "@mui/system": "^5.5.1",
-    "@reduxjs/toolkit": "^1.7.2",
-    "@types/node": "^12.20.48",
-    "@types/react": "^16.14.25",
-    "@types/react-dom": "^16.9.14",
-    "@types/react-redux": "^7.1.22",
```
kimberlythegeek, line 16: react-redux was updated to v8 but this one is still on version 7. v8 has converted the project to typescript, so this may not be needed anymore.

package.json
```diff
@@ -55,39 +54,40 @@
...
+    "eslint-plugin-react": "^7.30.1",
+    "eslint-plugin-react-hooks": "^4.6.0",
+    "gh-pages": "^4.0.0",
     "identity-obj-proxy": "^3.0.0",
-    "jest": "^27.5.1",
+    "jest": "^28.1.2",
     "jest-axe": "^6.0.0",
-    "prettier": "^2.6.0",
-    "react-scripts": "5.0.0",
+    "jest-environment-jsdom": "^28.1.2",
```
kimberlythegeek, line 85: this was need after updating to set the test environment

package.json
kimberlythegeek, line 16: react-redux version 8 updated the project to use typescript, this doesn't appear to be needed anymore

src/__tests__/App.test.tsx
```diff
@@ -1,15 +1,20 @@
...
 
 describe('App', () => {
-  test('Should render search view on default route', () => {
+  test('Should render search view on default route', async () => {
     render(<App />);
 
+    await act(async () => {
+      jest.runOnlyPendingTimers();
+    });
+
```
kimberlythegeek, line 15: some render statements were causing 'not wrapped in act' errors due to state updates occurring after the test finished. adding runOnlyPending timers within an act call resolved these errors.

```
console.error node_modules/react-dom/cjs/react-dom.development.js:530
  Warning: An update to <component> inside a test was not wrapped in act(...).

  When testing, code that causes React state updates should be wrapped into act(...):

  act(() => {
    /* fire events that update state */
  });
  /* assert on the output */

  This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
      in <component>
      ```

File: src/__tests__/Search/SearchView.test.tsx [hunk 4]
```diff
@@ -7,9 +8,22 @@ import { renderWithRouter, store } from '../utils/setupTests';
...
+  it('render correctly when there are no results', async () => {
+    const { testData } = getTestData();
+    global.fetch = jest.fn(() =>
+      Promise.resolve({
+        json: () => ({
+          results: testData,
+        }),
+      }),
+    ) as jest.Mock;
+
```
kimberlythegeek, line 20: without mocking the call to `fetchRecentRevisions` that occurs on mount of SearchView, the snapshot was displaying an error because fetch returned no results

src/__tests__/Search/__snapshots__/SearchResultsList.test.tsx.snap
```diff
@@ -434,9 +421,6 @@ exports[`SearchResultsList should match snapshot 1`] = `
                     </p>
                   </div>
                 </li>
-                <span
-                  class="MuiTouchRipple-root css-8je8zh-MuiTouchRipple-root"
-                />
```
kimberlythegeek, line 439: I'm not 100% sure what caused these changes, but I believe it's due to an update of material ui

src/__tests__/Search/__snapshots__/SearchView.test.tsx.snap
```diff
@@ -554,19 +551,10 @@ exports[`Search View should not hide search results when clicking search results
...
-                >
-                  <span
-                    class="css-y4cjyz-MuiTouchRipple-ripple MuiTouchRipple-ripple MuiTouchRipple-rippleVisible"
-                    style="width: 2.8284271247461903px; height: 2.8284271247461903px; top: -1.4142135623730951px; left: -1.4142135623730951px;"
-                  >
-                    <span
-                      class="MuiTouchRipple-child MuiTouchRipple-childLeaving"
-                    />
-                  </span>
-                </span>
```
kimberlythegeek, line 566: same here

package.json
```diff
@@ -111,8 +111,7 @@
             }
           }
         }
-      ],
-      "^.+\\.svg$": "<rootDir>/src/__tests__/utils/fileTransformer.js"
```
kimberlythegeek: this does not appear to be needed anymore

src/components/Search/SearchDropdown.tsx
```diff
@@ -21,7 +21,7 @@ function SearchDropdown(props: SearchDropdownProps) {
             id={repoMap[key]}
             value={repoMap[key]}
             key={repoMap[key]}
-            onClick={(e) => handleChangeDropdown(e)}
+            onClick={(e) => void handleChangeDropdown(e)}
```
kimberlythegeek, line 24: `Promise-returning function provided to attribute where a void return was expected  @typescript-eslint/no-misused-promises`

https://typescript-eslint.io/rules/no-misused-promises/

src/components/Search/SearchViewInit.tsx
```diff
@@ -1,16 +1,17 @@
...
+import type { Repository, State } from '../../types/state';
 
 // component to fetch recent revisions when search view is loaded
 function SearchViewInit(props: SearchViewInitProps) {
-  const dispatch = useDispatch();
+  const dispatch = useAppDispatch();
   const { repository } = props;
   useEffect(() => {
-    dispatch(fetchRecentRevisions(repository));
+    void dispatch(fetchRecentRevisions(repository));
```
kimberlythegeek, line 14: `Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator`

https://typescript-eslint.io/rules/no-floating-promises/

src/components/Search/SearchViewInit.tsx
```diff
@@ -1,16 +1,17 @@
...
 
+import { useAppDispatch } from '../../hooks/app';
 import { fetchRecentRevisions } from '../../thunks/searchThunk';
-import { Repository, State } from '../../types/state';
+import type { Repository, State } from '../../types/state';
 
 // component to fetch recent revisions when search view is loaded
 function SearchViewInit(props: SearchViewInitProps) {
-  const dispatch = useDispatch();
+  const dispatch = useAppDispatch();
```
kimberlythegeek, line 11: https://redux-toolkit.js.org/tutorials/typescript#define-typed-hooks

without using a typed hook, I was getting these errors:
```
TS2345: Argument of type 'AsyncThunkAction<Revision[], "mozilla-central" | "mozilla-beta" | "mozilla-release" | "try" | "fenix" | "autoland", { rejectValue: string; }>' is not assignable to parameter of type 'AnyAction'.
    13 |   const { repository } = props;
    14 |   useEffect(() => {
  > 15 |     void dispatch(fetchRecentRevisions(repository));
       |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    16 |   }, []);
    17 |   return null;
    18 | }
```

src/components/CompareResults/CompareResultsTable.tsx
```diff
@@ -25,8 +24,8 @@ const tableHeaders: CompareResultsTableHeaders[] = [
 
 function CompareResultsTable(props: CompareResultsProps) {
   const { mode } = props;
-  const compareResults: CompareResultsItem[] = useSelector(
-    (state: RootState) => state.compareResults,
```
kimberlythegeek, line 29: specifying the type is not needed when using typed hooks: https://redux-toolkit.js.org/tutorials/typescript#define-typed-hooks

src/__tests__/utils/fileTransformer.js
```diff
@@ -2,6 +2,8 @@ const path = require('path');
 
 module.exports = {
   process(src, filename) {
-    return `module.exports = ${JSON.stringify(path.basename(filename))};`;
+    return {
+      code: `module.exports = ${JSON.stringify(path.basename(filename))};`,
+    };
```
kimberlythegeek, line 7:
```
FAIL  src/__tests__/App.test.tsx
  ● Test suite failed to run

    ● Invalid return value:
      `process()` or/and `processAsync()` method of code transformer found at 
      "/home/circleci/project/src/__tests__/utils/fileTransformer.js" 
      should return an object or a Promise resolving to an object. The object 
      must have `code` property with a string of processed code.
      This error may be caused by a breaking change in Jest 28:
      https://jestjs.io/docs/upgrading-to-jest28#transformer
      Code Transformation Documentation:
      https://jestjs.io/docs/code-transformation
```

src/components/Search/SearchViewInit.tsx
eng-esther: it's new for me to see void on a function call.
I guess this is how react handles return type now ?

eng-esther (approved):

src/components/Search/SearchViewInit.tsx
julienw: This is a standard javascript operator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void

Basically this "voids" the value on the right. It's not common to use it directly (at least _I_ don't use it often if any :-) ), however you can see this in minified scripts, because minifiers usually replace `undefined` by `void 0` because it's shorter.

package.json
```diff
@@ -55,39 +54,40 @@
...
+    "gh-pages": "^4.0.0",
     "identity-obj-proxy": "^3.0.0",
-    "jest": "^27.5.1",
+    "jest": "^28.1.2",
     "jest-axe": "^6.0.0",
-    "prettier": "^2.6.0",
-    "react-scripts": "5.0.0",
+    "jest-environment-jsdom": "^28.1.2",
+    "prettier": "^2.7.1",
+    "react-scripts": "5.0.1",
```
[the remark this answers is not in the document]
kimberlythegeek: I'm actually working on a patch to eject the project, as far as I know this is the only way to remove react-scripts.

PR #193, opened 2022-07-14
FXP-2289 Add links to graphs

esanuandra:


beatrice-acasandrei (approved):

kimberlythegeek (changes requested):
Looks good--however I think that we should have some hover action where this icon changes color so that it behaves more like a typical link.

Perhaps use `color: inherit` on hover? or we could do the reverse, using the theme defaults for on hover.

src/components/CompareResults/CompareResultsTableRow.tsx
```diff
@@ -17,6 +19,14 @@ function CompareResultsTableRow(props: ResultsTableRowProps) {
           )}`}
         ></TableCell>
       </Tooltip>
+      <TableCell>
+        <Link
+          href={'#'}
+          className={`background-icon ${mode}-mode graph-icon-color`}
+        >
```
kimberlythegeek, line 26: can you also add alt text or an aria label for accessibility?

kimberlythegeek (approved):

PR #198, opened 2022-07-22
eject from react-scripts and update to react 18

kimberlythegeek:
my assumption is that what I've done is not an optimal solution and I don't consider it 'ready to merge', as there are many changes made here and I'm not completely confident I made the best decisions in implementation.

@julienw as far as the `async` `await` for `act()` calls, if I don't do that, the test will pass, but it throws a 'not wrapped in act' warning'

namely when calling `.jest.runOnlyPendingTimers()`. I tried many workarounds for this, but the only thing I found to resolve the issue was wrapping the call in `act` and using `await` `async`


julienw:
> @julienw as far as the async await for act() calls, if I don't do that, the test will pass, but it throws a 'not wrapped in act' warning'

From my limited testing, it seems to be the case for some tests (eg `App.test.tsx` seems to need it) but not in others (the test `Search/fetchRevisionByID.test.tsx` seems to work fine without the await/async act).

My rough guess is that in the ones that need it, the pending timers trigger promises, and these promises resolution callback end up out of the `act` call.

It's not clear to me what the best way forward is, let's say that keeping await/async in these cases makes sense (with comments), but that it would be good to not use it when we don't need it. The rationale here is that I believe it's good to avoid "magic incantations", that in the future we don't know anymore why they're needed and we end up adding them everywhere, and we're afraid of removing them.

package.json
```diff
@@ -55,67 +47,55 @@
...
-    "moduleNameMapper": {
-      "\\.(css)$": "identity-obj-proxy"
-    }
+    "ts-jest": "^28.0.5",
+    "ts-loader": "^9.3.1",
+    "typescript": "^4.7.4",
+    "webpack": "^5.74.0",
+    "webpack-cli": "^4.10.0",
+    "webpack-dev-server": "^4.9.3",
+    "webpack-merge": "^5.8.0"
```
kimberlythegeek, line 99: webpack-dev-server is for running the app locally, webpack-merge is used to merge configuration files

public/index.html
```diff
@@ -2,28 +2,17 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="icon" href="%PUBLIC_URL%/favicon.svg" type="image/svg+xml" />
+    <link rel="icon" href="favicon.svg" type="image/svg+xml" />
```
kimberlythegeek, line 5: `%PUBLIC_URL%` is no longer needed when building with webpack

public/index.html
```diff
@@ -2,28 +2,17 @@
...
       name="description"
       content="Web site created using create-react-app"
     />
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/favicon.svg" />
+    <link rel="apple-touch-icon" href="favicon.svg" />
     <!--
       manifest.json provides metadata used when your web app is installed on a
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
     -->
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
```
kimberlythegeek, line 17: generated file left over from `create-react-app`

src/__tests__/App.test.tsx
```diff
@@ -1,12 +1,14 @@
...
 
 import App from '../components/App';
 import { render } from './utils/setupTests';
 import { screen } from './utils/test-utils';
 
 describe('App', () => {
-  test('Should render search view on default route', () => {
+  test('Should render search view on default route', async () => {
     render(<App />);
+    await act(async () => void jest.runOnlyPendingTimers());
```
kimberlythegeek, line 11: this is sometime required to avoid `an update to ___ was not wrapped in act()` warnings

src/__tests__/App.test.tsx
```diff
@@ -24,11 +26,9 @@ describe('App', () => {
     });
 
     await user.click(darkModeButton);
-    jest.runOnlyPendingTimers();
```
kimberlythegeek, line 27: no longer needed

src/__tests__/App.test.tsx
```diff
@@ -24,11 +26,9 @@ describe('App', () => {
...
     expect(screen.queryByTestId('Brightness7Icon')).toBeInTheDocument();
 
     await user.click(darkModeButton);
-    jest.runOnlyPendingTimers();
```
kimberlythegeek, line 31: same

src/__tests__/Search/__snapshots__/SearchResultsList.test.tsx.snap
```diff
@@ -380,12 +370,9 @@ exports[`SearchResultsList should match snapshot 1`] = `
                     </p>
                   </div>
                 </li>
-                <span
-                  class="MuiTouchRipple-root css-8je8zh-MuiTouchRipple-root"
-                />
```
kimberlythegeek, line 385: this was altered due to a change in the await/act/run timers implementation. I believe this is an animation that is no longer visible

src/__tests__/utils/fileTransformer.js
```diff
@@ -2,6 +2,8 @@ const path = require('path');
 
 module.exports = {
   process(src, filename) {
-    return `module.exports = ${JSON.stringify(path.basename(filename))};`;
+    return {
+      code: `module.exports = ${JSON.stringify(path.basename(filename))};`,
+    };
```
kimberlythegeek, line 7:
```
    ● Invalid return value:
      `process()` or/and `processAsync()` method of code transformer found at 
      "/home/kimberly/Workspace/perfcompare/src/__tests__/utils/fileTransformer.js" 
      should return an object or a Promise resolving to an object. The object 
      must have `code` property with a string of processed code.
      This error may be caused by a breaking change in Jest 28:
      https://jestjs.io/docs/upgrading-to-jest28#transformer
      Code Transformation Documentation:
      https://jestjs.io/docs/code-transformation
```

src/components/Search/SearchDropdown.tsx
```diff
@@ -21,7 +21,7 @@ function SearchDropdown(props: SearchDropdownProps) {
             id={repoMap[key]}
             value={repoMap[key]}
             key={repoMap[key]}
-            onClick={(e) => handleChangeDropdown(e)}
+            onClick={(e) => void handleChangeDropdown(e)}
```
kimberlythegeek, line 24: `Promise-returning function provided to attribute where a void return was expected`

src/components/Search/SearchViewInit.tsx
```diff
@@ -1,16 +1,17 @@
...
+import type { Repository, State } from '../../types/state';
 
 // component to fetch recent revisions when search view is loaded
 function SearchViewInit(props: SearchViewInitProps) {
-  const dispatch = useDispatch();
+  const dispatch = useAppDispatch();
   const { repository } = props;
   useEffect(() => {
-    dispatch(fetchRecentRevisions(repository));
+    void dispatch(fetchRecentRevisions(repository));
```
kimberlythegeek, line 14: `Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.`

this promise is awaited in the thunk action

src/hooks/app.ts
```diff
@@ -0,0 +1,8 @@
+import { useDispatch, useSelector } from 'react-redux';
+import type { TypedUseSelectorHook } from 'react-redux';
+
+import type { RootState, AppDispatch } from '../common/store';
+
+// Use throughout your app instead of plain `useDispatch` and `useSelector`
+export const useAppDispatch: () => AppDispatch = useDispatch;
+export const useAppSelector: TypedUseSelectorHook<RootState> = useSelector;
```
kimberlythegeek, line 8: https://redux-toolkit.js.org/tutorials/typescript#define-typed-hooks

kimberlythegeek:
I realize this is a very large patch. It may be possible to break this down further, but I was unsure how to do so without breaking the build or the tests. I tried to document each notable change I made and the reason for the change.

package.json
```diff
@@ -55,67 +47,55 @@
     ]
   },
   "devDependencies": {
-    "@babel/core": "^7.17.8",
-    "@babel/preset-env": "^7.16.11",
-    "@babel/preset-react": "^7.16.7",
-    "@babel/preset-typescript": "^7.16.7",
-    "@swc/core": "^1.2.196",
```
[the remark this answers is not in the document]
julienw: My understanding is that they were recommending to install @swc/core (see their doc: https://swc.rs/docs/usage/jest)
But if that wotks without the entry in package.json I'm fine with that too, so that's up to you.

package.json
```diff
@@ -55,67 +47,55 @@
...
-    "@typescript-eslint/eslint-plugin": "^5.20.0",
-    "@typescript-eslint/parser": "^5.20.0",
-    "babel-jest": "^27.5.1",
-    "core-js": "^3.21.1",
+    "@types/node": "^18.6.1",
+    "@typescript-eslint/eslint-plugin": "^5.31.0",
+    "@typescript-eslint/parser": "^5.31.0",
+    "babel-loader": "^8.2.5",
+    "babel-preset-jest": "^28.1.3",
+    "case-sensitive-paths-webpack-plugin": "^2.4.0",
```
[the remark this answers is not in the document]
julienw: sounds a good idea

jest.config.ts
```diff
@@ -0,0 +1,52 @@
...
+          },
+        },
+      },
+    ],
+    '^.+\\.svg$': '<rootDir>/src/__tests__/utils/fileTransformer.js',
+  },
+  transformIgnorePatterns: [
+    '[/\\\\]node_modules[/\\\\].+\\.(js|jsx|mjs|cjs|ts|tsx)$',
+    '^.+\\.module\\.(css|sass|scss)$',
+  ],
```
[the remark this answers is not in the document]
julienw: Keeping this sounds good to me, we can look at removing this in a future PR if we see it's not needed.

.babelrc
```diff
@@ -0,0 +1,17 @@
+{
+  "targets": ["last 2 chrome versions", "last 2 firefox versions"],
+  "plugins": [
+    [
+      "@babel/plugin-transform-runtime",
+      {
+        "regenerator": true
+      }
+    ]
+  ],
```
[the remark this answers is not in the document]
julienw: a few notes about that (all optional for this patch):
1. my understanding is that `@babel/preset-env` will add this plugin automatically
2. `true` seems to be the default for the `regenerator` option
3. That said I doubt this will even be used if you target latest versions of chrome and firefox. :)

.babelrc
```diff
@@ -0,0 +1,17 @@
...
+  "presets": [
+    "@babel/preset-env",
+    ["@babel/preset-react", { "runtime": "automatic" }],
```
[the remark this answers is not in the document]
julienw: sounds great to me!

.babelrc
```diff
@@ -0,0 +1,17 @@
...
+  "plugins": [
+    [
+      "@babel/plugin-transform-runtime",
+      {
+        "regenerator": true
+      }
+    ]
+  ],
+  "presets": [
+    "@babel/preset-env",
```
julienw, line 12: This is how we configure preset-env in the profiler:
https://github.com/firefox-devtools/profiler/blob/a299ff00ea661579e96843d4c3d8cc7772133bc5/.babelrc#L4-L13

optional for this patch, it's better to look at this in another PR -- no high priority either.

.babelrc
```diff
@@ -0,0 +1,17 @@
...
+    ["@babel/preset-react", { "runtime": "automatic" }],
+    "@babel/preset-typescript",
+    "babel-preset-jest"
```
julienw, line 15: This seems to be automatically added when using babel-jest (which is also automatic when running jest and babel-jest is installed), so could possibly remove this line. Also I'm a bit worried that this transform is always present even when not running tests (that said I have no idea what this does).

update: I see that `babel-jest` is not installed in this project, so if everything is running fine with this config, I suggest to handle the proposed change in a future PR.

public/index.html
julienw: You can remove the comment above too.

julienw (approved):
Great work Kimberly!
I don't see big problems, only a few adjustments that could be made in small PRs.
I also gave the app a try locally, this looked good to me, so let's use the new React API!

So I'd say, let's land this and adjust later!

.babelrc
kimberlythegeek: #202

kimberlythegeek:
I added a couple of changes--I noticed the tests were taking a longer time to run again (~20 seconds). The culprit seemed to be the material-ui imports in `SearchView.tsx` that were pulling in unused modules.
See: https://mui.com/material-ui/guides/minimizing-bundle-size/#option-1

I tried a couple of other things, but the only significant difference I noticed was when I updated these imports.

also updated the readme instructions to indicate that `npm run start` is the command to run the development server.

This was changed originally to fix the heroku deployment. The issue was that heroku uses `npm run start` to build the app, and `start` used `react-scripts`. This package was moved to `devDependencies` due to a security vulnerability, and heroku will only install production dependencies when building.

see: https://github.com/mozilla/perfcompare/pull/81#issue-1205807378

PR #201, opened 2022-08-04
update mock data using results from actual tests

kimberlythegeek:


src/types/state.ts
```diff
@@ -46,18 +47,26 @@ export type CompareResultsItem = {
...
+  delta_percentage: number;
+  magnitude: number;
   is_improvement: boolean;
-  is_regression: boolean;
-  is_meaningful: boolean;
+  is_empty: boolean;
   test: string;
+  option_name: string;
+  extra_options: string;
 };
```
kimberlythegeek, line 71: updated type definition to match the response from the API

src/types/types.ts
```diff
@@ -28,7 +28,8 @@ export type MeasurementUnit =
   | 'mWh'
   | 'KB'
   | 'fps'
-  | 'bytes';
+  | 'bytes'
+  | null;
```
kimberlythegeek, line 32: measurement unit for some tests is `None`

src/utils/helpers.ts
```diff
@@ -45,10 +45,15 @@ const setPlatformClassName = (platformName: string) => {
   else return '';
 };
 
+const setConfidenceClassName = (confidenceText: string | null) => {
+  return confidenceText == null ? '' : confidenceText;
+};
```
kimberlythegeek, line 50: helper method to return empty string if confidence_text is null

src/components/CompareResults/CompareResultsView.tsx
```diff
@@ -11,10 +11,10 @@ import CompareResultsTable from './CompareResultsTable';
 function CompareResultsView(props: CompareResultsViewProps) {
   const { revisions, mode } = props;
   return (
-    <Container maxWidth="lg">
+    <Container maxWidth="xl">
```
kimberlythegeek, line 14: made the container wider so that the results table doesn't overflow and cause side-scrolling

eng-esther (approved):

beatrice-acasandrei (approved):

PR #203, opened 2022-08-09
replace usage of useSelector with typed hook useAppSelector

kimberlythegeek:
in #198 I added typed hooks for useDispatch and useSelector, so I have replaced usage of useSelector with the typed hook useAppSelector

Also removed `State` type and replaced with `RootState`

### DIFF
--- a/src/hooks/useSelectRevision.ts
+++ b/src/hooks/useSelectRevision.ts
@@ -1,7 +1,6 @@
 import { useSnackbar, VariantType } from 'notistack';
 
 import { maxRevisionsError } from '../common/constants';
-import type { RootState } from '../common/store';
 import { clearCheckedRevisions } from '../reducers/CheckedRevisions';
 import { setSelectedRevisions } from '../reducers/SelectedRevisions';
 import { truncateHash } from '../utils/helpers';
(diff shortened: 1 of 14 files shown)


src/hooks/useSelectRevision.ts
```diff
@@ -11,16 +10,14 @@ const useSelectRevision = () => {
...
-  );
+  const searchResults = useAppSelector((state) => state.search.searchResults);
 
   const checkedRevisions = useAppSelector(
-    (state: RootState) => state.checkedRevisions.revisions,
+    (state) => state.checkedRevisions.revisions,
   );
 
   const selectedRevisions = useAppSelector(
-    (state: RootState) => state.selectedRevisions.revisions,
```
eng-esther, line 23: [question] should the type be added, will the linter complain?

eng-esther (approved):

src/hooks/useSelectRevision.ts
kimberlythegeek: the typed hooks remove the need for specifying a type here: https://redux-toolkit.js.org/tutorials/typescript#define-typed-hooks

PR #205, opened 2022-08-11
make icons bigger

eng-esther:
[Make Icons bigger](https://mozilla-hub.atlassian.net/browse/FXP-2382)
Increased icon size to 17.5%


eng-esther:
branched off master instead of staging

PR #209, opened 2022-08-11
store revision instead of index in checkedRevisions

kimberlythegeek:
I ran in to an implementation problem while working on editing a revision from results view. since we are replacing a revision instaed of just adding a new one, we need a reference to the revision we are replacing.

previously the checkedRevisions state held a revisions index from searchResults. I've updated the state to hold the actual revision, which should resolve this problem.

### DIFF
--- a/src/__tests__/Search/SearchResultsList.test.tsx
+++ b/src/__tests__/Search/SearchResultsList.test.tsx
@@ -79,11 +79,9 @@ describe('SearchResultsList', () => {
     );
 
     await user.click(fleshWound);
-    expect(store.getState().checkedRevisions.revisions[0]).toBe(1);
+    expect(store.getState().checkedRevisions.revisions[0]).toBe(testData[1]);
     await user.click(fleshWound);
-    expect(
-      screen.getByTestId('checkbox-1').classList.contains('Mui-checked'),
-    ).toBe(false);
+    expect(fleshWound.classList.contains('Mui-checked')).toBe(false);
   });
 
   it('should not allow selecting more than four revisions', async () => {
--- a/src/__tests__/Search/__snapshots__/SearchResultsList.test.tsx.snap
+++ b/src/__tests__/Search/__snapshots__/SearchResultsList.test.tsx.snap
@@ -169,7 +169,6 @@ exports[`SearchResultsList should match snapshot 1`] = `
             >
               <div
                 class="MuiListItemButton-root MuiListItemButton-gutters MuiButtonBase-root css-16ac5r2-MuiButtonBase-root-MuiListItemButton-root"
-                id="0"
                 role="button"
                 tabindex="0"
               >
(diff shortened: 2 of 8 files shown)


src/__tests__/Search/__snapshots__/SearchResultsList.test.tsx.snap
```diff
@@ -169,7 +169,6 @@ exports[`SearchResultsList should match snapshot 1`] = `
             >
               <div
                 class="MuiListItemButton-root MuiListItemButton-gutters MuiButtonBase-root css-16ac5r2-MuiButtonBase-root-MuiListItemButton-root"
-                id="0"
```
kimberlythegeek, line 172: IDs were removed because they are no longer needed. previously the `useCheckRevision` hook was using the ID to add it to the checkedRevisions state, but now it stores the actual revision instead.

src/__tests__/Search/SearchResultsList.test.tsx
```diff
@@ -79,11 +79,9 @@ describe('SearchResultsList', () => {
     );
 
     await user.click(fleshWound);
-    expect(store.getState().checkedRevisions.revisions[0]).toBe(1);
```
kimberlythegeek, line 82: checkedRevisions now holds Revision items insead of indices

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -12,22 +12,16 @@ import { truncateHash, getLatestCommitMessage } from '../../utils/helpers';
 function SearchResultsListItem(props: SearchResultsListItemProps) {
   const { index, item } = props;
   const isChecked: boolean = useAppSelector((state) =>
-    state.checkedRevisions.revisions.includes(index),
+    state.checkedRevisions.revisions.includes(item),
   );
   const { handleToggle } = useCheckRevision();
 
-  const indexString = index.toString();
```
kimberlythegeek, line 19: I think this was added for typechecking lint errors on the string interpolation, but it doesn't appear to be needed anymore.

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -12,22 +12,16 @@ import { truncateHash, getLatestCommitMessage } from '../../utils/helpers';
...
-
   const revisionHash = truncateHash(item.revision);
   const commitMessage = getLatestCommitMessage(item);
 
   return (
     <>
-      <ListItemButton
-        key={item.id}
-        id={indexString}
```
kimberlythegeek, line 28: `id` was used by the `handleToggle` hook to check/uncheck a result. Now we are passing the revision itself to store in checkedRevisions.

src/hooks/useSelectRevision.ts
```diff
@@ -10,8 +10,6 @@ const useSelectRevision = () => {
   const dispatch = useAppDispatch();
   const { enqueueSnackbar } = useSnackbar();
 
-  const searchResults = useAppSelector((state) => state.search.searchResults);
```
kimberlythegeek, line 13: searchResults was used to determine which revision to add to selectedRevisions, which I realize now was a bit of a convoluted way to accomplish this.

no longer needed since we are storing Revision objects directly in checkedRevisions state.

eng-esther (approved):

PR #214, opened 2022-08-12
edit revision ui

kimberlythegeek:
I modified the edit revision UI to be cleaner and more reusable. Instead of using new component to display a minified UI, I used material UI's `dense` and `size` props, and other wise search results look the same on Search View and Compare View.

In the future I'd like to disable all other results when the max revisions is reached, and remove the need for the alert.


src/components/Shared/SelectedRevisionsTableRow.tsx
```diff
@@ -18,66 +14,43 @@ import {
...
-              id={`edit-revision-button-${index}`}
-              aria-label={`edit-revision-${row.id}`}
-              onClick={() => setIsOpen(true)}
-            >
-              <EditIcon />
-            </IconButton>
-          )}
-        </TableCell>
-      </TableRow>
-      <Popover
```
kimberlythegeek, line 68: popover moved to the new EditRevision component

src/hooks/useSelectRevision.ts
```diff
@@ -30,7 +30,7 @@ const useSelectRevision = () => {
 
       // Do not allow adding more than four revisions
       if (selectedRevisions.length == 4) {
-        enqueueSnackbar(maxRevisionsError, { variant });
+        enqueueSnackbar(maxRevisionsError as string, { variant });
```
kimberlythegeek, line 33: added to fix a lint errror

src/theme/components.js
```diff
@@ -55,29 +55,9 @@ const components = {
       },
     },
   },
-  MuiTable: {
```
kimberlythegeek, line 58: all the styling removed was in place to create a compact UI, now handled by `dense` and `size` props

eng-esther (approved):

PR #216, opened 2022-08-13
Implement edit revision from Compare Results View

kimberlythegeek:


kimberlythegeek:
this patch is based on both #209 and #214

src/components/Shared/RevisionSearch.tsx
```diff
@@ -17,23 +18,23 @@ import SearchResultsList from '../Search/SearchResultsList';
...
+  const { replaceSelectedRevision } = useSelectRevision();
 
   const handleFocus = (e: MouseEvent) => {
     if (
       (e.target as HTMLElement).matches(
-        `#search-revision-input, 
-          #search-results-list, 
-          #search-results-list *`,
+        `#revision-search-container, 
+        #revision-search-container *`,
```
kimberlythegeek, line 30: fixing focus issues

src/components/Shared/RevisionSearch.tsx
```diff
@@ -17,23 +18,23 @@ import SearchResultsList from '../Search/SearchResultsList';
...
+      ) &&
+      // do not open search results when cancel button is clicked
+      !(e.target as HTMLElement).matches(
+        '#cancel-edit-revision-button, #cancel-edit-revision-button *',
```
kimberlythegeek, line 34: this is to prevent the search results appearing before the popover is closed, when the cancel button is clicked

src/components/Shared/RevisionSearch.tsx
```diff
@@ -63,23 +64,41 @@ function RevisionSearch(props: RevisionSearchProps) {
...
       </Grid>
       <Grid item xs={9}>
-        <SearchInput setFocused={setFocused} view={view} />
+        <SearchInput view={view} setFocused={setFocused} />
       </Grid>
 
       <Grid item xs={1}>
         {view == 'search' && <AddRevisionButton setFocused={setFocused} />}
-        {view == 'compare-results' && (
+        {view == 'compare-results' && setPopoverIsOpen && prevRevision && (
```
kimberlythegeek, line 82: prevent type errors on value possibly being undefined. these props are optional as they are only used when render on Compare Results View

src/components/CompareResults/EditRevisionButton.tsx
```diff
@@ -36,7 +36,11 @@ function EditRevisionButton(props: EditRevisionButtonProps) {
           horizontal: 'left',
         }}
       >
-        <RevisionSearch view="compare-results" />
+        <RevisionSearch
+          view="compare-results"
+          prevRevision={item}
```
kimberlythegeek, line 41: optional prop used when replacing selected revision

src/components/Shared/RevisionSearch.tsx
```diff
@@ -17,23 +18,23 @@ import SearchResultsList from '../Search/SearchResultsList';
...
+      !(e.target as HTMLElement).matches(
+        '#cancel-edit-revision-button, #cancel-edit-revision-button *',
       )
     ) {
       setFocused(true);
-    } else if (
-      (e.target as HTMLElement).matches(
-        '#add-revision-button,#add-revision-button *',
-      )
-    ) {
```
kimberlythegeek, line 36: no longer need with new `.matches` call

src/hooks/useSelectRevision.ts
```diff
@@ -18,28 +19,38 @@ const useSelectRevision = () => {
     (state) => state.selectedRevisions.revisions,
   );
 
+  const warningVariant: VariantType = 'warning';
+
+  const enqueueAlreadySelectedAlert = (
```
kimberlythegeek, line 24: helper method since this error is used by both hooks

src/components/Shared/RevisionSearch.tsx
```diff
@@ -17,23 +18,23 @@ import SearchResultsList from '../Search/SearchResultsList';
...
     if (
       (e.target as HTMLElement).matches(
-        `#search-revision-input, 
-          #search-results-list, 
-          #search-results-list *`,
+        `#revision-search-container, 
+        #revision-search-container *`,
+      ) &&
+      // do not open search results when cancel button is clicked
+      !(e.target as HTMLElement).matches(
```
eng-esther, line 33: [question] are we 100% sure all users will use firefox for this tool? Does Safari/Chrome have the same API?

src/components/Shared/RevisionSearch.tsx
eng-esther, line 30: [question] what's the diff if (we remove the space)
```suggestion
        #revision-search-container*`,
```

eng-esther (approved):

src/components/Shared/RevisionSearch.tsx
kimberlythegeek: This is to match any element that is a child of the container element (however far nested). I'm unsure, but I think without the space it either would not work or it would try to match any element with an id that begins with `revision-search-container`

beatrice-acasandrei:
Right now after we click the edit button in the Compare View the selected revision is not highlighted in any way in the suggested list. Would it be worthwhile to have the checkbox selected if the revision is found in the list?
suggestion

Noticed that this is the behavior after replacing a revision, but not by default and in my opinion it would be nice to have it by default, either in this PR or another.

kimberlythegeek:
@beatrice-acasandrei 
> Right now after we click the edit button in the Compare View the selected revision is not highlighted in any way in the suggested list. Would it be worthwhile to have the checkbox selected if the revision is found in the list? 
> 
> Noticed that this is the behavior after replacing a revision, but not by default and in my opinion it would be nice to have it by default, either in this PR or another.

sounds good, I will file an issue in jira for this

PR #218, opened 2022-08-16
fix errors in accessibility test

kimberlythegeek:
there were `act` warnings appearing in the accessibility tests, and one test was missing the `view` prop

also added a test for compare view in dark mode


beatrice-acasandrei (approved):

eng-esther (approved):

kimberlythegeek:
I haven't yet merged this because I realized that an error was being thrown after I added a new test. (Currently CI does not fail the run if accessibility tests fail) 

I tried to resolve it but have not yet found a solution. For now I have commented this test out and filed a github issue #222

oburasamuel:
Hi Kimberly, how many dependancies are we installing for this test?

PR #229, opened 2022-08-30
Table filters results view

esanuandra:


src/common/constants.ts
```diff
@@ -29,6 +29,19 @@ export const frameworkMap: Record<Framework['id'], Framework['name']> = {
...
+export const frameworks: Framework[] = [{ id: 1, name: 'talos' },
+{ id: 2, name: 'build_metrics' },
+{ id: 4, name: 'awsy' },
+{ id: 6, name: 'platform_microbench' },
+{ id: 10, name: 'raptor' },
+{ id: 11, name: 'js-bench' },
+{ id: 12, name: 'devtools' },
+{ id: 13, name: 'browsertime' },
+{ id: 15, name: 'mozperftest' },
+{ id: 16, name: 'fxrecord' },
```
eng-esther, line 41: [question] id is not hard coded right ?

kimberlythegeek: Each framework has an ID that comes from the backend, yes. I assume, though I'm not sure, that this also correlates to an ID in the database?

This is the API endpoint: https://treeherder.mozilla.org/api/performance/framework/

src/__tests__/CompareResults/CompareResultsTable.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+    store.dispatch(addFilter({ name: 'platform', value: 'macosx1015-64-shippable-qr' }));
+
+    await act( async () => rerender(<CompareResultsTable mode="light" />));
+
+    const filteredRows = screen.getAllByTestId('table-row');
+
+    expect(filteredRows.length).toBeNull();
+  });
+
+});
```
kimberlythegeek, line 53: I've been trying to figure out why this test is failing and so far I have no ideas. I tried using test data for `compareResults` test, but it does not seem to be working. I created a simple reducer `setCompareResults`, and logging statements to track what is going on.

The reducer is being called and a `console.log` statement in the reducer shows it is being set with the test data. However when I call `store.getState().compareResults` from within the test, it always returns the original state.

If we can't resolve this within the next couple of days, I'd recommend filing a ticket to address this and prioritize it after the work week.

kimberlythegeek (commented):
ran again and the filters did work. Also I like the 'filters have been applied' message. Should we create a ticket to auto-dismiss this after a timeout? also a ticket for a 'clear all filters' button.

overall I think the UI looks good. Any ideas what is causing the long delay? I thought maybe it was due to the large file of mock results, but after removing all but two, there is still a long delay when filtering.

I don't want to be overly critical, but I can't but think there may be a more simple way to accomplish the filtering, such as using the `filter` function rather than creating a new reducer for filtered results, but I could be wrong about that. I realize this would require chaining `filter` functions for each of the "filter-able" properties. 

Can you request a review from @julienw to get his opinion also?

I know you've done a lot of work here, but if there is a more straightforward way to accomplish this, we can still merge this work to the `mocking` branch for the work week and address those issues later.

src/__tests__/CompareResults/CompareResultsTable.test.tsx
esanuandra: I made it fail on purpose to see what data is there. I found a way on how to filter the data in a test and updated the pr.

src/components/CompareResults/FilterOptionsList.tsx
```diff
@@ -0,0 +1,96 @@
...
+                <ListItemIcon>
+                  <Checkbox
+                    data-testid={`${value}-checkbox`}
+                    edge="start"
+                    disableRipple
+                    inputProps={{ 'aria-labelledby': labelId, value: value }}
+                    checked={activeFilters[
+                      column as keyof typeof activeFilters
+                    ].includes(value)}
+                    onChange={handleOnChangeOption}
```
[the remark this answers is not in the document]
esanuandra: I moved it on the local branch and it makes the filtering slower. Maybe we could work on this on another pr.

kimberlythegeek (approved):
I'm approving this for the mocking branch--can you file a ticket to add test coverage for this code?

src/components/CompareResults/CompareResultsTable.tsx
```diff
@@ -6,52 +10,237 @@ import TableContainer from '@mui/material/TableContainer';
...
+            {...params}
+            variant="standard"
+            label="Framework"
+            placeholder="Select framework"
+          />
+        )}
+      />
+      {isFiltered && <Alert severity="info">Filters have been applied</Alert>}
+      <TableContainer component={Paper}>
+        <Table sx={{ minWidth: 650 }} size="small" aria-label="a dense table">
```
julienw, line 179: Eventually (for a future PR) it would be good to have a proper aria-label naming this table.

src/components/CompareResults/CompareResultsTable.tsx
```diff
@@ -6,52 +10,237 @@ import TableContainer from '@mui/material/TableContainer';
...
+                              vertical: 'top',
+                              horizontal: 'right',
+                            }}
+                          >
+                            <FilterOptionsList
+                              options={options}
+                              column={headerId}
+                            />
+                          </Popover>
+                        </React.Fragment>
```
julienw, line 222: From our yesterday discussion, we've seen that it would be good for performance (and probably for readability as well) to extract this part to a separate component, so that any change to the state of the popover wouldn't render the full table again.

src/components/CompareResults/CompareResultsTable.tsx
```diff
@@ -6,52 +10,237 @@ import TableContainer from '@mui/material/TableContainer';
...
+
+  filter.platform.options = new Set(
+    compareResults.map((result) => result.platform),
+  );
+
+  const suiteOptions = compareResults.map((result) => result.suite);
+  filter.test.options = new Set([
+    ...compareResults.map((result) => result.test),
+    ...suiteOptions,
+  ]);
```
julienw, line 124: instead of changing the filter object above, I think I'd create the full filter object at each render. (in the extracted component as explained below). But I'm not so sure.
`active` doesn't seem used BTW, so maybe you can simplify the object:
```js
const filterOptions = {
  platform: new Set(...),
  test: ...,
  confidence: ...
};
```

julienw (commented):
I've added the comments from what we saw yesterday :) thanks!

kimberlythegeek (approved):
Looks good--thanks! I've filed this ticket to address later: [[PerfCompare] add 'on click' function to entire filter list item](https://mozilla-hub.atlassian.net/browse/FXP-2428)

esanuandra:
out of date

PR #230, opened 2022-09-01
Replace test and suite name

beatrice-acasandrei:


src/components/CompareResults/CompareResultsTable.tsx
```diff
@@ -12,15 +12,14 @@ import type { CompareResultsTableHeaders } from '../../types/types';
...
-  'Confidence',
-  'Total Runs',
+  { id: 'platform', label: 'Platform', align: 'center' },
+  { id: 'graph', label: 'Graph', align: 'center' },
+  { id: 'test-name', label: 'Test Name', align: 'left' },
+  { id: 'base-value', label: 'Base', align: 'center' },
+  { id: 'new-value', label: 'New', align: 'center' },
+  { id: 'delta-percent', label: 'Delta', align: 'center' },
+  { id: 'confidence', label: 'Confidence', align: 'center' },
+  { id: 'total-runs', label: 'Total Runs', align: 'center' },
```
eng-esther, line 22: hey, 

[question] : is there a reason why only test name is left?
How about making `align={header.align || 'center'}`

eng-esther (approved):

src/components/CompareResults/CompareResultsTable.tsx
beatrice-acasandrei: Original alignment is left for suite and test name and I've left it as it was before.

kimberlythegeek (approved):

PR #235, opened 2022-09-13
[PerfCompare] add note/alert for users to give feedback

beatrice-acasandrei:


kimberlythegeek:
I think this looks great. As far as testing, we can manually test for now, and file a ticket to add tests for this later.

I think it would be more visible if the button was at the top--when the search results are open, or when viewing the results page, the button gets pushed to the bottom of the page.

Maybe increase the auto-hide duration, and make the email address a link?

src/theme/components.js
```diff
@@ -191,6 +191,24 @@ const components = {
...
+          minHeight: '85vh',
+        },
+      },
+    },
+  },
+  MuiAlert: {
+    styleOverrides: {
+      root: {
+        '&.feedback-alert': {
+            backgroundImage: 'linear-gradient( 45deg, rgb(138,35,135) 10%, rgb(233,64,87) 50%, rgb(242,113,33) 100% )',
```
eng-esther, line 207: hey Andra, I didnt compile the code but does this infer that you're drawing an object here?
Would using an asset be good ?

src/common/constants.ts
```diff
@@ -7,6 +7,9 @@ export const treeherderBaseURL = 'https://treeherder.mozilla.org';
 
 export const maxRevisionsError = 'Maximum 4 revision(s).';
 
+export const userFeedbackMessage  = 'For any kind of suggestions please contact us at ';
```
eng-esther, line 10: nice

eng-esther (approved):

src/theme/components.js
beatrice-acasandrei: This is used to create a gradient background for the Alert, similar colors to the underline for the PerfCompare header, I left it so I get an opinion on how it looks- if not appropriate I'll change it back, this is why I added a separate commit for it.

src/components/Shared/ToggleDarkModeButton.tsx
```diff
@@ -7,12 +7,7 @@ import IconButton from '@mui/material/IconButton';
 function ToggleDarkMode(props: ToggleDarkModeProps) {
   const { theme, toggleColorMode } = props;
   return (
-    <Box
-      className="toggle-dark-mode"
-      display="flex"
-      justifyContent="flex-end"
-      alignItems="flex-end"
```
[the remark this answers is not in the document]
beatrice-acasandrei: I moved the styling to the outer box that contain both icons. Unfortunately the Box component does not allow to override the style.

kimberlythegeek (commented):
This looks great, thanks! However I noticed that the email link is missing the `href` value: `mailto:perfcompare-user-feedback@mozilla.com`

kimberlythegeek (approved):
lgtm--thanks!

PR #240, opened 2022-09-13
update snapshots

kimberlythegeek:
I'm unsure why this change happened, as far as I can tell the class names are now in a different order.


kimberlythegeek:
It appears that the snapshot tests are always failing because the material-ui classes are generated in a random order every time. I'm unsure what to do about this, or if we should consider disabling snapshot tests. Thoughts @beatrice-acasandrei @esanuandra ?

beatrice-acasandrei:
> It appears that the snapshot tests are always failing because the material-ui classes are generated in a random order every time. I'm unsure what to do about this, or if we should consider disabling snapshot tests. Thoughts @beatrice-acasandrei @esanuandra ?

We might attempt to find solutions for this, I saw that some issues with the generated classes can be fixed by creating our own `generateClassName`.

kimberlythegeek:
I'm closing this PR, as the latest test runs don't seem to be having this issue. I will create a ticket if the problem arises again.

PR #247, opened 2022-09-20
pagination

kimberlythegeek:


src/components/CompareResults/PaginatedCompareResults.tsx
```diff
@@ -0,0 +1,180 @@
+import { useState } from 'react';
```
kimberlythegeek, line 1: Most of the code in this component came directly from an example in the material ui docs: https://mui.com/material-ui/react-table/#custom-pagination-actions

src/components/CompareResults/CompareResultsTable.tsx
```diff
@@ -39,17 +34,7 @@ function CompareResultsTable(props: CompareResultsProps) {
...
-            compareResults.map((result, index) => (
-              <CompareResultsTableRow
-                key={index}
-                result={result}
-                index={index}
-                mode={mode}
-              />
-            ))}
-        </TableBody>
+        <PaginatedCompareResults mode={mode} />
```
beatrice-acasandrei, line 37: Initially the design included the pagination as an extra feature that can be enabled or disabled by users. Is this still the case?

src/components/CompareResults/PaginatedCompareResults.tsx
```diff
@@ -0,0 +1,180 @@
...
+          />
+        ))}
+        {emptyRows > 0 && (
+          <TableRow style={{ height: 53 * emptyRows }}>
+            <TableCell colSpan={6} />
+          </TableRow>
+        )}
+      </TableBody>
+      <TableFooter>
+        <TableRow>
```
esanuandra, line 153: When I click on the buttons it keeps changing position in the middle of the row because the table size changes. Maybe we could move the inside of this row to stay only on the right. What do you think?

src/components/CompareResults/CompareResultsTable.tsx
kimberlythegeek: There is an option in the 'results per row' to include 'All'. does this suffice?

src/components/CompareResults/PaginatedCompareResults.tsx
kimberlythegeek: :+1: turns out this was an easy fix, in the example I added, there's a prop `colSpan={3}`, which was probably the number of columns in the example. I changed this to match our # of columns and it is aligned to the right.

beatrice-acasandrei (approved):
Looks good! I have one suggestion though, if we don't have data I think we shouldn't show the pagination.

kimberlythegeek:
> Looks good! I have one suggestion though, if we don't have data I think we shouldn't show the pagination.

With the loading icon PR, it also displays an error message instead of the table if the api call returns no results. I can double-check to verify this.

PR #248, opened 2022-09-20
fix confidence tooltip

kimberlythegeek:


kimberlythegeek:
image
whenever confidence is unknown, it throws off the alignment of the cell slightly. after several of these in one table, the misalignment becomes much more apparent.

unsure of how I will resolve this, I will likely take a different approach than the one in this patch

kimberlythegeek:
will revisit this at a later time

PR #253, opened 2022-10-06
[PerfCompare] disable comparing 1 revision, or more than 2

beatrice-acasandrei:


kimberlythegeek (commented):
Instead of displaying a warning when adding revisions, perhaps we should display the warning when clicking the 'compare' button.

One issue I see is: a user searches for a revision, adds it, then enters a different search term (or selects from recent revisions), they would not be able to do so.

src/hooks/useSelectRevision.ts
```diff
@@ -36,17 +36,18 @@ const useSelectRevision = () => {
   const addSelectedRevisions = () => {
     const newSelected = [...selectedRevisions];
 
+    // Do not allow more then four total revisions (total, including the ones to be added)
```
kimberlythegeek, line 39: thanks for fixing this!

kimberlythegeek (approved):
looks good to me--thanks!

PR #256, opened 2022-10-11
add test coverage for compare api call

kimberlythegeek:


kimberlythegeek:
I will add tests for the uncovered lines in a separate patch. This requires reworking the `useEffect` call. Currently an API call is made using the URL params, whether or not they are different from currently selected revisions. I will address this in https://mozilla-hub.atlassian.net/jira/software/c/projects/FXP/boards/492/backlog?view=detail&selectedIssue=FXP-2429&issueLimit=100

beatrice-acasandrei (approved):

PR #263, opened 2022-10-27
Add filters on results table using endpoint data

esanuandra:


kimberlythegeek:
Since the framework selection is not functional in this patch, perhaps it should be removed. I'm working on a patch for that now: https://mozilla-hub.atlassian.net/browse/FXP-2397

also, should the 'apply' icon be at the top instead? or both top and bottom? just thinking that if the list is long, a user would have to scroll to the bottom to apply filters.

I think the `onClick` function should be applied to the entire list item. since we have `ListItemButton`, if a user clicks the list item it displays an animation that makes it appear to be 'clickable'. 

happy to discuss!

src/components/CompareResults/CompareResultsTable.tsx
```diff
@@ -4,49 +4,39 @@ import CircularProgress from '@mui/material/CircularProgress';
...
-  { id: 'delta-percent', label: 'Delta', align: 'center' },
-  { id: 'confidence', label: 'Confidence', align: 'center' },
-  { id: 'total-runs', label: 'Total Runs', align: 'center' },
-];
-
 function CompareResultsTable(props: CompareResultsProps) {
   const { mode } = props;
   const compareResults: CompareResultsState = useAppSelector(
-    (state) => state.compareResults,
+    (state: RootState) => state.compareResults,
```
kimberlythegeek, line 21: since we are using typed hooks (`useAppDispatch` and `useAppSelector`) you shouldn't need to define a type here: https://redux-toolkit.js.org/tutorials/typescript#define-typed-hooks

are you getting an error in your editor or lint command?

src/components/CompareResults/CompareResultsTableHead.tsx
```diff
@@ -0,0 +1,171 @@
...
+    event: React.MouseEvent<HTMLButtonElement>,
+    key: string,
+  ) => {
+    setOpenFilter(key);
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleCloseOptions = () => {
+    setAnchorEl(null);
+    setOpenFilter(null);
```
kimberlythegeek, line 113: any ideas how to add test coverage for this? I can take a look if that helps

src/components/CompareResults/CompareResultsTableHead.tsx
```diff
@@ -0,0 +1,171 @@
...
+            return (
+              <TableCell key={index} align={align}>
+                {label}
+                {filterKeys.includes(headerId) && (
+                  <React.Fragment>
+                    <IconButton
+                      data-testid={`${headerId}-options-button`}
+                      onClick={(e) => handleOpenOptions(e, headerId)}
+                    >
+                      <MoreVertIcon />
```
kimberlythegeek, line 138: I'm happy to defer to your judgement or user feedback, but my thought are that an icon like `FilterList` or `Tune` (https://mui.com/material-ui/material-icons/?query=filter&selected=FilterList) might better convey its purpose? Thoughts?

also could you file an issue for setting column widths so that the icon is always next to, rather than below the header?

kimberlythegeek (commented):
To summarize feedback: 
- When using `useAppSelector` hook, you shouldn't need to declare a type (`RootState`). if you're getting an error in the lint command or your editor I'm happy to look into it for you. (sometimes I get errors in my editor that do not show up when running `npm run dev` or `npm run lint`. it bugs the crap out of me :zany_face: )
- If you'd prefer to file a separate ticket to add test coverage for the new features, I'm okay with that. If there's anything in particular you're unsure how to test I can take a look and see if I can help.
- the icon for the filter seems a little unclear to me what its purpose is. However, we can defer to the UX designer or user feedback in the future to get other opinions on this. This patch has been open for a while due to PTO and work week, and I don't want to be overly picky because the functionality works as it should.
- for some columns the filter icon displays below the header text rather than next to it. I realize this is a big patch already, so you can address this in another patch or file an issue in github as "good first issue", I don't have a preference. (This may also be a screen size issue. I have external monitors that are smaller than my laptop screen, so perhaps a "responsiveness" ticket may be more appropriate? may be worth discussing in our next meeting?)

I really like the filter status that you've added, especially the "filter options have changed. apply the new filters" alert :100: 

the only thing I would say should be changed before landing is the `RootState` type declaration. everything else I believe can be address or discussed later. Feel free to request feedback from Julien as he is more knowledgeable in this area than I am.

Also very impressed with the improvement in performance--the filtering is much faster than the original patch. Thanks!!

src/components/CompareResults/CompareResultsTable.tsx
esanuandra: I was getting the errors from the editor (Visual Studio Code). It seems to happen to other people as well. https://stackoverflow.com/questions/51891407/ts-lint-in-visual-studio-code-suddenly-indicates-wrong-lint-errors

src/components/CompareResults/CompareResultsTableHead.tsx
esanuandra: Some help on how to test this would be great. Thank you

src/components/CompareResults/CompareResultsTableHead.tsx
esanuandra: I agree with the `FilterList` icon.

[Here is the issue ticket for the width](https://mozilla-hub.atlassian.net/browse/FXP-2471)

src/components/CompareResults/CompareResultsTable.tsx
kimberlythegeek: I've had problems with this at times also. which linter extension are you using?

esanuandra: I have the eslint extension and the @builtin extensions for typescript in vscode. The npm run lint doesn't give any linting error.

kimberlythegeek (approved):

PR #266, opened 2022-11-02
add framework selection dropdown

kimberlythegeek:


kimberlythegeek:
Coverage report: https://app.codecov.io/github/mozilla/perfcompare/commit/565eb1b741c9327654e5a3015635b5857ef792ed

beatrice-acasandrei (commented):
@kimberlythegeek If you select a framework with results and then select a framework that doesn't have data, the results from the previous choice are still displayed along with the error "Error: No results found".

kimberlythegeek:
> 

thanks for catching this! I've updated the `setFramework` reducer to clear results when called. I also moved the framework dropdown outside of the compare results table, because if the initial framework (talos) had no results, a user would not be able to switch to another framework.

kimberlythegeek:
I noticed that the dropdown will display even while the results are loading, so I'll need to address this

kimberlythegeek:
coverage report: https://app.codecov.io/github/mozilla/perfcompare/commit/8b617d94d0e77f02212ce6843e209c1ea01cb047

beatrice-acasandrei:
@kimberlythegeek Possible bug: If there's no data for a particular framework then the dropdown (for the frameworks) should still be displayed.

kimberlythegeek:
@beatrice-acasandrei I apologize--I thought I had resolved the issue but evidently I did not :facepalm:

PR #269, opened 2022-11-04
add status column

kimberlythegeek:


kimberlythegeek:
I also tried showing the icon next to the delta value, rather than adding a new column, but this change the alignment of the text for that cell because the icon font size is larger than the text font size. when I reduced the size of the icon font, it was too small to see clearly, but maybe there's another way to do this. 

I think it could be confusing to have a column status that is usually empty. I also tested with showing a "neutral" icon, in this case a dash (material ui's Remove icon), but I think I'd prefer only showing icons for significant improvements or regressions

beatrice-acasandrei:
> I also tried showing the icon next to the delta value, rather than adding a new column, but this change the alignment of the text for that cell because the icon font size is larger than the text font size. when I reduced the size of the icon font, it was too small to see clearly, but maybe there's another way to do this.
> 
> I think it could be confusing to have a column status that is usually empty. I also tested with showing a "neutral" icon, in this case a dash (material ui's Remove icon), but I think I'd prefer only showing icons for significant improvements or regressions

I agree that having an empty column might be confusing, we can have it like this for now but I definitely think we should revisit this and try to add it next to the delta value.
@esanuandra what do you think?

kimberlythegeek:
@beatrice-acasandrei sounds good to me. I've filed https://mozilla-hub.atlassian.net/browse/FXP-2473 for this

beatrice-acasandrei (approved):

esanuandra (approved):

PR #271, opened 2022-11-07
bump version to 0.9.0

kimberlythegeek:


kimberlythegeek:
attempting a different method for handling releases based on https://nvie.com/posts/a-successful-git-branching-model/


hopefully this resolves the issue where every merge to master showed all commits since the staging branch was created. I've set the release to 0.9.0 because the app is mostly functional/usable, but users are not yet able to select a framework.

PR #272, opened 2022-11-08
set selected revisions

kimberlythegeek:


Carla-Moz (commented):
If you go to your selected branch, I found three areas to add code to make sure the selected revisions table always shows on refresh and paste. I'm happy to discuss in person to see if this could be a possible solution. 

1.) in ```src/hooks/useHandleChangeSearch``` add ```searchByRevisionOrEmail``` to return list on line 67

2.) import ```useHandleChangeSearch``` in ```src/components/CompareResults/CompareResultsView```

and get ```  const { searchByRevisionOrEmail } = useHandleChangeSearch(); ```

add the following api hit to the ```useEffect``` and make sure to add an empty``` []``` as second argument in ```useEffect``` call. 

	if (revs && repos) {
			revs.forEach((rev, index) => {
				void searchByRevisionOrEmail(repos[index] as Repository['name'], rev);
			});
		}
```
3.) update the ```SelectedRevisions``` reducer with an extra reducer:

```
	extraReducers: (builder) => {
		builder.addCase(fetchRevisionByID.fulfilled, (state, action) => {
				state.revisions = state.revisions.concat(action.payload[0]).filter((revision, index, self) => self.findIndex((r) => r.id === revision.id) === index);
		});
	},
```

Issue #257, opened 2022-10-13
PerfCompare header underline resizes based on window size


kimberlythegeek:
The underline should stay the same regardless of window size

kimberlythegeek:
Of course 😊

Thanks for your interest in contributing!

GPain2007:
Yes I am. Still working on this.

kimberlythegeek:
cool, thanks! let us know if you need anything from us.

GPain2007:
Will do!

GPain2007:
<img width="1512" alt="Screen Shot 2022-11-08 at 7 55 06 PM" src="https://user-images.githubusercontent.com/97716889/200717936-bdcb9faa-16f9-4a99-89c4-095903f3ed00.png">

GPain2007:
<img width="1512" alt="Screen Shot 2022-11-08 at 7 54 57 PM" src="https://user-images.githubusercontent.com/97716889/200717967-87ab95a9-aa31-4194-a459-39df8e875908.png">
<img width="1512" alt="Screen Shot 2022-11-08 at 7 54 50 PM" src="https://user-images.githubusercontent.com/97716889/200717979-672ee22c-2b3d-4c5e-afc8-3f5f7e2a3559.png">

GPain2007:
I fixed the issue. Please let me know if this is what they want. If so i will push the code now. Also do i push to a branch?

kimberlythegeek:
Please open a pull request to the `staging` branch and we will test and review it :) 

Thanks!

kimberlythegeek:
fixed by #275

PR #275, opened 2022-11-09
 #257 PerfCompare header underline resizes based on window size

GPain2007:
Changed backgroundSize from "25%" to "290px". 

Does not change image size on screen resizing.
Screen Shot 2022-11-08 at 7 55 06 PM
Screen Shot 2022-11-08 at 7 54 57 PM
Screen Shot 2022-11-08 at 7 54 50 PM

### DIFF
--- a/src/theme/components.js
+++ b/src/theme/components.js
@@ -130,7 +130,7 @@ const components = {
             backgroundImage: `url(${zap})`,
             backgroundPosition: '55%',
             backgroundRepeat: 'no-repeat',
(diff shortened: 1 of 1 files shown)


kimberlythegeek (approved):
Looks good to me! thanks! As far as smaller screen sizes, the underline is bigger than the text, but we won't be focusing on responsive/mobile design for some time (if we decide it's something we want to do)

Mostly we're considered with different desktop screen sizes, and that looks good :) Thanks!

PR #276, opened 2022-11-09
Feat/set selected revisions

kimberlythegeek:
Taking a different approach here. I know the code is spaghetti but I at least have gotten it to work, sort of. One of the problems I'm facing is that when fetching the revisions from the API to set as `selectedRevisions`, the order of the revisions depends on which fetch call completes first.

This is an issue, because by default the first revision is `base` and any others are new. 

I appreciate any feedback, I know it's ugly but I had a hard time figuring out how best to implement this with our current structure.

### DIFF
--- a/src/components/CompareResults/CompareResultsViewInit.tsx
+++ b/src/components/CompareResults/CompareResultsViewInit.tsx
@@ -1,25 +1,21 @@
 import { useEffect, useState } from 'react';
 
-import { useSnackbar, VariantType } from 'notistack';
 import { useParams } from 'react-router-dom';
 
+import { repoMap } from '../../common/constants';
 import { useAppDispatch, useAppSelector } from '../../hooks/app';
 import useFetchCompareResults from '../../hooks/useFetchCompareResults';
-import {
-  clearSelectedRevisions,
-  setSelectedRevisions,
-} from '../../reducers/SelectedRevisions';
-import { fetchSelectedRevisions } from '../../thunks/selectedRevisionsThunk';
-import type { Revision } from '../../types/state';
+import { setSelectedRevisions } from '../../reducers/SelectedRevisions';
 import { fetchCompareResults } from '../../thunks/compareResultsThunk';
-import { repoMap } from '../../common/constants';
-// component to fetch set selected revisions when Compare View is loaded
+import { fetchSelectedRevisions } from '../../thunks/selectedRevisionsThunk';
+
+// component to fetch set selectedRevs revisions when Compare View is loaded
 function CompareResultsViewInit() {
   const dispatch = useAppDispatch();
   const selectedRevisions = useAppSelector(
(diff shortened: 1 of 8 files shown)


src/components/CompareResults/CompareResultsViewInit.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+import { repoMap } from '../../common/constants';
+import { useAppDispatch, useAppSelector } from '../../hooks/app';
+import useFetchCompareResults from '../../hooks/useFetchCompareResults';
+import { setSelectedRevisions } from '../../reducers/SelectedRevisions';
+import { fetchCompareResults } from '../../thunks/compareResultsThunk';
+import { fetchSelectedRevisions } from '../../thunks/selectedRevisionsThunk';
+
+// component to fetch set selectedRevs revisions when Compare View is loaded
+function CompareResultsViewInit() {
```
julienw, line 13: The variable names are a bit confusing to me. Maybe put the type in their names for a better understanding, eg: revisionList vs revisionCommitList vs revisionsAsString

src/components/CompareResults/CompareResultsViewInit.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const [newSelected, setNewSelected] = useState([]);
+
+  useEffect(() => {
+    let isSubscribed = true;
+
+    const fetchData = async (repo: string, rev: string) => {
+      const data = await dispatch(fetchSelectedRevisions({ repo, rev }));
+
+      if (isSubscribed) {
+        setNewSelected((prev) => [...prev, data.payload]);
```
julienw, line 39: if the problem is the order, why don't you use a Map to store the results? Then the order would be in `selectedRevs`?

src/thunks/selectedRevisionsThunk.ts
```diff
@@ -0,0 +1,31 @@
...
+  { repo: string; rev: string },
+  { rejectValue: string }
+>(
+  'selectedRevisions/fetchSelectedRevisions',
+  async ({ repo, rev }, { rejectWithValue }) => {
+    let response;
+
+    try {
+      response = await fetch(
+        `${treeherderBaseURL}/api/project/${repo}/push/?revision=${rev}`,
```
julienw, line 18: When I see a concatenation like this, I think we need to check that `repo` and `rev` can be only alphanumeric strings + comma like they should, because my understanding is that they can be changed easily in the URL.

julienw (commented):
I'm not sure I understand everything but I tried to give a few comments.

In the future, please write some STR to explain how we can try the code to test what's wrong. Also you mention "another approach", it would be good that you give some reference to the initial approach.

Also adding a comment to the specific part of the code where you'd like a feedback especially could be useful.

Please tell me if you need anything more from me!

PR #282, opened 2022-11-25
FXP-2467 Reset filter state after going back one page

esanuandra:


kimberlythegeek (approved):
looks good to me, thanks!

PR #291, opened 2022-12-15
FXP-2468 Fix table header is displayed when there's no data

esanuandra:
1. Select two revisions from try
2. Compare
3. If the comparison has no data, only an error should be displayed without the table header


kimberlythegeek (approved):
thanks!

PR #300, opened 2023-03-02
Style themes

Carla-Moz:
- Have a custom “library” (folder, or classes) with the components that we are going to use that are styled and respect the custom design
- Allow dark theme on these components
- Create strings folder for the views, components, and links


beatrice-acasandrei (approved):
Looks good! It just needs some lint fixing :) Thanks

esanuandra (approved):
Nice work!

julienw:
🥳

Issue #125, opened 2022-05-27
add file-loader to handle SVG files and remove fileTransformer.js 


kimberlythegeek:
from Nazim:
it looks like we are only using this fileTransformer for svg files. During webpack compilation, we can tell the webpack how to load a certain file type. For svg files, we can tell the webpack to load it as a "file" with a "file-loader". Here's how we are doing it on profiler codebase: https://github.com/firefox-devtools/profiler/blob/fe13f1f5e9d2205b2b6bde46d3080e7be3d277e0/webpack.config.js#L65-L68 This is the documentation for the loader: https://v4.webpack.js.org/loaders/file-loader/ when you add this loader, you won't have to have this transform, so you can remove it from package.json as well

anna-kulyk:
@kimberlythegeek hi, I am outreachy applicant and would like to be assigned to this issue

anna-kulyk:
@kimberlythegeek @Carla-Moz 

unfortunately installing file-loader doesn't help with jest, it fails with this error:

`       SyntaxError: Unexpected token '<'`

       8 | import osx from './img/osx.svg';
       9 | import windows from './img/windows.svg';
    > 10 | import zap from './img/zap-10.svg';
         |                                    ^
      11 |
      12 | const components = {
      13 |   MuiButton: {

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1796:14)
      at Object.<anonymous> (src/theme/components.js:10:58)

I did a little research and as far as I can tell, the only way to remove fileTransformer.js is to install jest-transformer-svg, but it basically does the same thing as fileTransformer.

PR #318, opened 2023-03-08
Fixed Confidence tooltip breaking UI and added some SVG loader to webpack to solve the import of SVGs  #258

### LINKED ISSUE #258: Confidence tooltip breaks UI
when there is a result that has unknown confidence, the alignment of the cell is disrupted, and after several of these in the same table, the difference becomes very apparent:

image

one suggestion would be to use SVG images directly and wrapping them in a `Tooltip` component, rather than wrapping the Table Cell, which seems to be where the issue arises.

When adding these icons originally, I encountered issues with using SVG images in typescript, which is why the background image approach was taken. Since ejecting and switching to webpack, this may no longer be an issue

Bukolab99:
Fixes #258 , My changes was to ensure that the TableCell component doesn't break the UI if result.confidence_text is null and to change the style for the missing-confidence-button class so that the IconButton component is centered vertically inside the TableCell component. To achieve this, the following steps were taken:

To prevent the TableCell component from breaking the UI, I  use a conditional statement to check if the result.confidence_text is null. If it is null, we can set a default value to be displayed in the TableCell component.

To center the IconButton component vertically inside the TableCell component, we can add the padding-top: 0px style property to the missing-confidence-button class.

Once the changes are made, the snapshot files should update to reflect the new UI.
And also fixed @NikolaDojic's follow up [comment ](https://github.com/mozilla/perfcompare/issues/258#issuecomment-1403332822)
With the changes I made, you should now be able to import and use SVGs in your React codebase with the help of the added SVGR loader. @kimberlythegeek Please review.


NikolaDojic:
I don't think this will resolve the #258 issue. Fixing the broken layout is not part of the changes.
However, it potentially resolves https://github.com/mozilla/perfcompare/issues/125.

Bukolab99:
> 

So I need to fix the broken layout to completely resolve issue #258 ?

NikolaDojic:
> > 
> 
> So I need to fix the broken layout to completely resolve issue #258 ?
I would assume so since the title of the #258 issue is _Confidence tooltip breaks UI_

anna-kulyk:
@NikolaDojic #125 was assigned to me recently. And it says to install file-loader. Do I have to stop working on it? Does svgr is a better choice?

Bukolab99:
> choice?

I think any applicant can work on any issue. It’s just going to be a matter of comparison.

NikolaDojic:
Just to be clear guys, I'm not associated with Mozilla in any way other than personal fondness, and some occasional code contribution.
All I can do is give my opinion and/or advice, but it is important that you understand where is this coming from.

That being said I would personally go for the implementation that is similar to the one already in the codebase. In this case, that means going with `file-loader`, based on the description from the tickets. 

Even if this kind of PR does not end up in the code base it is not a big waste ;)
In addition, if your only goal is to demonstrate your skill, then it does not matter if it ends up in the master branch.

Issue #267, opened 2022-11-03
columns in results view are not a consistent width


kimberlythegeek:
depending on the values in the tables, columns are of varying widths. ensure columns stay the same width

Phyya:
Hi,
I am an outreachy applicant. Can I be assigned to this issue so that I can start working on it? Thanks.

mayangiaretha:
Hey @kimberlythegeek, I am an outreachy applicant. If possible I would love to be assigned this issue or any other available issue, thanks.

Safei-Ashraf:
@kimberlythegeek  Would love to give this a try

siddhi-244:
Hello! I am an Outreachy applicant and would like to work on this issue !

Ayesha-Solanki15:
@kimberlythegeek, I would like to try my hands on this.

priyasarpate:
Hello @kimberlythegeek,

I would love to work on this!

sumairq:
Hi @kimberlythegeek ,
I am an Outreachy applicant. I would love to be assigned to this issue so that I can apply my frontend-skills to make the columns a consistent width.

Businge931:
Hi @kimberlythegeek please assign to this issue

satyamksharma:
Hi @kimberlythegeek I would like to work on thus issue and request you to assign me to this issue. Thank you

mayangiaretha:
Resolved in PR #319

PR #319, opened 2023-03-08
Fix columns in results view are not a consistent width #267

mayangiaretha:
**Related Issue**
- #267

**Screenshots**
Original                         | Updated
:------------------------------: | :------------------------------------:
 before | after


kimberlythegeek (commented):
results look good, thank you! however, we prefer that measurements use relative sizing rather than static sizes. We want all CSS/styling changes to be in the theme files rather than inline in components. 

We have chosen our intern based on earliest submission, and demonstration of skills. Thank you for your contribution.

mayangiaretha:
Thanks so much for the opportunity @kimberlythegeek

PR #320, opened 2023-03-08
fix/Table Row Close Button [accessibility issues]

Safei-Ashraf:
Fixing issue #321 

1- `Close Buttons` in `SelectedRevisionTableRow` have duplicated Id's => all elements got the same id `"close-button"`

2- Missing label or meaningful text

**Fix:** 

1- Added row id to the string to make it unique (only one button per row is used + row id is already in the component)

~~2- Updated test files  from : querySelectorAll('#close-button') to be => with test id + added data-testid="closeButton" to the button~~

- **_Another Error appeared about missing label or meaningful text and was resolved by_**
- Added `aria-label` to be "Close" 
- Updated snapshots

**Update**: 

2- I just replaced class/testId with querySelectorAll('aria attribute) that I added earlier on the first fix, this achieves the same result and removes extra attrib on the button.


Safei-Ashraf:
solving the issue in #321

kimberlythegeek (approved):
awesome, thank you!

Issue #225, opened 2022-08-23
fix width of edit revision popover


kimberlythegeek:
1. click input on search page
2. select two recent revisions and click `+` button 
3. click `compare`
4. on the results view, click the edit button (looks like a pencil) next to one of the revisions
5. the search popover is not the same length as the selected revisions table

cybraia:
Hi Kimberly, I would like to work on this. Can you please give some more details about it?

sabbCodes:
Hello, I would love to try this out. But, the description is missing.

kimberlythegeek:
@cybraia great, thanks! 

@sabbCodes FYI you can still work on this issue, despite it being assigned to Krishna. This can be a valuable way to compare applicants. It would help to send me a private message (@kimberlythegeek on https://chat.mozilla.org/#/login) so that I can better keep track of applicants and the issues they are working on.

kimberlythegeek:
1. click input on search page
2. select two recent revisions and click `+` button 
3. click `compare`
4. on the results view, click the edit button (looks like a pencil) next to one of the revisions
5. the search popover is not the same length as the selected revisions table

kimberlythegeek:
@cybraia @sabbCodes FYI, all new PRs should be submitted to the `beta` branch, not `staging`. it should do this by default.

Thanks!

okerekechinweotito:
...Also working on this

vinayakjaas:
Hii, Please assign  this task to me I try to solve this bugs

Businge931:
Hi @kimberlythegeek, I wanna work on this bug also. Kindly assign me to it

PR #323, opened 2023-03-09
fix width of edit revision popover #225

sabbCodes:
I've solved the width issue #225 with the popover using the PaperProps property.

Below are the screenshots of the before and after the bug fixing.
app before

app after


kimberlythegeek (commented):
while the changes look good on some screen sizes, on larger and smaller screens the width is smaller or larger than it should be. 

We have chosen our intern based on the earliest submissions and demonstration of skills, thank you for your contribution!

Issue #169, opened 2022-06-24
Add warning message when comparing builds from different projects


kimberlythegeek:
This may be done in the Snackbar alert, or a message displayed above or below the selected revisions table. Needs discussion

Jojowangfy:
Hi Kimberly, may I give this a try? @kimberlythegeek

sabbCodes:
Hi @kimberlythegeek, I'd like to give this a try please.

tessa5:
Hey @kimberlythegeek, I would like to contribute to this please.

ipriyanshi1708:
Hii @kimberlythegeek , I am an outreachy applicant, I would like to give it a try. Can you please assign it to me?

Thanks!

sabbCodes:
Added warning when comparing different revisions from different repositories. https://github.com/mozilla/perfcompare/pull/368

Jojowangfy:
Hi @kimberlythegeek , I've added the feature but I'm not really good at writing test. So sorry for that. It's in #447

PR #324, opened 2023-03-09
feat/Added warning when comparing different builds

Gathu17:
I added a warning snackbar message when comparing different builds. 
Issue #169

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ Performance Comparison Tool
 
 ![screenshot](screenshot.png)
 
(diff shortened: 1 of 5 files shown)


kimberlythegeek (commented):
Ideally we'd prefer that the warning appears before clicking the 'compare' button, rather than afterward, but we did not specify this in the issue. 

We had a very large number of applicants, and it took some time to review the submissions. We have selected the applicant for this internship. Thank you for your contribution!

PR #326, opened 2023-03-09
Fix- Columns in results view are not a consistent width #267

sumairq:
# In this PR ✋ 
@Carla-Moz @kimberlythegeek 

I have specified widths for the columns so that the widths stay consistent and the layout remains responsive.

These changes are to ensure that the width of the columns stays consistent and does not change depending on the data contained within the columns.

This includes adding the MUI colgroup in CompareResultsTable.tsx component and specifying column widths in percentages.

image  |  image

### Issue
#267


kimberlythegeek (approved):
Looks good to me! thank you for using relative sizing rather than static widths.

sumairq:
Thanks a lot @kimberlythegeek  🥳

Issue #327, opened 2023-03-09
Alignment for table data is not consistent with the alignment of the table head in the results view


sumairq:
I think the data within the columns should have the same alignment as the heading to give it a more streamlined look.

![image](https://user-images.githubusercontent.com/60618877/223978726-f0390f93-8da7-48be-9965-65115cb824c9.png)

sumairq:
Your feedback is highly appreciated 🙌

tobidosumu:
Hello @Carla-Moz. I would like to resolve this issue. Thanks.

tobidosumu:
@sumairq. Could you add some details (e.g. a link) about where to locate this issue in the view?

sumairq:
Hi @tobidosumu You can locate this problem in the results view table. 

![Animation](https://user-images.githubusercontent.com/60618877/224350945-b3526dcf-6211-4b6b-b4e7-710a1a25cb02.gif)

tobidosumu:
@sumairq I get it now. Thanks

tobidosumu:
@Carla-Moz Hello @kimberlythegeek 
Hello. I have submitted a Pull Request to this issue.

PR #328, opened 2023-03-09
Add alignment to the table data that is consistent with the table head #327

sumairq:
# In this PR ✋ 
@Carla-Moz @kimberlythegeek 

I have added alignment for the data within the columns so that it stays consistent with table headings.

The goal of these changes is to add visual symmetry between the table heads and table data.

This includes adding the MUI `align` prop to `<TableCell>` component and specifying column widths in percentages.

image  |  image

### Issue
#327


kimberlythegeek (approved):
We plan to move all inline styles to the material UI theme, but this is acceptable for now.

looks good to me, thanks!

kimberlythegeek:
@sumairq there is another PR for this that was subitted before yours, so I have merged that one instead. Thanks!

Issue #330, opened 2023-03-09
Unknown confidence icon becomes disconnected from the rest of the table cell on smaller screen sizes in results view.


sumairq:
The 'unknown confidence' icon becomes disjointed from the base of the table cell on smaller widths.
- It behaves differently from the rest because it is from the MUI library.
-  The current code sets it as a flex element to align it in the middle of the table cell.  This becomes a problem when the height of the table cell increases with the decrease in the width of the screen as expected it sticks to the top.   

950px screen width            | 
:-------------------------:|
![image](https://user-images.githubusercontent.com/60618877/224042642-f65d935c-b319-47f1-b2f5-3773e7809b64.png)  |

sumairq:
Thank you for pointing that out @anna-kulyk 🙌 . Let me close this one.

PR #331, opened 2023-03-09
Refactor unknown icon #330

sumairq:
# In this PR ✋ 
@Carla-Moz @kimberlythegeek 

I have refactored the alignment of the 'unknown confidence' icon.

The goal of these changes is to fix the dislocation of the 'unknown confidence' icon at smaller screen widths.

This includes adding the MUI `align` prop to `<TableCell>` component that contains the unknown confidence icon and removing the `display: flex` setting from it's styling.

image  |  image

### Issue
#330


kimberlythegeek (approved):
the issue appears to be resolved. thank you!

kimberlythegeek (commented):
I have noticed that pr #336 has also solved the issue, but added these settings to the theme rather than inline styling. We prefer styling to be in the theme configuration. We will be accepting that PR over this one.

sumairq:
Thank you @kimberlythegeek. 🙌  Will keep this preference in mind for the future. ✍️

PR #336, opened 2023-03-09
Confidence tooltip breaks UI #258

anna-kulyk:
# Summary
 Fixed confidence tooltip that was breaking UI
 #258
 Outreachy applicant
 
 ## Before:
 `TableCell` component was breaking UI when `result.confidence_text` was null  
 
 ## Changes:
 This pull request:
 
 * Changes style for `unknown-confidence` class so that `TableCell` component doesn't break UI if `result.confidence_text` is null.
 * Changes style for `missing-confidence-button` class (adds padding-top: 0px) so that `IconButton` component is centered vertically inside the `TableCell` component.
 * Updates snapshot files.
 
 ## After:
 * The confidence tooltip is no longer breaking UI.
 * After updating snapshots, tests run with `npm run test` are passing.
 * Code style and format is consistent with the rest of the repository as confirmed with `npm run lint` and `npm run format:check`.
 
@kimberlythegeek @Carla-Moz 

Screenshot

### DIFF
--- a/src/__tests__/CompareResults/__snapshots__/CompareResultsView.test.tsx.snap
+++ b/src/__tests__/CompareResults/__snapshots__/CompareResultsView.test.tsx.snap
@@ -17,12 +17,12 @@ exports[`CompareResults View Should match snapshot 1`] = `
             class="MuiTableRow-root MuiTableRow-head css-1eza0l0-MuiTableRow-root"
           >
             <th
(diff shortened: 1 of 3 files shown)


kimberlythegeek (approved):
looks good to me, thank you for adding your changes to the theme instead of inline styling. Thank you for your contribution!

anna-kulyk:
> looks good to me, thank you for adding your changes to the theme instead of inline styling. Thank you for your contribution!

Thank you for the feedback!

PR #342, opened 2023-03-10
Fixed the responsiveness

sabbCodes:
I noticed there's something wrong with the responsiveness, as the layout is broken in mobile and small screens. I fixed it by adding responsiveness that keeps the app in form in any view, even at extremely small screens, still preserving the original layout in the desktop view.


sabbCodes:
I noticed there's something wrong with the responsiveness, as the layout is broken in mobile and small screens. I fixed it by adding responsiveness that keeps the app in form in any view, even at extremely small screens, still preserving the original layout in the desktop view.

Below are the before and after images of the app;
mobView problem
mobView soln

Issue #226, opened 2022-08-23
checked revision is not cleared after editing selected revision in compare results view


kimberlythegeek:

beatrice-acasandrei:
@kimberlythegeek Is this still reproducing? Could you add some steps to reproduce?

kimberlythegeek:
yes, it is. 

STR:
1) add two revisions 
2) click "compare'
3) on results page, click edit button
4) focus search input
5) check a result and click `+` add button
6) click edit button again, and focus input

previously checked result is still checked

Formasitchijoh:
@kimberlythegeek  could i work on this please

tessa5:
Hi, @kimberlythegeek please assign this to me too please.

vinayakjaas:
Hii ,Please assign this task to me I try to solve this issue

OndigoWendy:
Hi @kimberlythegeek @Carla-Moz 
Kindly check out my PR on this issue https://github.com/mozilla/perfcompare/pull/345

Businge931:
Assign me to this issue please

PR #345, opened 2023-03-10
Fix - checked revision not cleared after editing selected revision in compare results view #226

OndigoWendy:
Issue: https://github.com/mozilla/perfcompare/issues/226

The new implementation imports the checkedRevisions reducer, adds the dispatchSetCheckedRevisions function as a prop to the CompareResultsView component and uses it in the useEffect hook to clear the checked revisions when the revisions prop changes.

In the previous implementation, the CompareResultsView component did not clear the checked revisions when the revisions prop changed. This caused a bug where the checked revisions were not cleared after editing selected revisions in the compare results view.

By adding the dispatchSetCheckedRevisions function and using it in the useEffect hook, the new implementation ensures that the checked revisions are cleared whenever the revisions prop changes, thereby fixing the bug.


src/components/CompareResults/CompareResultsView.tsx
```diff
@@ -24,16 +26,27 @@ function CompareResultsView(props: CompareResultsViewProps) {
...
+  }, [
+    revisions,
+    location.search,
+    dispatchFetchCompareResults,
+    dispatchSetCheckedRevisions,
+  ]);
 
   return (
-    <Container maxWidth="xl">
+    <Container maxWidth='xl'>
```
jrbenny35, line 44: Can you change these quotes to double quotes, as they were in the original code?

src/__tests__/Search/fetchRevisionByID.test.tsx
```diff
@@ -6,6 +6,7 @@ import getTestData from '../utils/fixtures';
 import { renderWithRouter, store } from '../utils/setupTests';
 import { screen } from '../utils/test-utils';
 
+jest.setTimeout(10000); // sets the timeout to 10 seconds
```
jrbenny35, line 9: Why did you set this timeout to 10 seconds?

src/__tests__/CompareResults/CompareResultsTable.test.tsx
```diff
@@ -463,7 +463,7 @@ describe('Compare Results Table', () => {
 
     popper = await waitFor(() => screen.getByRole('tooltip'));
     
-    expect(popper).toHaveAttribute('style', '');
+    expect(popper).toHaveAttribute('style', 'position: absolute; top: 0px; left: 0px; margin: 0px; right: 0px; transform: translate(0px, 0px);');
```
jrbenny35, line 466: What is this change doing?

jrbenny35 (changes requested):
This looks promising. I noticed you didn't include a test for this, can you please do that?

src/components/CompareResults/CompareResultsView.tsx
OndigoWendy: Okay...I missed that..making the change.

src/__tests__/Search/fetchRevisionByID.test.tsx
OndigoWendy: This was a change suggested on the logs...the previous timeout was causing some tests to fail.

src/__tests__/CompareResults/CompareResultsTable.test.tsx
OndigoWendy: This change was also suggested on the logs...a test was failing because the expected value expect(popper).toHaveAttribute('style', 'position: absolute; top: 0px; left: 0px; margin: 0px; right: 0px; transform: translate(0px, 0px);');  did not match the current value expect(popper).toHaveAttribute('style', '');

I will look into whether I can make the tests pass without the above changes.

OndigoWendy:
> This looks promising. I noticed you didn't include a test for this, can you please do that?

Alright...working on it.

OndigoWendy:
> adds

Issue #325, opened 2023-03-09
displays the name of the element when the cursor is hovering over it:


vinayakjaas:
**Sometimes It is difficult to understand symbolic  representation so we display name what does it work it is easy to understand.**
**I Solved this Issue Already So please assign this task to Me.**
![2023-03-09 (21)](https://user-images.githubusercontent.com/96948435/223967522-6f44059f-0a99-44ad-9587-d323c4d1b75e.png)

satyamksharma:
Hi there I would like to add this feature so please assign me

PR #349, opened 2023-03-10
displays the name of the element when the cursor is hovering over it: #325"

vinayakjaas:
**I Solve This Issue**
**Before**

**After**


src/components/styles.css
```diff
@@ -0,0 +1,45 @@
...
+    }
+    .MuiBox-root {
+      flex-direction: row;
+    }
+    .MuiSnackbarContent-root {
+      font-size: 18px;
+      padding: 16px;
+    }
+  }
+
```
jrbenny35, line 45: What do these changes do?

PR #360, opened 2023-03-11
Fixed confidence tooltip breaking UI

### LINKED ISSUE #258: Confidence tooltip breaks UI

Bukolab99:
Fixes #258 , My changes was to ensure that the TableCell component doesn't break the UI if result.confidence_text is null and to change the style for the missing-confidence-button class so that the IconButton component is centered vertically inside the TableCell component. To achieve this, the following steps were taken:

To prevent the TableCell component from breaking the UI, I use a conditional statement to check if the result.confidence_text is null. If it is null, we can set a default value to be displayed in the TableCell component.

To center the IconButton component vertically inside the TableCell component, we can add the padding-top: 0px style property to the missing-confidence-button class.

Once the changes are made, the snapshot files should update to reflect the new UI.


Bukolab99:
Hi @kimberlythegeek, I'm relatively new to using github and I just wrote a fix for issue (#258), But I have 4 of your commits in my pull request and I do not understand why.

PR #371, opened 2023-03-13
Removing state type

kelani34:
Replaced all references to State with RootState. This involved changing any type annotations, function parameters, and variable declarations that use State.

Update your Redux store setup code to use reducer instead of store.getState.

Tested the changes to ensure that everything is still working as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ Performance Comparison Tool
 
 ![screenshot](screenshot.png)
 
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -22,7 +25,7 @@ PerfCompare is hosted on heroku, and is updated every time commits are pushed to
 ## Setup
 
 ### Requirements
-
+- [git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
```
jrbenny35, line 28: The changes in this file shouldn't be included in this patch

jrbenny35 (changes requested):
Thanks for the PR @kelani34, I just had 1 nit. I'll let @kimberlythegeek take a closer look

Issue #372, opened 2023-03-13
Display revisions list upon selecting from dropdown


Formasitchijoh:
Having to focus on the input field before the revision list is displayed is a bit disturbing at time for most users .I propose the list should be displayed immediate the user selects the appropriate version from the drop down

Formasitchijoh:
@Carla-Moz  @kimberlythegeek  i would like to work on this please

Businge931:
@Carla-Moz and @kimberlythegeek I would like to work on this as well

Formasitchijoh:
@Businge931  we could work on it if you dont mind ,

Businge931:
Yes I agree. Lets wait for us to be assigned to it

Formasitchijoh:
We really dont have to wait to be assigned we just have to work on it then
 submit a PR and wait for it to be approved

Businge931:
Great then let's start on this ASAP!

Businge931:
Have you done something yet?

Formasitchijoh:
I am current ly working on it i could share with you my idea

Businge931:
Can we create a room and discuss it more in detail

Formasitchijoh:
that will be great

Businge931:
What platform do you suggest, Slack?

Formasitchijoh:
I surgest it be done here so others can learn and also share their ideas. What do you think

Businge931:
Okay great.  So shall we merge our code

Formasitchijoh:
have you done something yet

Formasitchijoh:
Becauses like the mentors said more than one person can work on an issue, then they should indicate on the issue to the appropriate mentor. so i think that wont be a problem

Businge931:
Okay great

Businge931:
@Formasitchijoh Am gonna submit my idea also

Businge931:
@Formasitchijoh how did you create a PR?

Formasitchijoh:
> 

No problem @Businge931  we need more ideas onsolving the problem

PR #377, opened 2023-03-14
Fixed: Display Revision list onSelect from the dropDown #372

Formasitchijoh:
This pull request  references issue #372  
 From the previous implementation the user has to focus on the text area before the list of revisions is rendered, which causes poor user interaction  since they have to perform two operations to to get a single result 

Solution: 
Using a displayList state to check if the revision is selected or not the setDisplayList is passed as a prop to the unHandleChangeDropdown and when the searchDropDown is called the displayList is set to true  and they triggers the display of the list hence the user need to select the appropriate revision from the drop down and the list is instantly rendered, and also the user can search specified revision


Formasitchijoh:
@Carla-Moz  @kimberlythegeek  could you have a look at this pull request is your chanced

PR #387, opened 2023-03-16
Confidence tooltip breaks UI

tobidosumu:
### Issue Description  🐛
When there is a result that has unknown confidence, the alignment of the cell is disrupted, and after several of these in the same table, the difference becomes very apparent:
Please, find issue  👉 #258 

### This PR Fixes  💡:

1. the bug causing `confidence` and its `vertical elipsis icon` to collapse and stack on top of  each other when the **vast majority** of the results of the `confidence` table column are `not available`/`unkown` (that is, when it is mostly `question mark`s).
2. the bug causing the `question mark` table cell to break the UI of the table.
3. ensures that the UI remains responsive and consistent across various screen sizes i.e. table UI is mobile responsive. 

- **Fixed Desktop View**
confidenceToolTipBreaksUi-desktop

- **Fixed Mobile View**
confidenceToolTipBreaksUi-mobile

### How I Fixed the Bug  💡:

1.  **causing the 'not available confidence icon (i.e. the question mark icon)' to break the UI**

 I simply refactored the `Confidence` table cell code from this:

      ```
       <TableCell
          data-testid="confidence-icon"
          className={`${setConfidenceClassName(result.confidence_text)}`}
        >
          <Tooltip title="Confidence not available">
            <IconButton className="missing-confidence-button">
              <QuestionMarkIcon className="missing-confidence-icon" />
            </IconButton>
          </Tooltip>
        </TableCell>
      ```
             
   to this:

          ```
         <Tooltip title="Confidence not available">
            <IconButton className="missing-confidence-button" 
              sx={{
                position: 'relative',
                top: '50%',
                transform: 'translateY(-50%)',
              }}
           >
              <QuestionMarkIcon className="missing-confidence-icon" />
            </IconButton>
          </Tooltip>
        </TableCell>
       ```

-  I also removed the borderBottom styling from the tableCells and then added the borderBottom styling to the `CompareResultsTableRow.tsx` component instead. This way, the table cell UI will not break when the `question mark icons` appear instead of the `downward-pointing caret icons` under the `Confidence` table cell column.

  @kimberlythegeek @Carla-Moz

### DIFF
--- a/src/__tests__/CompareResults/__snapshots__/CompareResultsView.test.tsx.snap
+++ b/src/__tests__/CompareResults/__snapshots__/CompareResultsView.test.tsx.snap
@@ -11,13 +11,13 @@ exports[`CompareResults View Should match snapshot 1`] = `
         class="MuiTable-root css-d8sagy-MuiTable-root"
       >
         <thead
-          class="MuiTableHead-root css-15wwp11-MuiTableHead-root"
+          class="MuiTableHead-root css-1y54jnf-MuiTableHead-root"
         >
           <tr
             class="MuiTableRow-root MuiTableRow-head css-1eza0l0-MuiTableRow-root"
           >
             <th
-              class="MuiTableCell-root MuiTableCell-head MuiTableCell-alignCenter MuiTableCell-sizeSmall css-1l87mn7-MuiTableCell-root"
+              class="MuiTableCell-root MuiTableCell-head MuiTableCell-alignRight MuiTableCell-sizeSmall css-19fbjjp-MuiTableCell-root"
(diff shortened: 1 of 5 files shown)


tobidosumu:
Hello @OndigoWendy 

Thank you.
No, there are no potential edge cases where this fix might not work as expected. I tested the fix with all the available data we currently have it worked just fine. This means that the UI did not break when there is only one result with unknown confidence in a table or when there other numbers of unknown confidence in the result.

Appologies for the delayed reply. 🙏

@Carla-Moz @kimberlythegeek

PR #429, opened 2023-03-20
[Fix Accessibility ♿] - Add Missing Tooltip for Close Icon Button in SelectedRevisionsTableRow Component

tobidosumu:
**This is a PR to issue #427.**

### Description
This PR adds a tooltip to the **close** icon button in the `SelectedRevisionsTableRow` component to improve accessibility for users. The tooltip will be displayed when a user hovers over the button, providing additional context and information about the button's purpose.

### Changes Made
Added title attribute to the Button component in the `SelectedRevisionsTableRow.tsx` file.

The title attribute contains the string "Close revision", which will be displayed as the tooltip when the user hovers over the **close** icon button.

### Testing

- Tested the changes locally to ensure that the tooltip appears correctly when hovering over the `close` icon buttons
- Confirmed that all existing tests pass with the addition of the title attribute

### Before

SelectedRevisionsCloseBtnAccessibility-before

### After

SelectedRevisionsCloseBtnAccessibility-after

@Carla-Moz @kimberlythegeek

### DIFF
--- a/src/__tests__/__snapshots__/SelectedRevisionsTable.test.tsx.snap
+++ b/src/__tests__/__snapshots__/SelectedRevisionsTable.test.tsx.snap
@@ -136,7 +136,9 @@ exports[`Search View should match snapshot 1`] = `
                   class="MuiTableCell-root MuiTableCell-body MuiTableCell-sizeMedium css-pmplfi-MuiTableCell-root"
                 >
                   <button
+                    aria-label="Close revision"
                     class="MuiButtonBase-root MuiIconButton-root MuiIconButton-sizeMedium css-1qlpk5e-MuiButtonBase-root-MuiIconButton-root"
+                    data-mui-internal-clone-element="true"
                     id="close-button"
                     tabindex="0"
                     type="button"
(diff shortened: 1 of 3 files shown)


Carla-Moz (commented):
Is there a way to make this accessible without adding a tooltip?

tobidosumu:
Hello @Carla-Moz 

Yes, there are other ways to make this component accessible without adding a tooltip. 
Here are some suggestions:

1. an aria-label attribute can be added to the close icon button with a descriptive label such as "Close Revision". 
This will provide a text alternative for screen readers to announce the purpose of the button.

2. the size of the close icon button can be increased to ensure there is enough contrast between the button color and the background color and to make it more visible.

3. a screen reader can also be used to test the component to make sure it is announced correctly.

4. some additional visual cues, such as changing the color or adding an underline to the selected row, to make it clearer that the close icon button is associated with the selected row can also be added.

I think these methods would help provide a more accessible solution without the need for a tooltip. Additionally, since we already know the demographics of our user base, it is reasonable to expect that they are familiar with the functionality of a close button.

I would like to update the PR with the suggestions mentioned earlier, is that okay?

Issue #435, opened 2023-03-21
Scrollable Search Results List


timmy3000:
Hi @kimberlythegeek,

May i please make the Search Results List Scrollable so that the search input can be more accessable?

Thank you.

![screen-capture](https://user-images.githubusercontent.com/66499629/226742197-a23ef591-4a1e-4c5a-af08-8485d29df61c.gif)

tobidosumu:
@timmy3000 @kimberlythegeek @Carla-Moz 

Hello.
This issue was raised in #421 and a PR to fix it was made in #422.

timmy3000:
@tobidosumu Sorry, I was not aware that your issue was for solving the same problem; Thanks for the heads-up.

tobidosumu:
No problem @timmy3000 

I was only trying to prevent the replication of the same issue.

Please, feel free to make as many PRs to the issues as you deem necessary. I believe this will give the Maintainers the opportunity to merge the best PRs to the issues raised. 

And please be reminded that the overall goal with this project is to create the best product possible for the users.

PR #436, opened 2023-03-21
Make search results list scrollable

timmy3000:
issue: #435 
i made the search results list scrollable.

screen-capture21-03-2023-23-25

### DIFF
--- a/src/components/Search/SearchResultsList.tsx
+++ b/src/components/Search/SearchResultsList.tsx
@@ -1,3 +1,5 @@
+import { useEffect } from 'react';
+
 import Box from '@mui/material/Box';
 import List from '@mui/material/List';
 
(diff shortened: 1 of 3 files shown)


timmy3000:
Issue: #435

src/components/Search/SearchResultsList.tsx
```diff
@@ -7,15 +9,43 @@ import SearchResultsListItem from './SearchResultsListItem';
 function SearchResultsList(props: SearchResultsListProps) {
   const { searchResults, view } = props;
 
+  useEffect(() => {
```
Carla-Moz, line 12: Is this useEffect really necessary? How would you accomplish the same result without it?

timmy3000: I used `useEffect` hook to get the `search-results-list` element once it is mounted and then calculate the amount of height it can occupy. Should I use a different approach?

Carla-Moz: It's not whether _you should_ use a different approach, but thinking through whether it's necessary or not to accomplish your goal.

timmy3000: I think the `useEffect` hook is necessary since it will help me get the element once it is mounted, is there a reason why you think I shouldn't use it?

PR #451, opened 2023-03-31
Dark mode

Carla-Moz:
This PR adds dark mode functionality to the header and also persists the dark mode selection on page refresh or entering url. 

**Note:** I moved the header to the `App` file because it's a constant on every page; only one page requires an update on the header which can easily be changed. I also updated the tests to reflect the new dark mode functionality. Please let me know what you think about the test and header placement updates. 

I left the 'Learn more' button for style purposes. We also need to figure out where to place the feedback alert button.

Screenshot 2023-03-31 at 12 58 29 PM


src/components/CompareResults/CompareResultsTable.tsx
```diff
@@ -26,25 +26,25 @@ function CompareResultsTable(props: CompareResultsProps) {
...
+          <Grid item xs={1} className='compare-results-error'>
             <CircularProgress />
           </Grid>
         </Box>
       )}
       {compareResults.error && (
-        <Box display="flex" justifyContent="center">
-          <Grid item xs={6} className="compare-results-error">
-            <Alert severity="error" sx={{ textAlign: 'center' }}>
+        <Box display='flex' justifyContent='center'>
```
Carla-Moz, line 45: I don't know why it's complaining about coverage here so I'd appreciate some eyes on this. Thanks!

src/components/App.tsx
```diff
@@ -35,24 +34,25 @@ function App() {
           </Link>
           .
         </Alert>
+        {/* What would we like to do with this feedback box ?*/}
```
beatrice-acasandrei, line 37: @kimberlythegeek @esanuandra If we still find it useful then perhaps we can incorporate the feedback message in the warning that we already display on the first page - `This is an unstable pre-release version. Some features may not yet be supported. Please file any bugs on the [Github Repo](https://github.com/mozilla/perfcompare/issues).` - I think it's better then having it hidden.

src/components/CompareResults/PaginatedCompareResults.tsx
```diff
@@ -205,7 +205,7 @@ function PaginatedCompareResults(props: PaginatedCompareResultsProps) {
 }
 
 interface PaginatedCompareResultsProps {
-  mode: 'light' | 'dark';
+  theme: string;
```
beatrice-acasandrei, line 208: If we already know that the only two possible themes are `light` or `dark` then I prefer `theme: 'light' | 'dark'`, but that's just a personal preference.

src/components/App.tsx
```diff
@@ -35,24 +34,25 @@ function App() {
...
         <Box display='flex' justifyContent='flex-end' alignItems='flex-end'>
           <FeedbackAlert />
-          <ToggleDarkMode
-            toggleColorMode={toggleColorMode}
-            theme={protocolTheme}
-          />
         </Box>
+        <PerfCompareHeader
```
beatrice-acasandrei, line 41: From your note "I moved the header to the App file because it's a constant on every page; only one page requires an update on the header which can easily be changed.". -> Only the PerfCompare logo is constant on Search View and Results View. Did you take this into consideration when moving `PerfCompareHeader`?

src/components/App.tsx
Carla-Moz: Moved the leave feedback icon to top banner

src/components/App.tsx
Carla-Moz: Perfcompare header imported only in Search and Results view; removed import from App

src/components/CompareResults/PaginatedCompareResults.tsx
Carla-Moz: Valid point.  I changed all to 'light' | 'dark'

beatrice-acasandrei (approved):

esanuandra (approved):

PR #457, opened 2023-04-10
PCF-226 Results: Display table header (platform, status and confidence will be disabled)

esanuandra:
image

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -68,7 +68,7 @@ function App() {
             />
             <Route
               path='/beta/compare-results'
-              element={<ResultsView />}
+              element={<ResultsView protocolTheme={protocolTheme} />}
             />
           </Routes>
         </Router>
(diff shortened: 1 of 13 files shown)


beatrice-acasandrei (approved):

src/components/App.tsx
```diff
@@ -67,7 +68,7 @@ function App() {
             />
             <Route
               path='/beta/compare-results'
-              element={<CompareResultsView theme={protocolTheme} />}
+              element={<ResultsView />}
```
Carla-Moz, line 71: How will dark mode be handled in the Results view?

Carla-Moz: Or would you like me to make a ticket to do so when the results view is done?

esanuandra: Fixed, added the dark mode now.

Carla-Moz (approved):

PR #470, opened 2023-05-12
Add some documentation about deployment

julienw:
I wrote some documentation about how to deploy to production. I mostly copied/pasted from the profiler doc. I also copied the "revert to previous version" script from there, and of course adjusted it for this project too.

Tell me what you think!

### DIFF
--- a/Deployment.md
+++ b/Deployment.md
@@ -1,0 +1,40 @@
+# Deployments
+
+PerfCompare is hosted on Netlify, and is updated every time commits are pushed to the following branches:
+
+| Branch Name | URL                                               | Description                                                                                                   |
+| ----------- | ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------- |
+| production  | https://perf.compare/                             | Production branch, updated every 1-2 weeks.                                                                   |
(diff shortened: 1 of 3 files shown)


Deployment.md
```diff
@@ -0,0 +1,40 @@
...
+
+## How to deploy beta to production
+
+The easiest by far is to
+[create a pull request on GitHub](https://github.com/mozilla/perfcompare/compare/production...beta?expand=1).
+It would be nice to write down the main changes in the PR description.
+
+After the PR is created all checks should run. When it's ready the PR can be
+merged. Be careful to always use the **create a merge commit** functionality,
+not *squash* or *rebase*, to keep a better history.
```
Carla-Moz, line 21: The team prefers to squash the commits, however.

julienw: That's totally fine when merging to the main branch, but not to production. 

If you squash all commits to 1 production commit:
- you can't easily know if a specific commit from the main branch is in production.
- you can't check it production has extra commits (which would be bad).
- I'm not sure how the next PR would look like: indeed the commits on production would be different than the commits on the source branch. I don't know if git/GitHub can deal with this. (Maybe?)

Hope this clears up this topic!

Carla-Moz (approved):
Thanks for clearing that up!

PR #475, opened 2023-05-17
Update tests for the new SearchView component

sumairq:
## Hi ✋ 
# In this pull request

### The following tests were updated for the Search View Beta component.

- Searchview renders correctly when there are no results.
- Search container renders "compare with base".
- Base Search renders repository dropdown in closed condition.
- Base Search should hide search results when clicking outside of search input.
- Base Search should hide the search results when Escape key is pressed.
- Base Search should not call fetch if search value is not a hash or email.
- Base search should clear search results if the search value is cleared.
- Base search should not hide search results when clicking search results.
- Base search should update error state with generic message if fetch error is undefined.
- Base search should disable adding more than one revision for the base search.

**Note:** Similar tests were previously used for the search view component, they have been updated and configured to test the new `SearchViewBeta` component.
For now, only the Base search has been tested. Later the tests for Revision Search will also be added. 

### Refactors

- Search view component render operation has been abstracted away with a reusable function `renderComponent()`.
- Test data fetch operation has been abstracted away with a reusable function `fetchTestData()` which takes `data` as its argument.

image

### DIFF
--- a/src/__tests__/Search/SearchViewBeta.test.tsx
+++ b/src/__tests__/Search/SearchViewBeta.test.tsx
@@ -1,26 +1,276 @@
 import { renderHook } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { act } from 'react-dom/test-utils';
 
 import SearchViewBeta from '../../components/Search/beta/SearchView';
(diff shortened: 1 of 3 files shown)


Carla-Moz (approved):
LGTM!

PR #482, opened 2023-05-21
Clean revision input

Carla-Moz:
Note: An updated PR for the revision input. 

This PR fulfills https://mozilla-hub.atlassian.net/browse/PCF-233

- [x]     Revisions and an info icon (with tooltip text “Revisions (typically including your changes) to compare against the selected base revision.”):
- [x]         repository dropdown and 
- [x]         a search for the revision, three revisions max (not in the design)
- [x]  Updated tests

Screenshot 2023-05-17 at 17 40 25


src/hooks/useHandleChangeSearch.ts
```diff
@@ -18,41 +18,72 @@ let timeout: null | ReturnType<typeof setTimeout> = null;
 
 const useHandleChangeSearch = () => {
   const dispatch = useAppDispatch();
-  const getRepository = useAppSelector((state) => state.search.repository);
+  const baseRepository = useAppSelector((state) => state.search.baseRepository);
+  const newRepository = useAppSelector((state) => state.search.newRepository);
 
-  const searchByRevisionOrEmail = async (
-    repository: Repository['name'],
+  const warningText =
```
beatrice-acasandrei, line 24: Perhaps it could be moved to `src\resources\Strings.tsx`?

src/resources/Strings.tsx
```diff
@@ -33,11 +33,19 @@ export const Strings = {
         img: 'https://user-images.githubusercontent.com/88336547/233237125-1534220b-c343-421a-9133-ce8f151cb979.png',
         imgDark:
           'https://user-images.githubusercontent.com/88336547/233250674-004d071a-7c23-40f4-b348-0687a3fef6e3.png',
-        collapedBase: {
-          selectLabel: 'Base',
-          inputPlaceholder: 'Search base by ID number or author email',
-          tooltip:
-            'The baseline revision (no changes) your revision will be compared against.',
+        collaped: {
```
beatrice-acasandrei, line 36:
```suggestion
        collapsed: {
```

src/hooks/useHandleChangeSearch.ts
Carla-Moz: Yes, true!

src/components/Search/SearchDropdown.tsx
julienw, line 1: I believe you should just get rid of the old files, this makes things more complicated than needed for an app that's not used by anybody yet.

src/components/Search/beta/SearchDropdown.tsx
```diff
@@ -20,12 +20,46 @@ import {
...
+
+  const handleRepoSelect = (e: React.MouseEvent<HTMLLIElement>) => {
+    const { classList, id } = e.currentTarget;
+    let name = '';
+    classList.forEach((item) => {
+      if (item.includes('Repository')) {
+        name = item;
+        return;
+      }
+    });
```
julienw, line 55: This looks very brittle. Why don't you use `base` directly, to decide which repository to use?

Also it's not clear to me why we use `onClick` on the menu items instead of the property `onChange` on `<Select>`? Then the selected value can be retrieved in `event.target.value` (see https://mui.com/material-ui/api/select/).

src/components/Search/beta/SearchDropdown.tsx
```diff
@@ -20,12 +20,46 @@ import {
...
+  //searchType is to distinguish between base and new search dropdowns for handleChangeDropdown hook
+  const [repoSelect, setRepoSelect] = useState({
+    baseRepository: 'try',
+    newRepository: 'try',
+    searchType: base,
+  });
+
+  useEffect(() => {
+    void handleChangeDropdown(repoSelect);
+  }, [repoSelect]);
```
julienw, line 45: It's not clear to me why you don't call it directly in `handleRepoSelect` below.

src/components/Search/beta/SearchDropdown.tsx
```diff
@@ -20,12 +20,46 @@ import {
...
+  base,
+}: SearchDropdownProps) {
   const size = view == 'compare-results' ? 'small' : undefined;
   const { handleChangeDropdown } = useHandleChangeDropdown();
 
+  //searchType is to distinguish between base and new search dropdowns for handleChangeDropdown hook
+  const [repoSelect, setRepoSelect] = useState({
+    baseRepository: 'try',
+    newRepository: 'try',
+    searchType: base,
```
julienw, line 40: The `searchType` never changes, so it's not clear to me why you put it in the state. It seems to me this is a prop for the dropdown but will never change for this specific dropdown. Am I missing something?

Also I don't see why we keep both values `baseRepository` and `newRepository`, it looks like to me that we'll ever change one of these values.

Then `handleChangeDropdown` would take 2 parameters `searchType` and `repositoryName`.

src/components/Search/beta/SearchInput.tsx
```diff
@@ -1,24 +1,46 @@
...
   const size = view == 'compare-results' ? 'small' : undefined;
 
+  useEffect(() => {
+    handleChangeSearch(searchState);
+  }, [searchState]);
+
+  const updateSearchState = (
+    e: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+  ) => {
+    const { name, value } = e.currentTarget;
```
julienw, line 37: same comment as above, it's not clear to me why we don't use the prop `base` directly.

src/components/Search/beta/SearchInput.tsx
```diff
@@ -1,24 +1,46 @@
...
+  base,
+  inputError,
+  inputHelperText,
+}: SearchInputProps) {
   const { handleChangeSearch } = useHandleChangeSearch();
+  //searchType is to distinguish between base and new search inputs for handleChangeSearch hook
+  const [searchState, setState] = useState({
+    baseSearch: '',
+    newSearch: '',
+    searchType: base,
```
julienw, line 26: Same comment as above: it's not clear to me why we keep `searchType` that never changes for this specific `SearchInput`. Also it's not clear to me why we keep both base/newSearch values here, where only one of them will ever be changed for this specific input.

src/components/Search/beta/SearchInput.tsx
```diff
@@ -1,24 +1,46 @@
...
+  });
   const size = view == 'compare-results' ? 'small' : undefined;
 
+  useEffect(() => {
+    handleChangeSearch(searchState);
```
julienw, line 31: It's not clear to me why you don't call it directly in `updateSearchState` below.

julienw (commented):
Thanks for this patch, it seems to work fine! Thanks also for all the new tests!

However I think the architecture for this new code could be improved. I believe the patch is a lot more complex than it could be as a result. This is a problem because I believe this will make future changes more difficult. That's why I strongly suggest to change it now, despite that I know the clock is ticking.

1. I've put this as a comment but I'll put it here again: I think we should get rid of the "non beta" parts. There's no point keeping some code that's not used by anybody. This makes things more complex than needed, because you need to keep some state and other code because of that.

2. Think of each `SearchComponent` as independent to each other. Currently both instances keep state for both cases. Instead, pass to one instance everything it needs to manage its own state, but only its own state. Each `SearchComponent` could be connected to redux too, so that it could take just `searchType` as a own prop, and take all other values from the state. (see also point 3 below)

3. The state itself could be improved as well, by using `new` and `base` as keys, so that the state, reducers and selectors would be much simpler to manipulate.

I hope this won't be too much change for you, but I do think that changing this now will make it easier (and thus faster) later.

Happy to discuss! Please tell me if I missed anything!

src/types/state.ts
```diff
@@ -76,10 +76,14 @@ export type CompareResultsItem = {
...
+  baseRepository: Repository['name'];
+  newRepository: Repository['name'];
   searchResults: Revision[];
+  baseSearchResults: Revision[];
+  newSearchResults: Revision[];
   searchValue: string;
-  inputError: boolean;
+  inputErrorBase: boolean;
+  inputErrorNew: boolean;
   inputHelperText: string;
```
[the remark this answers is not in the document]
Carla-Moz: This is so much cleaner. You're right. It would have gotten out of hand with a third input and it will reduce the complexity.

src/reducers/SearchSlice.ts
```diff
@@ -27,18 +36,52 @@ const search = createSlice({
     updateSearchValue(state, action: PayloadAction<string>) {
       state.searchValue = action.payload;
     },
-    updateSearchResults(state, action: PayloadAction<Revision[]>) {
-      state.searchResults = action.payload;
+
+    updateSearchResults(
+      state,
+      action: PayloadAction<{
+        payload: Revision[];
```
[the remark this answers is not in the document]
Carla-Moz: Yeah, that works for me.

src/reducers/SearchSlice.ts
```diff
@@ -27,18 +36,52 @@ const search = createSlice({
...
+        searchType: 'base' | 'new';
```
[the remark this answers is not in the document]
Carla-Moz: Yes, with the new search state architecture, this will help simplify the other reducers and remove all the tenerary statements.

src/reducers/SearchSlice.ts
```diff
@@ -27,18 +36,52 @@ const search = createSlice({
...
+
+    updateBaseRepository(state, action: PayloadAction<Repository['name']>) {
+      state.baseRepository = action.payload;
     },
-    updateRepository(state, action: PayloadAction<Repository['name']>) {
-      state.repository = action.payload;
+
+    updateNewRepository(state, action: PayloadAction<Repository['name']>) {
+      state.newRepository = action.payload;
     },
```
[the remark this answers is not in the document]
Carla-Moz: Agreed! This SearchSlice is looking so much cleaner. Thank you!

src/types/state.ts
```diff
@@ -76,10 +76,14 @@ export type CompareResultsItem = {
 };
 
 export type SearchState = {
-  repository: Repository['name'];
+  baseRepository: Repository['name'];
+  newRepository: Repository['name'];
   searchResults: Revision[];
+  baseSearchResults: Revision[];
+  newSearchResults: Revision[];
   searchValue: string;
```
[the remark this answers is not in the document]
Carla-Moz: Yes, true; this gets resolves with the new Search type.

src/thunks/searchThunk.ts
```diff
@@ -1,74 +1,99 @@
...
+interface FetchDataArgsIDEmail {
+  repository: Repository['name'];
+  search: string;
+  searchType: InputType;
+}
+
+interface FetchDataMeta {
+  error: string;
+  searchType: InputType;
+}
```
julienw, line 22: I think you don't need to reject with the `searchType`, because in the reducer you use `meta.arg` directly. So you could reject with the error directly, like before.

src/thunks/searchThunk.ts
```diff
@@ -1,74 +1,99 @@
...
+  'search/fetchRecentRevisions',
+  async ({ repository, searchType }, { rejectWithValue }) => {
+    let response;
+
+    try {
+      response = await fetch(
+        `${treeherderBaseURL}/api/project/${repository}/push/?hide_reviewbot_pushes=true`,
+      );
+    } catch (err) {
+      const error = (err as Error).message;
```
julienw, line 38: (not for this patch, could be better as a follow-up, because it's also common to other places) I think it's more useful to have the full error object instead of the message, because we're losing the stack information. Possibly it could be good to output the full error message in the console with something such as: `console.error("An error happened while fetching recent revisions", err)`. This would make debugging a lot easier, especially for errors that are not expected.

src/thunks/searchThunk.ts
```diff
@@ -1,74 +1,99 @@
...
+      return rejectWithValue({ error, searchType });
```
julienw, line 39: See above for the comment that you don't need the searchType in the rejection value. (and this applies below too of course)

src/thunks/searchThunk.ts
```diff
@@ -1,74 +1,99 @@
...
     }
     const json = (await response.json()) as APIPushResponse;
     if (json.results.length > 0) {
       return json.results;
     }
 
-    return rejectWithValue('No results found');
+    return rejectWithValue({ error: 'No results found', searchType });
```
julienw, line 96: In the same logic as above, it would be better to reject with a full Error object (`new Error("No results found")`) so that there's a stack. This could be handled in a follow-up.

julienw (commented):
This looks pretty good to me!

I left some more simple comments, and of course I'd like some green tests like we discussed :)

Also there might be a race: at load time, when I click on the first input, I properly see the dropdown. However, when I click on the second one, the dropdown doesn't show. This works fine as soon as I start changing some values though.
It could be better to handle this issue in a follow-up, so you don't need to handle it in this PR. Just please file an issue if you think it's unexpected.

Carla-Moz:
> This looks pretty good to me!
> 
> I left some more simple comments, and of course I'd like some green tests like we discussed :)
> 
> Also there might be a race: at load time, when I click on the first input, I properly see the dropdown. However, when I click on the second one, the dropdown doesn't show. This works fine as soon as I start changing some values though. It could be better to handle this issue in a follow-up, so you don't need to handle it in this PR. Just please file an issue if you think it's unexpected.

It's actually due to a typo. I was fetching results using "base" searchType for both components. Fixed it and works fine now.

src/components/Shared/SearchComponent.tsx
```diff
@@ -0,0 +1,173 @@
...
+import { RootState } from '../../common/store';
+import { useAppDispatch } from '../../hooks/app';
+import { useAppSelector } from '../../hooks/app';
+import useSelectRevision from '../../hooks/useSelectRevision';
+import { clearCheckedRevisions } from '../../reducers/CheckedRevisions';
+import { SearchStyles } from '../../styles';
+import type { RevisionsList, InputType } from '../../types/state';
+import SearchDropdown from '../Search/SearchDropdown';
+import SearchInput from '../Search/SearchInput';
+import SearchResultsList from '../Search/SearchResultsList';
```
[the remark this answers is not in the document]
Carla-Moz: It's in the Shared folder because the Results view will eventually use this Search Component too, so it makes sense to leave it in Shared?

src/thunks/searchThunk.ts
Carla-Moz: Yes, will handle in follow up PR.

src/thunks/searchThunk.ts
Carla-Moz: Ah, gotcha.

src/reducers/SearchSlice.ts
```diff
@@ -5,73 +5,121 @@ import {
...
+    inputError: false,
+    inputHelperText: '',
+  },
+  new: {
+    repository: 'try',
+    searchResults: [],
+    searchValue: '',
+    inputError: false,
+    inputHelperText: '',
+  },
```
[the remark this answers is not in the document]
Carla-Moz: I hear you. Will update in next PR.

src/components/Shared/SearchComponent.tsx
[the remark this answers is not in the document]
julienw: I'd say that either you put everything in `/Shared/Search`, or everything in `/Search`.

The rationale is that we try to have imports between directories work in one direction only, to avoid "spaghetti". Shared libraries are usually at the bottom of the import tree, they don't import from other directories (except maybe a `utils` directory that usually contains low-level shared code) exactly because other directories import from `shared`.

Also, I tend to think of a `shared` library as a bag of things that are independent with each other, and don't warrant their own directory in a sense. I know developers that don't want a `shared` directory at all and prefer to group their shared components by bigger topics.

In the case of the `Search` component, I think it definitely warrants its own directory, with many other dependency files. Being in its own directory `/Search` doesn't prevent it from being imported by different other components!

I strongly believe that Search and its dependencies should be together, but in the end I'll let you decide if you want everything in `components/Shared` or in `components/`.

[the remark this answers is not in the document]
Carla-Moz: Well, yeah, I know putting it in `Search` won't stop it from being imported, of course, lol. However, the rationale makes sense so I don't mind moving it out of 'Shared' and keeping all Search components in one directory and so on.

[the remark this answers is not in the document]
Carla-Moz: Will follow up in next PR.

src/components/Shared/SearchComponent.tsx
```diff
@@ -0,0 +1,173 @@
...
+    if (e.key === 'Escape') {
+      setFocused(false);
+    }
+  };
+
+  useEffect(() => {
+    if (focused == false) {
+      dispatch(clearCheckedRevisions());
+    }
+  }, [focused]);
```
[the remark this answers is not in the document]
Carla-Moz: Noted. I'll update in next PR.

src/components/Shared/SearchComponent.tsx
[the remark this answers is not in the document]
julienw: I see that SearchComponent is loaded in CompareBase, so after all it might make sense to move all Search components to a Shared subdirectory. Looking at this in a separate PR sounds good to me, as this may need more changes than just moving things around.

src/__tests__/SelectedRevisionsTable.test.tsx
```diff
@@ -1,286 +0,0 @@
-import { renderHook } from '@testing-library/react';
```
julienw, line 1: I see a lot of files like this one are deleted, maybe you meant to just skip the tests instead of delete them? In that case you can use `describe.skip` instead of just `describe`, and keep the file in this patch.

Of course we'll want to unskip it later, but I'm fine if you want to work on it in a dedicated follow-up PR.

(Same comment for other deleted test files)

Carla-Moz: Honestly, these tests and reducers need be to redone. We will have to create new tests when these beta versions of these components are made. And like you said before, the history is there on github.

Carla-Moz: My next PR will focus on SelectedRevisions so all these tests will pop back again.

PR #490, opened 2023-06-16
comparison warning

Carla-Moz:
This PR fulfills issue https://mozilla-hub.atlassian.net/browse/PCF-241:

_Add warning when comparing try and mozilla-central: Comparing “try” repository and “mozilla-central” repository is not recommended._

Screenshot 2023-06-16 at 14 12 38


src/components/Search/CompareWithBase.tsx
```diff
@@ -23,18 +26,31 @@ interface Expanded {
...
 
   useEffect(() => {
     setFormHeight(formWrapperRef.current?.clientHeight || formHeight);
   }, [formHeight]);
 
+  useEffect(() => {
+    if (
+      (baseRepository === 'try' && newRepository === 'mozilla-central') ||
+      (baseRepository === 'mozilla-central' && newRepository === 'try')
+    ) {
```
julienw, line 49: It would be good to check with mozilla-beta and mozilla-release too. I've been told that the main issue that makes it difficult to compare is that there are different levels of optimization between the mozilla* branches and try (by default), but then there's no reason to single out mozilla-central.

Probably it would be good to explicit the no-warning pairs:
try, try
mozilla-XXX, mozilla-XXX

The check could be extractd to a function to make it easier to read here.

What do you think?

julienw (commented):
Thanks! This looks mostly good to me.
I just wonder why mozilla-central is singled out.
Also it could be good to have a test for this, if that's not too much to ask.

(for another patch) I see that the close button has a weird color:

image

src/components/Search/CompareWithBase.tsx
gmierz: I left a comment with some more context about this issue in the jira ticket: https://mozilla-hub.atlassian.net/browse/PCF-241

To summarize, I think we could simply display a warning anytime a try push is not being compared to another try push. For the other cases like m-c vs autoland, the differences in configurations aren't as much of an issue as try vs others.

esanuandra (approved):

beatrice-acasandrei (approved):

Issue #486, opened 2023-06-07
Make `Compare with Base` more accessible and write tests for the improvements


sumairq:
Make `Compare with base` in the `Search view` accessible through keyboard while making the interface more intuitive for screen reader users.

The regimen for improvements to accessibility in this case could be:

- Add a hidden "Skip to content" link.
- Work on improving the tab index and check improvements with manual testing.
- Enable announcement of live regions to screen reader users.
- Work on visible focus styles.
- Make sure that if we are opening a layer on screen we are actually trapping the focus within the top layer.
- Write tests for accessibility.

julienw:
Moved to bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1855187

Change performed by the [Move to Bugzilla add-on](https://addons.mozilla.org/en-US/firefox/addon/move-to-bugzilla/).

PR #491, opened 2023-06-16
Implement aria-live attribute for notistack alerts

sumairq:
## This PR adds `aria-live` attribute to notistack alerts. 
Is part of solution for #486 issue.
### Context
It is very important for a user with visual impairment to know how the interface is reacting to their requests in real-time. It requires us to make sure that alerts are announced to someone using assistive technologies such as a screen reader. For this reason, we use the `aria-live` attribute. 
Aria-live attribute exposes the alert message in the accessibility tree so the screen reader can announce the message in real-time as the alert goes off.

### How it was implemented

Faced with the issue that custom attributes cannot be added to notistack alert itself. The `domRoot` attribute made available by notistack  in `<snackbarProvider>` is passed with an alert container that contains the alert and holds the `aria-live` attribute.
A component is used for this container. This component forwards its `div` element via `forwardRef`. Finally, the element is managed with the `useState()` hook.

- Added `'react/prop-types' : [ 'off' ]` to eslint configuration as the type checking for props is present by default in TypeScript.

- Added test to check for `aria-live` attribute in the notistack alerts in `Snackbar.test.tsx`.

image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -63,6 +63,7 @@ module.exports = {
       rules: {
         // https://github.com/typescript-eslint/typescript-eslint/issues/1824
         '@typescript-eslint/indent': 'off',
+        'react/prop-types': ['off']
       },
     },
     {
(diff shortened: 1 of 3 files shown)


Carla-Moz (approved):
This looks great! Thank you!

PR #494, opened 2023-06-27
Update version of typescript target and slightly update babel config

julienw:
tsconfig was targeting an old version of javascript, which could trigger typescript error when using newer features of javascript.

I checked on https://kangax.github.io/compat-table/es2016plus/, and it looks like all browsers support es2022 just fine, except one minor thing in RegExp that we don't use.

I also removed an outdated configuration in babel.

I checked that all scripts were still working correctly and that the project would build. Please double check and with your IDE too, to see if everything looks right on your end too!

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,13 +1,5 @@
 {
   "targets": ["last 2 chrome versions", "last 2 firefox versions"],
-  "plugins": [
-    [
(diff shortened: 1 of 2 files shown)


beatrice-acasandrei (approved):
Looks good on my end!

esanuandra (approved):
It works for me too

PR #496, opened 2023-06-27
Implement "Skip to search" link for Search View

sumairq:
## This PR implements a Skip to search link. 
Is part of the solution for #486 issue.
### Why a Skip Link
Skip links allow keyboard-only users, sighted or not, to bypass large or repetitive blocks of content, so they don't have to endlessly tab through the page to the section they want to get to.

It's important for the users to be familiar with all the information on the page, however, on subsequent visits, it becomes pointless to make a keyboard user go through the same information every time there is a page refresh for example.

A skip link takes a huge weight off of the users of assistive technology and they can access the features or information most relevant to them in a much faster way without having to endlessly tab through the same information over and over.

In PerfCompare's case, we would be saving the user a total of at least 7-8 tabs through irrelevant content when using this skip Link.
### Its Use
When the page is loaded the `Skip to search` link is hidden and becomes visible on focus (pressing tab) after which the user presses the enter key which takes the focus directly to the Search Container.

skiplink

### How it was implemented

- Implemented a reusable SkipLink component that passes the `onClick` Handler and `containerRef` to its child.
- Used a containerRef that connects the Seach container with the `SkipLink`.

- Added Styles in `src\styles\AccessibilityComponents.ts` and used `#42414d` as background to achieve AAA WCAG contrast ratio. 

- Added tests in `src\__tests__\Search\SearchView.test.tsx`


src/styles/AccessibilityComponents.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const skipLink = style({
+    left: '0%',
+    top: '0%',
+    transform: 'translateY(-110%)',
+    borderRadius: '5px',
+    position: 'absolute',
+    border: 'none',
+    color: 'white',
+    background: '#424242',
```
Carla-Moz, line 11: Could you please use the Color variables here from `Colors.ts`

Carla-Moz: Lines 10-11, use the color variables for consistency

Carla-Moz (commented):
Just had a few comments; otherwise, lgtm!

sumairq:
Thank you for the comments 😄

Carla-Moz (approved):
Thanks!

PR #497, opened 2023-06-30
Framework compare

Carla-Moz:
This PR finishes 1 epic, 1 subtask, and 1 bug:

1. [Search: Add component for Compare with a base]( https://mozilla-hub.atlassian.net/browse/PCF-212)
2. [Search: Framework Dropdown and Compare Button section](https://mozilla-hub.atlassian.net/browse/PCF-234)
3. [Search: Get correct repository in selected / checked revisions](https://mozilla-hub.atlassian.net/browse/PCF-264)

The user can now select the testing framework, click on the Compare Button, which leads to the results page with the selected revisions which will remain on page on refresh. 

Screenshot 2023-07-12 at 13 34 32

Note: WIP: Need to update tests and coverage


src/common/constants.ts
```diff
@@ -13,14 +13,17 @@ export const userFeedbackMessage =
...
   1: 'mozilla-central',
   2: 'mozilla-beta',
   3: 'mozilla-release',
   4: 'try',
+  6: 'mozilla-beta',
+  7: 'mozilla-release',
   30: 'fenix',
   77: 'autoland',
+  108: 'fenix',
 };
```
beatrice-acasandrei, line 27:
```suggestion
export const repoMap: Dictionary<Repository['name']> = {
  1: 'mozilla-central',
  4: 'try',
  6: 'mozilla-beta',
  7: 'mozilla-release',
  77: 'autoland',
  108: 'fenix',
};
```
We should remove the duplicated repositories.

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -1,36 +1,82 @@
...
+import { style } from 'typestyle';
 
+import useFetchCompareResults from '../../../hooks/useFetchCompareResults';
+import useHandleChangeSearch from '../../../hooks/useHandleChangeSearch';
+import { SearchContainerStyles } from '../../../styles';
+import { background } from '../../../styles';
+import { Repository } from '../../../types/state';
+import CompareWithBase from '../../Search/CompareWithBase';
 import PerfCompareHeader from '../../Shared/PerfCompareHeader';
-import InputsReplacement from './InputsReplacement';
```
beatrice-acasandrei, line 6: Could you also remove this component since we no longer use it?

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -1,36 +1,82 @@
...
   return (
-    <Container maxWidth='xl' data-testid='beta-version-compare-results'>
+    <div
+      className={styles.container}
+      data-testid='beta-version-compare-results'
+    >
       <PerfCompareHeader
         themeMode={themeMode}
         toggleColorMode={toggleColorMode}
+        view='compare-results'
```
beatrice-acasandrei, line 68: The constant can be used here.

src/components/Search/SearchDropdown.tsx
```diff
@@ -97,6 +97,15 @@ function SearchDropdown({
...
   };
 
+  const repoKeys = Object.keys(repoMap);
+  const filteredRepoList = repoKeys
+    .map((key) => repoMap[key])
+    .filter((repo, index, self) => {
+      return self.indexOf(repo) === index;
+    });
+
+  // debugger;
```
beatrice-acasandrei, line 107:
```suggestion
```

src/types/state.ts
```diff
@@ -10,8 +10,11 @@ export type Repository =
   | { id: 2; name: 'mozilla-beta' }
   | { id: 3; name: 'mozilla-release' }
```
beatrice-acasandrei, line 11:
```suggestion
```

src/types/state.ts
```diff
@@ -10,8 +10,11 @@ export type Repository =
...
   | { id: 4; name: 'try' }
+  | { id: 6; name: 'mozilla-beta' }
+  | { id: 7; name: 'mozilla-release' }
   | { id: 30; name: 'fenix' }
```
beatrice-acasandrei, line 15:
```suggestion

```

src/components/Search/SearchComponent.tsx
```diff
@@ -137,12 +143,19 @@ function SearchComponent({
...
+      {(checkedRevisionsList.length > 0 ||
+        (selectedRevisions && selectedRevisions.length > 1)) && (
         <Grid>
           <SelectedRevisions
             searchType={searchType}
             mode={mode}
             isWarning={isWarning}
+            selectedRevisions={
+              searchType == 'base'
+                ? selectedRevisions.filter((_, index) => index === 0)
```
julienw, line 156: What do you think of:
```suggestion
                ? [selectedRevisions[0]]
```
This is O(0) while your filter operation is O(n).

It's probably not a big deal here because n is low, but still it's also simpler.

src/components/Search/SearchDropdown.tsx
```diff
@@ -121,14 +130,14 @@ function SearchDropdown({
           variant='standard'
           onChange={(e) => void handleRepoSelect(e)}
         >
-          {Object.keys(repoMap).map((key) => (
+          {filteredRepoList.map((key) => (
```
julienw, line 133: They aren't the keys anymore but the values. Maybe you can use `repo` as variable name?

src/components/Shared/PerfCompareHeader.tsx
```diff
@@ -66,23 +39,23 @@ function PerfCompareHeader(props: PerfCompareHeaderProps) {
...
+            align='center'
+            gutterBottom
+            className='perfcompare-tagline'
+          >
+            {strings.tagline}
+          </Typography>
+        )}
+        {view == 'search' && (
+          <Button className='learn-more-btn'>{strings.learnMore}</Button>
+        )}
```
julienw, line 54: What do you think of gathering the 2 items in the same block?
```suggestion
        {view === 'search' && (
          <>
            <Typography
              component='div'
              align='center'
              gutterBottom
              className='perfcompare-tagline'
            >
              {strings.tagline}
            </Typography>
            <Button className='learn-more-btn'>{strings.learnMore}</Button>
          </>
        )}
```
I think this makes the flow easier to read.

julienw (commented):
Thanks, it's great to finally see things getting together!
I didn't do a thorough review because I believe Beatrice already did it, but left a few comments still.

I still looked at the deploy preview, and there are some problems in the compare results page:
* the form is still displayed (but maybe that's expected?)
* the base is also displayed in the Revisions part.
 
But I think we can look at these problems in follow-ups.


Tell me what you think :-)

src/common/constants.ts
Carla-Moz: Yes, the duplicates need to be removed. I don't know why I left them. I think when updating this list I thought the keys would remain the same instead of changed. I've requeried the db and got the unique keys and values. I've fixed this; this will making mapping through easier later down.

Carla-Moz:
> Thanks, it's great to finally see things getting together! I didn't do a thorough review because I believe Beatrice already did it, but left a few comments still.
> 
> I still looked at the deploy preview, and there are some problems in the compare results page:
> 
>     * the form is still displayed (but maybe that's expected?)
> 
>     * the base is also displayed in the Revisions part.
> 
> 
> But I think we can look at these problems in follow-ups.
> 
> Tell me what you think :-)

Yes, the results view for this component will be handled in the next epic: [Display the inputs selected for Compare with a base](https://mozilla-hub.atlassian.net/browse/PCF-215 ). That's why you see the form still displayed etc

src/components/Search/SearchDropdown.tsx
```diff
@@ -97,6 +97,15 @@ function SearchDropdown({
     }),
   };
 
+  const repoKeys = Object.keys(repoMap);
+  const filteredRepoList = repoKeys
+    .map((key) => repoMap[key])
+    .filter((repo, index, self) => {
+      return self.indexOf(repo) === index;
+    });
```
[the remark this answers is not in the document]
Carla-Moz: Yes, it remove the duplicates, but it's no longer needed since all the values are unique now.

src/components/Search/SearchDropdown.tsx
Carla-Moz: Yes, I forgot to update when changeing the mapped list. Again, not necessary since there are no duplicates.

src/components/Search/SearchComponent.tsx
Carla-Moz: Oh I did that first but typescript was unhappy about it so I had to filter the list.  Yes, stupid. I know.

src/components/Shared/PerfCompareHeader.tsx
Carla-Moz: Agreed. I rushed through this one so thanks for catching it.

src/hooks/useSelectedRevisions.ts
```diff
@@ -0,0 +1,32 @@
...
+  const selectedRevisions = useAppSelector(
+    (state) => state.selectedRevisions.revisions,
+  );
+
+  const addSelectedRevisions = () => {
+    const newSelected = [...selectedRevisions];
+    newSelected.push(...baseCheckedRevisions, ...newCheckedRevisions);
+
+    dispatch(setSelectedRevisions(newSelected));
+    dispatch(clearCheckedRevisions());
```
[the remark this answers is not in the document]
Carla-Moz: Ah yes, I keep defaulting to multiple actions. Bad habit. Yeah, I can clean this up and have the SearchSlice reducer handle clearing the checked revisions when a user confirms a selected revision. Thanks for mentioning.

src/thunks/repositoryThunk.ts
```diff
@@ -0,0 +1,34 @@
...
+>(
+  'search/fetchRecentRevisions',
+  async ({ repository, searchType }, { rejectWithValue }) => {
+    let response;
+
+    try {
+      response = await fetch(`${treeherderBaseURL}/api/perfcompare/results/?`);
+    } catch (err) {
+      const error = (err as Error).message;
+      return rejectWithValue(error);
```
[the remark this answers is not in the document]
Carla-Moz: Agreed, will update wherever I find.

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -1,36 +1,82 @@
...
+
+  const location = useLocation();
+  const { dispatchFetchCompareResults } = useFetchCompareResults();
+  const { searchByRevisionOrEmail } = useHandleChangeSearch();
+
+  useEffect(() => {
+    const searchParams = new URLSearchParams(location.search);
+    const repos = searchParams.get('repos')?.split(',');
+    const revs = searchParams.get('revs')?.split(',');
+    void dispatchFetchCompareResults(repos as Repository['name'][], revs);
```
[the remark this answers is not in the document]
Carla-Moz: Yes it should. Thanks for catching.

src/components/Search/SearchComponent.tsx
Carla-Moz: Scratch that. I'm going to handle this in the reducer to just get base and new revisions easily. Also, I forgot to put them in an array. -_- Hopefully my brain will work better next sprint.

Carla-Moz:
@julienw The test will take a day to complete so please feel to review the changes at this commit [resolved suggestions](https://github.com/mozilla/perfcompare/pull/497/commits/6c4ce2bbb91cc9f7a7f05bce56f14efe84b5454f).

src/thunks/repositoryThunk.ts
```diff
@@ -20,10 +20,11 @@ export const fetchRecentRevisions = createAsyncThunk<
     let response;
 
     try {
-      response = await fetch(`${treeherderBaseURL}/api/perfcompare/results/?`);
+      response = await fetch(`${treeherderBaseURL}/api/perfcompare/result/?`);
     } catch (err) {
-      const error = (err as Error).message;
-      return rejectWithValue(error);
+      const error = err as Error;
+      console.log('FetchRecentRevisions ERROR: ', error);
```
julienw: thanks for adding all these console.log!
optional nit: you can use `console.error` for this too, this will stand out with a red background in the devtools console.

src/hooks/useSelectedRevisions.ts
```diff
@@ -21,9 +20,12 @@ const useSelectRevision = () => {
   const addSelectedRevisions = () => {
     const newSelected = [...selectedRevisions];
     newSelected.push(...baseCheckedRevisions, ...newCheckedRevisions);
+    const filteredSelected = newSelected.filter(
+      (revision, index, self) =>
+        self.findIndex((r) => r.id === revision.id) === index,
+    );
```
julienw, line 26: Is it still needed now that you removed the duplicates?
If yes, then I don't understand really what this is doing :-)

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -39,9 +43,13 @@ function ResultsView(props: ResultsViewProps) {
...
     const repos = searchParams.get('repos')?.split(',');
     const revs = searchParams.get('revs')?.split(',');
-    void dispatchFetchCompareResults(repos as Repository['name'][], revs);
 
     if (revs && repos) {
+      dispatch(clearSelectedRevisions());
+
+      void dispatchFetchCompareResults(repos as Repository['name'][], revs);
+
+      //On component mount, use the repos and revs in hash to search for the base and new revisions. Store the results in state via the SelectedRevisionsSlice: see extra reducer, fetchRevisionsByID. Now can always display the selected revisions despite page refresh or copying and pasting url
```
julienw, line 63: nit: please split this on several lines.
(I think prettier would do it, but without prettier you can do it yourself :) )

src/hooks/useSelectedRevisions.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  const addSelectedRevisions = () => {
+    const newSelected = [...selectedRevisions];
+    newSelected.push(...baseCheckedRevisions, ...newCheckedRevisions);
+    const filteredSelected = newSelected.filter(
+      (revision, index, self) =>
+        self.findIndex((r) => r.id === revision.id) === index,
+    );
+
+    dispatch(setSelectedRevisions({ selectedRevisions: filteredSelected }));
```
julienw, line 28: I wonder if the action could rather be `addSelectedRevisions` and would handle the push. Then you don't need the `selectedRevisions` selector above. I feel like this is a more correct way of working with the reducer.

Also I see you removed `clearCheckedRevisions` like I suggested but I don't see the reducer change to clear these values. Is that not needed anymore?

src/components/Search/SelectedRevisions.tsx
```diff
@@ -29,23 +46,30 @@ function SelectedRevisions({
     return selectedRep;
   });
 
+  useEffect(() => {
+    if (selectedRevisions?.length) {
+      dispatch(clearCheckedRevisions());
+    }
+  }, [dispatch]);
```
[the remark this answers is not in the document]
julienw: Also it's not clear to me what this is doing. Is that to replace the call to `clearCheckedRevisions` when changing the selected revisions? I would have expected a `extraReducer` in SearchSlice to handle the `setSelectedRevisions` action (or `addSelectedRevisions` if you follow my suggeston above). See https://redux-toolkit.js.org/api/createSlice#extrareducers for more information about these, but I see there are already such reducers there, so it should be straightforward to just add another case.

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -2,52 +2,100 @@ import { useEffect } from 'react';
...
+  const { dispatchFetchCompareResults } = useFetchCompareResults();
+  const { searchByRevisionOrEmail } = useHandleChangeSearch();
+
+  useEffect(() => {
+    const urlSearchParams = new URLSearchParams(location.search);
+    const repos = urlSearchParams.get('repos')?.split(',');
+    const revs = urlSearchParams.get('revs')?.split(',');
+
+    if (revs && repos) {
+      dispatch(clearSelectedRevisions());
```
julienw, line 59: I kind of wonder if we want to clear them. What if the user wants to move back to the previous page, do we want them to select them again, or rather that they see the same revisions already selected? What are your thoughts on that?

julienw (commented):
Thanks for this!
I left a few more comments, but I also don't want to hold you back. Most of these comments could be handled in follow-ups or later, so feel free to move forward with what you have as long as the tests are green.

Of course it would be good to have another review after your test work and latest fixes, but Andra comes back next Wednesday only, so maybe @canova could give a quick sanity check before landing.

Also I think it's OK to not have 100% test coverage. Use your common sense about this as well :)

Keep the good work!

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -20,9 +23,10 @@ interface ResultsViewProps {
 }
 
 function ResultsView(props: ResultsViewProps) {
+  const dispatch = useAppDispatch();
   const { protocolTheme, toggleColorMode } = props;
   const themeMode = protocolTheme.palette.mode;
-  const view = 'compare-results';
+  const view = compareView;
```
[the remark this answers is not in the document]
Carla-Moz: True!

src/components/CompareResults/beta/ResultsView.tsx
Carla-Moz: Actually now that I'm looking at it, this isn't necessary at all. Whether or not the user moves back to the previous page, the home page, they still won't see the same revisions. They have to check the revisions they want again.

src/components/Search/SelectedRevisions.tsx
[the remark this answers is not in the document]
Carla-Moz: I think it would be clearer to say that `SelectedRevisions` only shows up on the Results View page and `Checked Revisions` on the Search Page. I cleared the checked revisions here so that they would not render on the Results View. The UI is different for both. 

My update: I will pass the `view` to the `SelectedRevisions` Component, eliminating the need for the whole `useEffect` above, and render `checked revisions` when the view is `'search'` and `selected revisions` when the view is `'compare-results'`. 

I will have the `SearchSlice `handle the `setSelectedRevisions` during the update process in the ResultsView.

src/hooks/useSelectedRevisions.ts
Carla-Moz: I'll add a comment. This is to remove duplicates in the revisions not the repositories. For example, if a user selects the same revisions for base and new.

src/hooks/useSelectedRevisions.ts
Carla-Moz: Yes, since the checked revisions will only render on the search page, I don't need to clear them. 

I'll revisit executing this action as `addSelectedRevisions` in the `SearchSlice` reducer in the next patch.

PR #498, opened 2023-07-05
Modified mocks and type CompareResultsItem

beatrice-acasandrei:
Frontend part of https://mozilla-hub.atlassian.net/browse/PCF-272.
Contains:
- [x] added base_rev and new_rev fields to CompareResultsItem
- [x] removed function that capitalizes the value of confidence_text, the backend will send the  capitalized value
- [x] deleted unnecessary fields: confidence_text_long (these values are kept in `/src/resources/Strings.tsx`), t_value_confidence, t_value_care_min
- [x] updated the `processResults` function so that it can differentiate between results based on the new revision


src/reducers/CompareResults.ts
```diff
@@ -1,8 +1,8 @@
 import { createSlice } from '@reduxjs/toolkit';
 
-import { comparisonResults as firstRevisionResults } from '../mockData/3c72ca1b13e1';
-import { comparisonResults as secondRevisionResults } from '../mockData/b4b491b3ca96';
-import { comparisonResults as thirdRevisionResults } from '../mockData/d0df2e2dd962';
+import { comparisonResults as secondRevisionResults } from '../mockData/9d5066525489';
+import { comparisonResults as thirdRevisionResults } from '../mockData/a998c42399a8';
+import { comparisonResults as firstRevisionResults } from '../mockData/bb6a5e451dac';
```
julienw, line 5: Not for this patch: we should probably use async imports so that the bundler put these away from the main bundle.

src/components/CompareResults/beta/ResultsTable.tsx
```diff
@@ -24,10 +24,12 @@ type Results = {
 function processResults(results: CompareResultsItem[]) {
   const processedResults: Map<string, CompareResultsItem[]> = new Map<string, CompareResultsItem[]>();
   results.forEach(result => {
-      if (processedResults.has(result.header_name)) {
-        (processedResults.get(result.header_name) as CompareResultsItem[]).push(result);
+      const { new_rev: newRevision, header_name: header } = result;
+      const rowIdentifier = header.concat(' ', newRevision);
```
julienw, line 28: 20y in this industry and I just learnt about `String.prototype.concat` :-)

julienw (approved):
`added base_rev and new_rev fields to CompareResultsItem`
Can you give some more information about that? Is it new from the server or was it already present in the server response before? If it's new, is it already deployed?

Otherwise this looks good to me :)

beatrice-acasandrei:
> `added base_rev and new_rev fields to CompareResultsItem` Can you give some more information about that? Is it new from the server or was it already present in the server response before? If it's new, is it already deployed?
> 
> Otherwise this looks good to me :)

You can look at https://mozilla-hub.atlassian.net/browse/PCF-272. I've added the changes to the response concomitantly with this PR, which has already been merged, but not yet deployed. I will ask tomorrow for a deploy so that Carla can have access to the new response when she's back on Monday.

src/reducers/CompareResults.ts
beatrice-acasandrei: I'll add a ticket for that, thanks.

julienw:
OK then it could be better to wait until it's deployed to merge this PR, otherwise https://beta--mozilla-perfcompare.netlify.app and local development will be broken.

PR #499, opened 2023-07-06
Results: Add logic to decide between single run or multiple runs

beatrice-acasandrei:
Added check to display the distribution graphs and info only if we have multiple runs for the base or new revision. Updated the hard-coded values to use actual values from the mock for now.

https://mozilla-hub.atlassian.net/browse/PCF-260


src/components/CompareResults/beta/RevisionRowExpandable.tsx
```diff
@@ -9,10 +9,17 @@ import Distribution from './Distribution';
 const strings = Strings.components.expandableRow;
 const { singleRun, note } = strings;
 
+function displayGraphDistribution(baseRuns: Array<number>, newRuns: Array<number>): boolean {
```
julienw, line 12: because this returns a boolean, it's better to name it like a question, for example `shouldDisplayGraphDistribution`

src/components/CompareResults/beta/RevisionRowExpandable.tsx
```diff
@@ -9,10 +9,17 @@ import Distribution from './Distribution';
...
+    return true;
+  return false;
+}
+
 
 function RevisionRowExpandable(props: RevisionRowExpandableProps) {
   const { themeMode, result } = props;
-  const { platform, delta_percentage: deltaPercent, delta_value: deltaVal, confidence_text: confidenceText } = result;
+  const { platform, delta_percentage: deltaPercent, delta_value: deltaVal, confidence_text: confidenceText, base_runs: baseRuns, new_runs: newRuns } = result;
+  const displayGraph = displayGraphDistribution(baseRuns, newRuns);
```
julienw, line 22: same thing for the variable

src/components/CompareResults/beta/RunValues.tsx
```diff
@@ -23,34 +23,36 @@ const styles = {
...
       <div>
         <GraphDistribution name={name} values={values} />
       </div>
       <div>
         <div className={styles.values}>
-          {values.map((v, index) => (
-            <div key={`${v}-${index}`} className={styles.value}>
-              {v}
+          {values.map((value, index) => (
+            <div key={`${value}-${index}`} className={styles.value}>
```
julienw, line 38: I believe it's not really useful to use the value if we already use the index that changes for every line. Just use the index as the key.

src/components/CompareResults/beta/RunValues.tsx
```diff
@@ -23,34 +23,36 @@ const styles = {
 };
 
 function RunValues(props: RunValuesProps) {
-  const { name, mean, runValues: values } = props.revisionRuns;
+  const { name, median, values, stddev, stddevPercent } = props.revisionRuns;
```
julienw, line 26: Any reason we're displaying median and not mean? I'm just curious.

Of course not for this patch: in the future we'll have to display something better than just median or mean, we'll have to find out if the distribution is bimodal / trimodal for example.

julienw (approved):
Looks good to me!

src/components/CompareResults/beta/RunValues.tsx
beatrice-acasandrei: I believe @gmierz suggested using median instead of mean back when we were starting to develop PerfCompare.

PR #501, opened 2023-07-10
Updated the sign and mean difference note logic

beatrice-acasandrei:


src/components/CompareResults/beta/RevisionRow.tsx
```diff
@@ -144,15 +150,15 @@ function RevisionRow(props: RevisionRowProps) {
         <div className='base-container'> {baseMedianValue} {baseUnit} </div>
       </TableCell>
       {/* TODO: Add logic for comparison sign */}
-      <TableCell className='comparison-sign'>&gt;</TableCell>
-      <TableCell className='new-value'>{newMedianValue} {newUnit}</TableCell>
+      <TableCell className='comparison-sign'>{determineSign(baseMedianValue, newMedianValue)}</TableCell>
```
julienw, line 153: The meaning of the sign isn't clear to me. Currently I see it's ">" is the old is higher than the new, but I would have expected the opposite.

Also I think we need a tooltip on this, to explain what is the meaning of this sign.

beatrice-acasandrei: I think the purpose of the sign is just to quickly identify which value is higher. Not sure what you meant by expecting the opposite.

julienw: I was expecting the opposite because in the "delta" column we do "new - base", but with the sign we do basically the opposite.
But let's do that and we can adjust later.

About the tooltip: I think like you that it's better when things in general don't need any tooltip. Indeed if there is a tooltip we also need some affordance (for example `<abbr>` automatically adds an underline to outline this), so that the user knows that a tooltip is available. The tooltip will be useful when the user might be confused. I believe this is the case here but we can decide later too.

julienw (approved):
OK let's move forward with that.

In the future, can you please give some more context about the change in the pull request description? For example a reference to the specification or something like that? That would make the review easier :-) Thanks!

PR #502, opened 2023-07-10
Results: Display the "No results found" message when data is not available

beatrice-acasandrei:


src/__tests__/CompareResults/beta/ResultsTable.test.tsx
```diff
@@ -19,4 +19,10 @@ describe('Results Table', () => {
     expect(screen.getByTestId('results-table')).toBeInTheDocument();
     expect(document.body).toMatchSnapshot();
   });  
+
+  it('Display message for not finding results', () => {
+    renderWithRouter(<ResultsTable themeMode={themeMode} />);
+
+    expect(screen.getByTestId('no-results-found-msg')).toBeInTheDocument();
```
julienw, line 26:
```suggestion
    expect(screen.getByText(/No results found/)).toBeInTheDocument();
```
And remove the useless test-id.

In general we try to avoid using test-ids and rely instead on what the user sees or can interact with.

julienw (approved):
thanks, looks good to me

It would probably be good to have `padding-bottom: 24px` on the `MuiContainer-root` in this case so that there's a white padding below the message, but I don't know how it would look in the case where there are results.

src/components/CompareResults/beta/NoResultsFound.tsx
```diff
@@ -16,7 +16,7 @@ function NoResultsFound() {
   }; 
 
   return (
-    <Container>
+    <Container data-testid='no-results-found-msg'>
       <Grid container alignItems='center' justifyContent='center'>
```
esanuandra, line 20: We need to add `paddingTop: Spacing.xxLarge` here, so that the magnifying glass is a bit lower.

src/components/CompareResults/beta/NoResultsFound.tsx
```diff
@@ -16,7 +16,7 @@ function NoResultsFound() {
   }; 
 
   return (
```
esanuandra, line 18: Are we going to add the `Learn more` button here or in another pr?

beatrice-acasandrei: In another PR it's best as we still need to decide where the user is redirected to after clicking this button.

esanuandra (approved):
Nice work!

PR #507, opened 2023-07-27
PCF - 250 Results: Add implementation for switching between comparisons Part 1

esanuandra:
This PR fulfills [Results: Add implementation for switching between comparisons](https://mozilla-hub.atlassian.net/browse/PCF-250)
subtask: [Add reducer for switching between comparison results](https://mozilla-hub.atlassian.net/browse/PCF-283)

STR:
1. Go to http://localhost:3000/#/compare-results/?fakedata=true or deploy link below + `/#/compare-results/?fakedata=true`
2. Click on `All revisions` dropdown
3. Change to another revision `9d5066525489`
4. It should display comparison results only for the selected revision `9d5066525489`


src/reducers/ComparisonSlice.ts
```diff
@@ -0,0 +1,28 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+interface IinitialState {
+  activeComparison: string;
+}
+
+const initialState: IinitialState = {
+  activeComparison: '',
```
beatrice-acasandrei, line 8:
```suggestion
  activeComparison: 'All revisions',
```
This is the default I believe.

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -31,16 +33,25 @@ function ResultsView(props: ResultsViewProps) {
...
   const fakeDataParam: string | null = searchParams.get('fakedata');
-
   // TODO: Populate store with real data or fake data pased on URL params
   useEffect(() => {
     if (fakeDataParam === 'true') {
       dispatch(switchToFakeData());
     }
   }, [fakeDataParam]);
 
+  useEffect(() => {
```
beatrice-acasandrei, line 49: Is it possible to use only one `useEffect` function?

src/reducers/ComparisonSlice.ts
```diff
@@ -0,0 +1,28 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+interface IinitialState {
```
beatrice-acasandrei, line 3:
```suggestion
interface InitialState {
```

src/reducers/CompareResults.ts
```diff
@@ -20,9 +20,14 @@ const compareResults = createSlice({
     switchToFakeData(state) {
```
beatrice-acasandrei, line 20: I think this reducer is no longer needed.

src/hooks/useComparison.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (activeComparison === truncateHash(firstRevisionResults[0].new_rev)) {
+      dispatch(setCompareData({ data: firstRevisionResults }));
+    }
+    if (activeComparison === truncateHash(secondRevisionResults[0].new_rev)) {
+      dispatch(setCompareData({ data: secondRevisionResults }));
+    }
+    if (activeComparison === truncateHash(thirdRevisionResults[0].new_rev)) {
+      dispatch(setCompareData({ data: thirdRevisionResults }));
+    }
+    if (activeComparison === 'All revisions') {
```
esanuandra, line 32: move 'All revisions' to strings file

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -31,15 +33,22 @@ function ResultsView(props: ResultsViewProps) {
       backgroundColor: background(themeMode),
     }),
   };
+  const activeComparison: string = useAppSelector(
+    (state: RootState) => state.comparison.activeComparison,
```
julienw, line 37: optional nit: by using ` useAppSelector` I believe you don't need to explicitely type `RootState`, that's exactly the purpose for `useAppSeelctor`.

```suggestion
    (state) => state.comparison.activeComparison,
```

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -31,15 +33,22 @@ function ResultsView(props: ResultsViewProps) {
...
   const [searchParams] = useSearchParams();
   const fakeDataParam: string | null = searchParams.get('fakedata');
-
   // TODO: Populate store with real data or fake data pased on URL params
   useEffect(() => {
     if (fakeDataParam === 'true') {
-      dispatch(switchToFakeData());
+      switchComparisonData();
+    }
+    if (fakeDataParam === 'false' || fakeDataParam === null) {
```
julienw, line 47: I'm not sure why you're using a second if block?
```suggestion
    } else {
```

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -31,15 +33,22 @@ function ResultsView(props: ResultsViewProps) {
...
+      // To be changed with real data
+      dispatch(setCompareData({ data: [] }));
     }
-  }, [fakeDataParam]);
+  }, [activeComparison, fakeDataParam]);
```
julienw, line 51: I don't think we're using `activeComparison`.

src/components/CompareResults/beta/RevisionHeader.tsx
```diff
@@ -51,12 +51,16 @@ function getExtraOptions(extraOptions: string) {
 function RevisionHeader(props: RevisionHeaderProps) {
   const { header } = props;
   const extraOptions = getExtraOptions(header.extra_options);
+  const shortHash = truncateHash(header.new_rev);
   return (
-    <TableRow className='revision-header'>
+    <TableRow
+      className='revision-header'
+      data-testid={`revision-header-${shortHash}`}
```
julienw, line 58: optional: using test-ids is not recommended by testing-library developers, instead it's better to directly look for actual displayed text using `getByText` or `getByRole` (or their `findBy*`counterparts).

For example in this case I believe that in the test you could do a `screen.getByText(shortHash)` instead.

src/__tests__/CompareResults/beta/RevisionSelect.test.tsx
```diff
@@ -1,12 +1,216 @@
...
+    renderWithRouter(
+      <ResultsView
+        protocolTheme={protocolTheme}
+        toggleColorMode={toggleColorMode}
+      />,
+    );
+
+    let comparisonHeaders = await waitFor(() =>
+      screen.getAllByTestId(/revision-header/),
+    );
```
julienw, line 70: use `await screen.findAllByTestId` directly, you don't need `waitFor`.


```suggestion
    let comparisonHeaders = await screen.findAllByTestId(/revision-header/);
```

src/components/CompareResults/beta/RevisionSelect.tsx
```diff
@@ -17,16 +21,25 @@ const styles = {
...
+  'bb6a5e451dac',
+  '9d5066525489',
+  'a998c42399a8',
 ];
 
 function RevisionSelect() {
+  const { handlerChangeComparison } = useComparison();
+  const { activeComparison } = useAppSelector((state: RootState) => {
+    return state.comparison;
+  });
```
julienw, line 33:
```suggestion
  const { activeComparison } = useAppSelector((state) => state.comparison);
```

src/components/CompareResults/beta/RevisionSelect.tsx
```diff
@@ -41,6 +54,8 @@ function RevisionSelect() {
             </Box>
           );
         }}
+        onChange={(event) => onChangeHandler(event)}
```
julienw, line 57: (optional) I think you can simplify by reducing the number of functions we need to find and read to understand what's happening:
```suggestion
        onChange={(event) => handlerChangeComparison(event.target.value)}
```
(possibly with `void` I'm not sure)

src/hooks/useComparison.ts
```diff
@@ -0,0 +1,47 @@
...
+import { useAppSelector } from './app';
+
+const useComparison = () => {
+  const dispatch = useDispatch();
+  const comparisonResults = firstRevisionResults.concat(
+    secondRevisionResults,
+    thirdRevisionResults,
+  );
+  const activeComparison: string = useAppSelector(
+    (state: RootState) => state.comparison.activeComparison,
```
julienw, line 20:
```suggestion
    (state) => state.comparison.activeComparison,
```

src/hooks/useComparison.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Strings } from '../resources/Strings';
+import { truncateHash } from '../utils/helpers';
+import { useAppSelector } from './app';
+
+const useComparison = () => {
+  const dispatch = useDispatch();
+  const comparisonResults = firstRevisionResults.concat(
+    secondRevisionResults,
+    thirdRevisionResults,
+  );
```
julienw, line 18: This will be executed everytime `useComparison` is called. I'd suggest to do it in the last if block in `switchComparisonData` instead.

src/components/CompareResults/beta/RevisionSelect.tsx
```diff
@@ -17,16 +21,25 @@ const styles = {
 
 const revisionsOptions = [
   'All revisions',
```
julienw, line 23: You should probably use the Strings here?

src/hooks/useComparison.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (activeComparison === truncateHash(firstRevisionResults[0].new_rev)) {
+      dispatch(setCompareData({ data: firstRevisionResults }));
+    }
+    if (activeComparison === truncateHash(secondRevisionResults[0].new_rev)) {
+      dispatch(setCompareData({ data: secondRevisionResults }));
+    }
+    if (activeComparison === truncateHash(thirdRevisionResults[0].new_rev)) {
+      dispatch(setCompareData({ data: thirdRevisionResults }));
+    }
+    if (activeComparison === allRevisions) {
```
julienw, line 35: I think the "All revisions" option should use a key that wouldn't be translated (the Strings will be localized in the future). This key will never change and can be compared in the event handler. Then the (hardcoded) key is what is compared, not the string.

src/hooks/useComparison.ts
```diff
@@ -0,0 +1,47 @@
...
+      dispatch(setCompareData({ data: comparisonResults }));
+    }
+  };
+
+  const handlerChangeComparison = (selectedRevision: string) => {
+    dispatch(updateComparison({ activeComparison: selectedRevision }));
```
julienw, line 41: I have several comments for this piece of code.

### 1. Keeping the same store structure, I believe we could call `switchComparisonData` directly here instead of using an effect in ResultsView.

Instead you're doing this dance:
1. The user changes the option.
2. We dispatch this action to change the string.
3. React renders the UI.
4. `useEffect` is called in the other component. This component calls `switchComparisonData`.
5. `switchComparisonData` dispatches again.
6. React renders the UI again.

This isn't performant, and this introduces inter-dependencies between the hook and the user of the hook, while the hook should better be more self-contained. With my proposal it would be much more self-contained.

See https://react.dev/learn/you-might-not-need-an-effect#chains-of-computations for a more complete explanation.

### 2. But another structure is possible

Instead of storing the full comparison data in the store, it could be _computed_ in a selector.

The store only has the source data to compute the comparison data:
* the full data (with `setCompareData`)
* the revision value

Then the store will _compute_ the returned value from the full data and the revision value. To handle the fakeData params, we can put the fake data as the full data during the fetch operation.

I think this is both easier to code, easier to follow, and easier to maintain.
Also I believe that the current code as it is now works fine with the fake data but will need to be redone with the full data. With the approach I'm suggesting here, the code would work with the fake data and with the real data.

Tell me what you think!

julienw (commented):
I left a few comments, but in short I think that the redux structure isn't the best and could be improved. Especially you could use some selectors.

Tell me what you think!

src/components/CompareResults/beta/ResultsTable.tsx
```diff
@@ -64,10 +65,28 @@ function processResults(results: CompareResultsItem[]) {
...
+    if (activeComparison === allRevisionsOption) {
+      return processResults(data);
+    } else {
+      const results = data.filter(
+        (rev) => truncateHash(rev.new_rev) === activeComparison,
+      );
+
+      return processResults(results);
+    }
+  });
```
julienw, line 89: This looks like a good selector!
Now the next step is to make it memoized using `createSelector`.

Step 1: extract this selector to a new function exported from the ComparisonSlice. It could be called `selectProcessedResults` for example. Of course `processResults` will need to be moved there too. In the future we will possibly want to move that to separate files but let's keep things simple for now.
Then you can import the function here and use in `useAppSelector`.
(documentation: https://redux.js.org/tutorials/essentials/part-2-app-structure#reading-data-with-useselector)

Step 2: use `createSelector` to actually make it memoized. `createSelector` is implemented by `reselect` but it's reexported by redux-toolkit so you can import it with `import { createSelector } from '@reduxjs/toolkit'`.

Then this can look something like this:
```js
import { createSelector } from '@reduxjs/toolkit';
...
export const selectProcessedResults = createSelector(
  // input simple or memoized selectors (in this case they're simple as they only access a property)
  state => state.compareResults.data,
  state => state.comparison.activeComparison,
  // output computed value
  (data, activeComparison) => {
    if (activeComparison === allRevisionsOption) {
      return processResults(data);
    } else {
      const results = data.filter(
        (rev) => truncateHash(rev.new_rev) === activeComparison,
      );

      return processResults(results);
    }
  }
);
```
The selector will be called for every rendering, but:
* if the state object doesn't change, then the previous memoized output will be returned
* if the state object changes, then the 2 input selectors are called. If their results are equals to the previous call, then the previously memoized output will be returned.
* The computing function will be called only if the state object changes and at least one of the input selector results changes.

This has 2 advantages.
1. If the computation is heavy, then we don't do it when we don't have to do it.
2. This ensures that the object reference stays the same if the inputs stays the same => the component where the selector is used won't rerender for no reason (see the doc for `useSelector` above).

(documentation:
https://react-redux.js.org/api/hooks#using-memoizing-selectors
https://redux.js.org/usage/deriving-data-selectors#writing-memoized-selectors-with-reselect
https://github.com/reduxjs/reselect
)

julienw (commented):
Yes, this looks a lot more like it, congrats for figuring it out!

The next step is to extract this selector to a memoized selector. I outlined this in a comment below.

I think it would be good that you fix the other few smaller comments that I left before and we can merge the commit like this without memoization.
Then you can work on the memoization in another patch.
This way we work incrementally and this doesn't get in the way of other patches.
Tell me what you think :-)

src/__tests__/CompareResults/beta/RevisionSelect.test.tsx
```diff
@@ -1,12 +1,212 @@
...
     renderWithRouter(<RevisionSelect />);
 
     expect(screen.getByTestId('revision-select')).toBeInTheDocument();
     expect(document.body).toMatchSnapshot();
   });
+
+  it("Should change 'All revisions' option to bb6a5e451dac", async () => {
+    renderWithRouter(<RevisionSelect />);
+
+    let selectButton = await waitFor(() => screen.getByRole('button'));
```
julienw, line 34: `await screen.findByRole` could be used here :-) 
Adnd the same is true for everywhere below, all the `waitFor` could be removed.

I'm not asking that you change all existing code, just your new code.

src/components/CompareResults/beta/ResultsTable.tsx
```diff
@@ -64,10 +65,23 @@ function processResults(results: CompareResultsItem[]) {
...
-  );
-  const processedResults = processResults(compareResults);
+  const allRevisionsOption =
+    Strings.components.comparisonRevisionDropdown.allRevisions;
+
+  const processedResults = useAppSelector((state) => {
+    const { data } = state.compareResults;
+    const { activeComparison } = state.comparison;
+
+    if (activeComparison === allRevisionsOption) {
```
julienw, line 75: In a follow-up, it would be good to change this to a key instead of using the string like I explained in another comment (that was hidden because of the code change).

src/components/CompareResults/beta/RevisionSelect.tsx
```diff
@@ -41,6 +55,8 @@ function RevisionSelect() {
             </Box>
           );
         }}
+        onChange={(event) => handlerChangeComparison(event.target.value)}
+        value={activeComparison}
       >
         {revisionsOptions.map((option) => (
           <MenuItem key={option} value={option}>
```
julienw, line 62: (For the follow-up to use a key for "All Revisions", this is where the key could be different than the value)

src/components/CompareResults/beta/RevisionHeader.tsx
julienw: Just so that you see this, this comment is still there. It's still optional to fix and could be done in a follow-up.

julienw (approved):
Thanks!
I didn't run the code so I trust you for running it and testing it manually.

src/components/CompareResults/beta/RevisionSelect.tsx
esanuandra: [PCF-287 Change string to key for 'All revisions' option](https://mozilla-hub.atlassian.net/browse/PCF-287)

src/components/CompareResults/beta/ResultsTable.tsx
esanuandra: [PCF-287 Change string to key for 'All revisions' option](https://mozilla-hub.atlassian.net/browse/PCF-287)

src/components/CompareResults/beta/ResultsTable.tsx
esanuandra: [PCF-288 Add selectProcessedResults with memoization](https://mozilla-hub.atlassian.net/browse/PCF-288)

src/hooks/useComparison.ts
esanuandra: [PCF-287 Change string to key for 'All revisions' option](https://mozilla-hub.atlassian.net/browse/PCF-287)

Issue #508, opened 2023-07-27
Add Home link button with arrow left on Results View Page


Carla-Moz:
We currently don't have a home button in the Results View page that allows the user to navigate back to the home page.

See ''Content Help" in the image below. 

<img width="893" alt="Screenshot 2023-07-27 at 14 10 21" src="https://github.com/mozilla/perfcompare/assets/88336547/31b2d958-71ef-46a4-8cda-290ccac7f10e">

beatrice-acasandrei:
This issue is fulfilling https://bugzilla.mozilla.org/show_bug.cgi?id=1846020.

PR #509, opened 2023-07-28
Results: Add Home link with arrow left

sumairq:
## This PR implements a "Home" link.
It is a solution for #508 issue. 
 A home link in the Results View page that allows the user to navigate back to the home page.
 

-  Used MUI's `Link` 
- Aligned text and icon using MUI's `Stack`   

image


src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -84,6 +87,12 @@ function ResultsView(props: ResultsViewProps) {
         view={compareView as View}
       />
       <section className={sectionStyles.container}>
+      <Link href="/" aria-label="link to home">
```
beatrice-acasandrei, line 90: I think you need to format this file. If you are using Visual Studio Code the easiest way to do so is to install the extension `Prettier - Code formatter` and use the `Format Document` option.

esanuandra (approved):

PR #512, opened 2023-08-03
Change page title based on route

sumairq:
## This PR implements title based on rendered view.
It is a solution for https://bugzilla.mozilla.org/show_bug.cgi?id=1846009.   

Changes in this PR replace the generic title "PerfCompare" and add more specific titles based on the view being rendered.

| View      | Title |
| ----------- | ----------- |
| Search     | PerfCompare - Search       |
| Results    | PerfCompare - Results         |

Used browser `document.title` API to change titles.


beatrice-acasandrei (approved):
LGTM

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -47,6 +48,10 @@ function ResultsView(props: ResultsViewProps) {
   const { dispatchFetchCompareResults } = useFetchCompareResults();
   const { searchByRevisionOrEmail } = useHandleChangeSearch();
 
+  useEffect(() => {
+    document.title = props.title;
```
julienw, line 52: you can use `title` directly here, this would be more correct too because that's the variable in the dependency below

julienw: The same comment applies below

src/components/App.tsx
```diff
@@ -74,6 +74,7 @@ function App() {
                   <SearchView
                     toggleColorMode={toggleColorMode}
                     protocolTheme={protocolTheme}
+                    title='PerfCompare - Search'
```
julienw, line 77: Probably these strings should be part of the `resources/Strings.tsx` instead of hardcoded.

sumairq:
Thank you for the comments. 🙌

PR #515, opened 2023-08-09
Results: Integrate results data

beatrice-acasandrei:
Acceptance criteria:

- [x] keep the results in state from all the comparations between base revisions and all the new revision
- [x] add framework param to the URL
- [x] use the selected framework to make the call
- [x] by default the displayed data should be a combination of responses from all the calls to the backend

### DIFF
--- a/src/components/CompareResults/beta/ResultsTable.tsx
+++ b/src/components/CompareResults/beta/ResultsTable.tsx
@@ -43,7 +43,6 @@ function processResults(results: CompareResultsItem[]) {
       processedResults.set(rowIdentifier, [result]);
     }
   });
-
   const restructuredResults: Results[] = Array.from(
     processedResults,
     function (entry) {
@@ -126,10 +125,9 @@ function ResultsTable(props: ResultsTableProps) {
               results={res.value}
             />
           ))}
-          {/* )} */}
         </Table>
       )}
-      {processedResults.length == 0 && <NoResultsFound />}
+      {!loading && processedResults.length == 0 && <NoResultsFound />}
     </TableContainer>
   );
 }
--- a/src/components/CompareResults/beta/ResultsTable.tsx
+++ b/src/components/CompareResults/beta/ResultsTable.tsx
@@ -77,7 +77,9 @@ function ResultsTable(props: ResultsTableProps) {
     const { activeComparison } = state.comparison;
 
     if (activeComparison === allRevisionsOption) {
-      const allResults = [].concat(...Object.values(data));
+      const allResults = ([] as CompareResultsItem[]).concat(
+        ...Object.values(data),
+      );
       return processResults(allResults);
     } else {
       const results = data[activeComparison];
--- a/src/components/CompareResults/beta/ResultsTable.tsx
+++ b/src/components/CompareResults/beta/ResultsTable.tsx
@@ -1,3 +1,5 @@
+import Box from '@mui/material/Box';
+import CircularProgress from '@mui/material/CircularProgress';
 import Paper from '@mui/material/Paper';
 import Table from '@mui/material/Table';
 import TableContainer from '@mui/material/TableContainer';
@@ -11,7 +13,6 @@ import type {
   RevisionsHeader,
   ThemeMode,
 } from '../../../types/state';
-import { truncateHash } from '../../../utils/helpers';
 import NoResultsFound from './NoResultsFound';
 import TableContent from './TableContent';
 import TableHeader from './TableHeader';
@@ -42,6 +43,7 @@ function processResults(results: CompareResultsItem[]) {
       processedResults.set(rowIdentifier, [result]);
     }
   });
+
   const restructuredResults: Results[] = Array.from(
     processedResults,
     function (entry) {
@@ -68,17 +70,17 @@ function ResultsTable(props: ResultsTableProps) {
   const allRevisionsOption =
     Strings.components.comparisonRevisionDropdown.allRevisions;
 
+  const loading = useAppSelector((state) => state.compareResults.loading);
+
   const processedResults = useAppSelector((state) => {
     const { data } = state.compareResults;
     const { activeComparison } = state.comparison;
 
     if (activeComparison === allRevisionsOption) {
-      return processResults(data);
+      const allResults = [].concat(...Object.values(data));
+      return processResults(allResults);
     } else {
-      const results = data.filter(
-        (rev) => truncateHash(rev.new_rev) === activeComparison,
-      );
-
+      const results = data[activeComparison];
       return processResults(results);
     }
   });
@@ -107,17 +109,24 @@ function ResultsTable(props: ResultsTableProps) {
       data-testid='results-table'
       sx={customStyles}
     >
-      <Table>
-        <TableHeader themeMode={themeMode} />
-        {processedResults.map((res, index) => (
-          <TableContent
-            themeMode={themeMode}
-            key={index}
-            header={res.revisionHeader}
-            results={res.value}
-          />
-        ))}
-      </Table>
+      {loading ? (
+        <Box display='flex' justifyContent='center' alignItems='center'>
+          <CircularProgress />
+        </Box>
+      ) : (
+        <Table>
+          <TableHeader themeMode={themeMode} />
+          {processedResults.map((res, index) => (
+            <TableContent
+              themeMode={themeMode}
+              key={index}
+              header={res.revisionHeader}
+              results={res.value}
+            />
+          ))}
+          {/* )} */}
+        </Table>
+      )}
       {processedResults.length == 0 && <NoResultsFound />}
     </TableContainer>
   );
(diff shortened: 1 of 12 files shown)


src/reducers/CompareResults.ts
```diff
@@ -24,6 +26,31 @@ const compareResults = createSlice({
       state.data = comparisonResults;
     },
   },
+  extraReducers: (builder) => {
+    builder
+      .addCase(fetchCompareResults.fulfilled, (state, action) => {
```
beatrice-acasandrei, line 31: I'm encountering issues here. When we select 3 new revisions, there's 3 calls to the backend and I'd like to store the results in state - but when I want to use this state I get error "Uncaught TypeError: state.compareResults.results[0] is undefined".

src/reducers/CompareResults.ts
```diff
@@ -24,6 +26,31 @@ const compareResults = createSlice({
...
+        state.data = action.payload;
+        const results = {
+          revision: action.payload[0].new_rev,
+          data: action.payload,
+        };
+        const fetchedResults = state.results.concat(results).filter(
```
beatrice-acasandrei, line 37: Removed duplicated data

src/hooks/useFetchCompareResults.ts
```diff
@@ -8,6 +8,7 @@ function useFetchCompareResults() {
   const dispatchFetchCompareResults = async (
     repos: Repository['name'][] | null,
     revs: string[] | undefined,
+    framework: string | undefined,
```
Carla-Moz, line 11: Shouldn't this be` framework: string` ?

src/reducers/CompareResults.ts
Carla-Moz: Hi Beatrice, you need to use a `useEffect `because the data hasn't loaded yet, that's why it's undefined. 

So in `src/components/CompareResults/beta/ResultsTable.tsx`
try instead:

 ```
 const results = useAppSelector(
    (state: RootState) => state.compareResults.results,
  );

  const [loading, setLoading] = useState(true);
  const [compareData, setCompareData] = useState<CompareResultsItem[]>([]);

  useEffect(() => {
    if (results.length > 0) {
      setCompareData(results[0].data);
      setLoading(false);
    }
  }, [results]);

  const processedResults = processResults(compareData);
```

and below:

```
         {loading ? (
            <caption>Loading...</caption>
          ) : (
            processedResults.map((res, index) => (
              <TableContent
                themeMode={themeMode}
                key={index}
                header={res.revisionHeader}
                results={res.value}
              />
            ))
          )}
```

Comment (sumairq) [inline src/reducers/CompareResults.ts:31]: ### Optional
Here are some of my observations, totally optional, if fixed can improve performance and load time.

- Due to some reason, the calls to the API endpoint are duplicated. For example, if there are 3 new revisions selected there are 6 calls to the API endpoint where there should be only 3.

- (Related to duplication of API calls) ResultsView appears to render twice initially triggering duplicated calls of the useEffect() that contains a call to `dispatchFetchCompareResults`. If we can figure out why ResultsView is rendering twice, we can solve the problem of duplicated API calls.

File: src/reducers/CompareResults.ts [hunk 4]
```diff
@@ -12,6 +13,7 @@ const comparisonResults = firstRevisionResults.concat(
 
 const initialState: CompareResultsState = {
   data: [],
+  results: [],
```
sumairq, line 16: ### Optional

This is an optional suggestion.

- We can wait for all the calls to the API to finish before rendering the results table. 
To achieve this we can add a piece of state that tracks the number of API requests. 
For example `completedApiRequests`
```
const initialState: CompareResultsState = {
  data: [],
  completedApiRequests: 0,
  results: [],
  loading: false,
  error: undefined,
};
```
We can increment this number after each fetch 
```
  .addCase(fetchCompareResults.fulfilled, (state, action) => {
        state.data = action.payload;
        state.completedApiRequests += 1;
```

In ResultsTable component we can utilize `completedApiRequests` by conditionally rendering the processedResults in the table when `completedApiRequests === selectedRevisions.new.length` is true.

src/reducers/CompareResults.ts
Carla-Moz: We will need a loader to give the user feedback while they wait instead of the "No results" component.

beatrice-acasandrei: It only happens locally because of [StrictMode](https://stackoverflow.com/questions/61254372/my-react-component-is-rendering-twice-because-of-strict-mode), but not on prod.

Carla-Moz (approved):

src/reducers/CompareResults.ts
beatrice-acasandrei: Carla left some suggestions in the comments. I followed her approach instead so I'll resolve this conversation since it's marked as optional.

src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -56,9 +56,14 @@ function ResultsView(props: ResultsViewProps) {
...
     const repos = urlSearchParams.get('repos')?.split(',');
     const revs = urlSearchParams.get('revs')?.split(',');
+    const framework = urlSearchParams.get('framework');
 
     if (revs && repos) {
-      void dispatchFetchCompareResults(repos as Repository['name'][], revs);
+      void dispatchFetchCompareResults(
+        repos as Repository['name'][],
+        revs,
+        framework as string,
```
julienw, line 65: Isn't it a string already? It's really not clear to me what we pass as the framework value...

src/thunks/compareResultsThunk.ts
```diff
@@ -10,16 +10,20 @@ export const fetchCompareResults = createAsyncThunk<
...
+  async (
+    { baseRepo, baseRev, newRepo, newRev, framework },
+    { rejectWithValue },
+  ) => {
     let response;
     try {
       //Note: We can now select and add the framework in the url
       response = await fetch(
-        `${treeherderBaseURL}/api/perfcompare/results/?base_repository=${baseRepo}&base_revision=${baseRev}&new_repository=${newRepo}&new_revision=${newRev}&framework=1&interval=86400&no_subtests=true`,
+        `${treeherderBaseURL}/api/perfcompare/results/?base_repository=${baseRepo}&base_revision=${baseRev}&new_repository=${newRepo}&new_revision=${newRev}&framework=${framework}&interval=86400&no_subtests=true`,
```
julienw, line 26: All these values are coming from the URL, so they can be controlled by somebody. Even if the server itself is public, I think it would be good to sanitize this better. We can do it simply with this code:
```js
const searchParams = new URLSearchParams({
  base_repository: baseRepo,
  base_revision: baseRev,
  new_repository: newRepo,
  new_revision: newRev,
  framework,
  interval: 86400,
  no_subtests: "true",
});
const url = `${treeherderBaseURL}/api/perfcompare/results/?${searchParams.toString()}`;
```
And then if any of the variables contain bad characters, they will be properly encoded.

julienw (commented):
Here are a few comments, tell me what you think!

src/reducers/CompareResults.ts
beatrice-acasandrei: Outdated code

src/components/CompareResults/beta/ResultsView.tsx
beatrice-acasandrei: It's the framework id from the URL param

src/thunks/compareResultsThunk.ts
```diff
@@ -10,16 +10,20 @@ export const fetchCompareResults = createAsyncThunk<
     baseRev: string;
     newRepo: Repository['name'];
     newRev: string;
+    framework: string;
```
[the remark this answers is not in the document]
beatrice-acasandrei: It's the framework id from the URL param

src/components/CompareResults/beta/ResultsTable.tsx
```diff
@@ -64,10 +68,25 @@ function processResults(results: CompareResultsItem[]) {
...
+
+  const [compareData, setCompareData] = useState<CompareResultsItem[]>([]);
+
+  useEffect(() => {
+    const data = [].concat(...Object.values(resultsData)); // concatenate all results
+    if (Object.keys(resultsData).length > 0) {
+      setCompareData(data);
+    }
+  }, [resultsData]);
+
```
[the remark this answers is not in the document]
beatrice-acasandrei: I will not add the code snippet to `processResults` because I want to be able to also display results from a selected revision, if I add this logic then I'll only display the results from all the revisions. I removed the `useEffect()`.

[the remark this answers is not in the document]
julienw: as long as there's no state + effect, I'm fine. In the future we'll move this selector to a memoized selector anyway, so we'll be able to revisit.

src/components/CompareResults/beta/ResultsView.tsx
beatrice-acasandrei: It's the value from the URL param. If the param is not mentioned in URL then it could be null.

julienw (approved):

PR #517, opened 2023-08-10
Allow edit revisions

Carla-Moz:
WIP - Need to add test coverage

This is a big PR, and I apologize but it made sense to do this in one run. There a lot of file changes because I did some refactoring but I'll be sure to direct you to the relevant files for this PR.

This PR closes the [Results: Display the inputs selected for Compare with a base PCF-215](https://mozilla-hub.atlassian.net/browse/PCF-215) epic

Specifically the following sub-tasks:

- [x] [Allow adding new revisions PCF-258](https://mozilla-hub.atlassian.net/browse/PCF-258)
- [x] [ Differences: The search is shown PCF-259](https://mozilla-hub.atlassian.net/browse/PCF-259)
- [x] [Allow editing through edit icon PCF-257](https://mozilla-hub.atlassian.net/browse/PCF-257)

I made some design changes in this PR related to editing and updating the revisions in the Results View. 

**In the original design, there is an edit button for each selected revision.** 
Screenshot 2023-08-10 at 16 45 30

**In my PR, there is an edit button for the Base and Revision sections respectively**
Screenshot 2023-08-10 at 16 11 40

**In the original design, clicking the edit buttons shows the repo dropdown and search input for each selected revision.**
Screenshot 2023-08-10 at 16 47 26

**I used instead the Search View design; one dropdown and search input per revision type, base and new**
Screenshot 2023-08-10 at 16 16 37

This was easier than creating multiple instances of the dropdown and search input for each selected revision. Also, editing the revision is now very similar to checking revisions in the Search View. 

Most relevant files for this PR:
```
src/hooks/useSelectedRevisions.ts
src/reducers/SearchSlice.ts
src/components/Search/SearchComponent.tsx
src/components/CompareResults/beta/ResultsView.tsx
```


sumairq (commented):
### Optional
- [Optional] It might be preferable to add a `notistack` alert when the user tries to compare without saving the edit session.
The alert could warn users with a message like **'Please save your edits before comparing'**.

An alert might be better than users getting confused about why the results do not get updated according to their edits (when they don't press the save button).

 _This could be done in a subsequent PR as well._

src/components/Search/SelectedRevisionItem.tsx
```diff
@@ -82,14 +81,33 @@ function SelectedRevisionItem({
...
 
+  const renderCloseIcon = () => {
+    const isSearchViewOrEditBtnVisible = view === searchView || !editBtnVisible;
+    const isBaseAndEditBtnVisible = searchType === 'base' || !editBtnVisible;
+
+    const iconClassName = isSearchViewOrEditBtnVisible
+      ? 'icon icon-close'
+      : isBaseAndEditBtnVisible
+      ? 'icon icon-close-base'
+      : 'icon icon-close results';
```
julienw, line 100: I don't see where the CSS for `results` is, am I missing something?

src/components/Search/SelectedRevisionItem.tsx
```diff
@@ -82,14 +81,33 @@ function SelectedRevisionItem({
...
+      removeCheckedRevision(item);
     } else {
       deleteSelectedRevisions(item);
     }
     return;
   };
 
+  const renderCloseIcon = () => {
+    const isSearchViewOrEditBtnVisible = view === searchView || !editBtnVisible;
+    const isBaseAndEditBtnVisible = searchType === 'base' || !editBtnVisible;
```
julienw, line 94: It could be good to have some comment outlining a bit more the various contexts that this component is used in, because there are a lot of cases here.

src/hooks/useSelectedRevisions.ts
```diff
@@ -23,22 +25,67 @@ const useSelectRevision = () => {
...
     newSelected.push(...baseCheckedRevisions, ...newCheckedRevisions);
-    //remove duplicates of the same revision
-    const filteredSelected = newSelected.filter(
-      (revision, index, self) =>
-        self.findIndex((r) => r.id === revision.id) === index,
-    );
+    //create a new array with unique values
+    const filteredSelected = [...new Set(newSelected)];
 
     dispatch(setSelectedRevisions({ selectedRevisions: filteredSelected }));
```
julienw, line 31: thought for a later patch (but not for now): using a Set for `selectedRevisions` would make it unnecessary to convert it back and forth between arrays. But we would till have to create a new set when changing it, and this may have performance concerns.
although in this case where it's at most 3 revisions the performance concerns are not an issue.

src/hooks/useSelectedRevisions.ts
```diff
@@ -23,22 +25,67 @@ const useSelectRevision = () => {
...
   };
 
+  const editSelectedRevisions = (searchType: InputType) => {
+    let revisionsForEdit = [...selectedRevisions];
```
julienw, line 35: the clone is unnecessary because you're already cloning it below.

src/hooks/useSelectedRevisions.ts
```diff
@@ -23,22 +25,67 @@ const useSelectRevision = () => {
...
+    switch (searchType) {
+      case 'base':
+        revisionsForEdit = [revisionsForEdit[0]];
+        break;
+      case 'new':
+        revisionsForEdit = revisionsForEdit.slice(1);
+        break;
+      default:
```
julienw, line 43: if the default isn't possible, it would be good to throw an error so that we find programming errors easily.
(same below)

src/hooks/useSelectedRevisions.ts
```diff
@@ -23,22 +25,67 @@ const useSelectRevision = () => {
...
+      case 'base':
+        //remove old base revision
+        updatedRevisions = updatedRevisions.slice(1);
+        //add new base revision
+        updatedRevisions.unshift(...baseCheckedRevisions);
+        break;
+      case 'new':
+        //keep the base revision, add the new checked revisions
+        updatedRevisions = [
+          ...updatedRevisions.slice(0, 1),
```
julienw, line 64:
```suggestion
          updatedRevisions[0]
```

src/hooks/useSelectedRevisions.ts
```diff
@@ -23,22 +25,67 @@ const useSelectRevision = () => {
...
+      default:
+        break;
+    }
+    dispatch(
+      setCheckedRevisionsForEdit({ revisions: revisionsForEdit, searchType }),
+    );
+  };
+
+  const updateSelectedRevisions = (searchType: InputType) => {
+    let updatedRevisions = [...selectedRevisions];
```
julienw, line 52: same as above: you're already creating new objects below so you don't need to do it here.

julienw (approved):
I just skimmed quickly through the changes but this looks mostly good to me. I didn't try to run the code either so please make sure you test the code manually too.
Please tell me if you'd like a more thorough review, but otherwise I think this could go in like this and polish can happen later.

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -63,9 +62,7 @@ function SearchResultsListItem({
   const { handleToggle } = useCheckRevision(searchType);
   const revisionHash = truncateHash(item.revision);
   const commitMessage = getLatestCommitMessage(item);
-  const maxRevisions = view == 'compare-results' ? 1 : 3;
-  const isBase = searchType == 'base' ? 1 : maxRevisions;
-
+  const isBase = searchType === 'base' ? 1 : 3;
```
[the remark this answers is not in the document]
Carla-Moz: Oh right, I agree.

src/hooks/useSelectedRevisions.ts
Carla-Moz: Good idea

src/components/Search/SelectedRevisionItem.tsx
Carla-Moz: Yeah, I simplified this. I don't know what I was thinking here lol. This is why I appreciate PR reviews lol.

PR #520, opened 2023-08-25
Results: Enable "Download JSON" button

sumairq:
## This PR enables the "Download JSON" button 

- In `Button` a data URL for the JSON data is generated to embed data directly into the URL.
- Set the download attribute to `download="perf-compare.json"` i.e. the downloaded file will have the name `perf-compare.json`
- The `formatDownloadData` function is implemented to give data a different structure similar to treeherder.

The restructured data has the following structure:

image

- On the top level we have an Array [ ]
- On the second level we have an object with a property created from the header_name property of the result item and an array of objects as its value.
- Against the key header_name, the result items with the same header_name are grouped together.

screenshot of example data:
image


src/components/CompareResults/beta/DownloadButton.tsx
```diff
@@ -14,9 +15,21 @@ const styles = {
...
   return (
     <div className={styles.downloadButton}>
-      <Button disabled>Download JSON</Button>
+      <Button
+        href={`data:text/json;charset=utf-8,${encodeURIComponent(
+          JSON.stringify(results),
+        )}`}
+        download='perf-compare.json'
+      >
+        Download JSON
```
sumairq, line 31: @beatrice-acasandrei  Here I have not changed the format of results from the redux store. I am not sure if this would be correct or not. Do I need to convert the results into a specific format? 
(Results are converted to JSON as it is here)

src/components/CompareResults/beta/DownloadButton.tsx
```diff
@@ -13,10 +14,61 @@ const styles = {
...
+
+interface ResultObject {
+  header_name: string;
+}
+
 function DownloadButton() {
+  const results = useAppSelector((state) => {
+    const { data } = state.compareResults;
+    return data;
+  });
```
beatrice-acasandrei, line 29:
```suggestion
  const results = useAppSelector((state) => state.compareResults.data)
```

src/components/CompareResults/beta/DownloadButton.tsx
```diff
@@ -13,10 +14,58 @@ const styles = {
...
+interface GroupedResults {
+  [key: string]: object[];
+}
+
+interface ResultObject {
+  header_name: string;
+}
+
 function DownloadButton() {
+  const results = useAppSelector((state) => state.compareResults.data);
```
esanuandra, line 26: This is good in case the activeComparison is 'All revisions' otherwise we should something like this
```suggestion
  const results = useAppSelector((state) => state.compareResults.data[activeComparison]);
```

beatrice-acasandrei: Good catch!

src/components/CompareResults/beta/DownloadButton.tsx
```diff
@@ -13,10 +15,73 @@ const styles = {
...
+          }
+          grouped[result.header_name].push(result);
+          return grouped;
+        },
+        {},
+      );
+
+      const transformedGroupEntries = Object.keys(groupedResults).map(
+        (header_name) => ({
+          [`${header_name}${truncateHash(
```
beatrice-acasandrei, line 63:
```suggestion
          [`${header_name} ${truncateHash(
```

src/components/CompareResults/beta/DownloadButton.tsx
```diff
@@ -13,10 +15,73 @@ const styles = {
...
+
+interface ResultObject {
+  header_name: string;
+  new_rev: string;
+}
+
 function DownloadButton() {
+  let fileName = 'perf-compare-all-revisions.json';
+  const results = useAppSelector((state) => {
+    if (state.comparison.activeComparison === 'All revisions') {
```
esanuandra, line 30: you can import `'All revisions'` from `resources/Strings.tsx`

esanuandra (approved):

beatrice-acasandrei (approved):

src/components/CompareResults/beta/DownloadButton.tsx
```diff
@@ -13,10 +16,76 @@ const styles = {
...
+  const results = useAppSelector((state) => {
+    if (
+      state.comparison.activeComparison ===
+      Strings.components.comparisonRevisionDropdown.allRevisions
+    ) {
+      return state.compareResults.data;
+    } else {
+      fileName = `perf-compare-${truncateHash(
+        state.comparison.activeComparison,
+      )}.json`;
```
julienw, line 39: changing the filename inside the selector doesn't sound right. Selectors shouldn't have side effects.
Instead you can make the fileName the result of another `useAppSelector` call.

src/components/CompareResults/beta/DownloadButton.tsx
```diff
@@ -13,10 +16,76 @@ const styles = {
...
+      return {
+        [state.comparison.activeComparison]:
+          state.compareResults.data[state.comparison.activeComparison],
+      };
+    }
```
julienw, line 44: This should probably be a memoized selector using `createSelector`, returning the full result of `encodeURIComponent(JSON.stringify(formatDownloadData(results)))`.
Because currently we do all this at each state change, and that doesn't sound right performance-wise, especially if the result data is big.

src/components/CompareResults/beta/DownloadButton.tsx
```diff
@@ -13,10 +16,76 @@ const styles = {
...
+  };
+
   return (
     <div className={styles.downloadButton}>
-      <Button disabled>Download JSON</Button>
+      <Button
+        disabled={!results}
+        href={`data:text/json;charset=utf-8,${encodeURIComponent(
+          JSON.stringify(formatDownloadData(results)),
+        )}`}
```
julienw, line 84: Ideally we should use a Blob and a blob url, that would avoid the need to `encodeURIComponent` and also would avoid putting the full data in the DOM which can have a performance issue if it's big. But it is a bit difficult to get right. Especially it's important to revoke the url and with React it's not super easy, [we have a component just for this in the profiler code](https://github.com/firefox-devtools/profiler/blob/main/src/components/shared/BlobUrlLink.js) that we could reuse... But let's keep it for later.

PR #532, opened 2023-09-05
PCF-289 Remove data-testid for TableRow in RevisionHeader

esanuandra:
This PR is a follow up for #507 to address the [comment](https://github.com/mozilla/perfcompare/pull/507#discussion_r1290110332) about removing the data-testid

[Jira ticket](https://mozilla-hub.atlassian.net/browse/PCF-289)

### DIFF
--- a/src/__tests__/CompareResults/beta/RevisionSelect.test.tsx
+++ b/src/__tests__/CompareResults/beta/RevisionSelect.test.tsx
@@ -96,6 +96,8 @@ describe('Revision select', () => {
     fireEvent.mouseDown(selectButton);
 
     // check to display results only for revision bb6a5e451dac
+    // findByRole doesn't work anymore because the root element
+    // has aria-hidden but we don't know why yet
     firstRevisionHeaders = await screen.findAllByText(/bb6a5e451dac/, {
       selector: 'a',
     });
@@ -159,6 +161,8 @@ describe('Revision select', () => {
     fireEvent.mouseDown(selectButton);
 
     // check to display results only for revision 9d5066525489
+    // findByRole doesn't work anymore because the root element
+    // has aria-hidden but we don't know why yet
     const secondRevisionHeaders = await screen.findAllByText(/9d5066525489/, {
       selector: 'a',
     });
@@ -222,6 +226,8 @@ describe('Revision select', () => {
     fireEvent.mouseDown(selectButton);
 
     // check to display results only for revision a998c42399a8
+    // findByRole doesn't work anymore because the root element
+    // has aria-hidden but we don't know why yet
     const thirdRevisionHeaders = await screen.findAllByText(/a998c42399a8/, {
       selector: 'a',
     });
--- a/src/__tests__/CompareResults/beta/RevisionSelect.test.tsx
+++ b/src/__tests__/CompareResults/beta/RevisionSelect.test.tsx
@@ -61,12 +61,24 @@ describe('Revision select', () => {
       />,
     );
 
-    let comparisonHeaders = await screen.findAllByTestId(/revision-header/);
-    expect(comparisonHeaders[0].textContent).toContain('bb6a5e451dac');
+    // check to display results for all revisions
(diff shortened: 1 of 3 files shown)


src/__tests__/CompareResults/beta/RevisionSelect.test.tsx
```diff
@@ -133,10 +163,24 @@ describe('Revision select', () => {
...
+    secondRevisionHeaders = secondRevisionHeaders.filter((header) =>
+      header.hasAttribute('href'),
+    );
+    expect(secondRevisionHeaders.length).toBe(7);
 
-    expect(firstRevisionHeader[0].textContent).toContain('9d5066525489');
+    let thirdRevisionHeaders = await screen.findAllByText(/a998c42399a8/);
+    thirdRevisionHeaders = thirdRevisionHeaders.filter((header) =>
+      header.hasAttribute('href'),
+    );
```
julienw, line 182: It's not super clear to me what is the thing with the attribute. Are we looking for links only?

In that case, I believe this would work:
```js
await screen.findByRole('link', { name: /a998c42399a8/ });
```

Links with an href attribute have the ARIA role "link". The `name` is the accessible name and can come from various sources, but is usually the text of the element.

More about this query in https://testing-library.com/docs/queries/byrole#description

esanuandra: hey @julienw,
`findAllByText` was giving me also the elements from the select dropdown. 
The revision headers from the table have the push link and with that I could differentiate them from the select options.

I added your suggestion with `findByRole`. It works alright for the first part of the tests
After changing the select option, `findByRole` can't find the revision headers. It looks like they don't have the `role` attribute anymore which is strange for me.

esanuandra: I replaced the `findAllByRole` with `queryAllByRole` for the headers that should not be displayed.

I have a problem [here](https://github.com/mozilla/perfcompare/pull/532/commits/8f078d7f17321efb2d6beda54d5859c1b351a51f#diff-f96d2299bd0dcd9bced027859ee7785025d4a0f23e61e822ef73aa074ba3b293R99) where `findAllByRole` should find the displayed revision headers for the selected revision option.

julienw: I spent too much time debugging this :)
for a reason that I don't understand yet, the top `div` has `aria-hidden`  at this point (but not at the start!). The `byRole` query looks at this, and so it decides that it's not available. They could give more debugging information about that, I'll file a bug there...

I still don't know why the top `div` has `aria-hidden` but I feel like i spent too much time on this at the moment. So I'd suggest to replace the `xxxByRole('link', { name: /xxx/ })` with `xxxByText(/xxx/, { selector: 'a' })`.

It's especially important to change this for the ones that return nothing too, to prevent false nagatives.

Please tell me if that works for you.

esanuandra: `xxxByText(/xxx/, { selector: 'a' })` worked great! Thank you very much for looking into this.

julienw (approved):
thanks!

src/__tests__/CompareResults/beta/RevisionSelect.test.tsx
```diff
@@ -86,12 +95,23 @@ describe('Revision select', () => {
...
     fireEvent.mouseDown(selectButton);
 
-    // check every revision header to be for revision bb6a5e451dac
-    comparisonHeaders = await screen.findAllByTestId(/revision-header/);
-
-    comparisonHeaders.forEach((resultHeader) =>
-      expect(resultHeader.textContent).toContain('bb6a5e451dac'),
-    );
+    // check to display results only for revision bb6a5e451dac
+    firstRevisionHeaders = await screen.findAllByText(/bb6a5e451dac/, {
```
julienw, line 99: Maybe add a comment about the fact that findByRole doesn't work anymore because the root element has `aria-hidden` but we don't know why yet?

PR #533, opened 2023-09-06
Fix existing typescript errors

julienw:
I plan to add typescript in the CI in a next patch.


julienw:
Removing the r request because the linting job is failing.

src/components/Search/FrameworkDropdown.tsx
```diff
@@ -118,20 +110,15 @@ function FrameworkDropdown({ view, mode }: FrameworkDropdownProps) {
...
           name='Framework'
         >
           {Object.entries(frameworkMap).map(([id, name]) => (
-            <MenuItem
-              id={id}
-              value={name}
-              key={name}
-              className='framework-dropdown-item'
-            >
+            <MenuItem value={id} key={name} className='framework-dropdown-item'>
```
julienw, line 121: We now use the `id` as the `value`. This triggered a series of other changes in this file. For example we need to convert back and forth between number and string because of how the types are made.

esanuandra (approved):

PR #534, opened 2023-09-06
Configure several jobs in CircleCI and remove the nightly job

julienw:
The goal is to split the big job into separate ones, so that they can run in parallel. I did it especially so that adding the typescript check wouldn't make the overall CI run too long.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,54 +1,57 @@
 version: 2.1
 orbs:
-  node: circleci/node@5.0.1
+  node: circleci/node@5.1.0
   codecov: codecov/codecov@3.2.2
+
 jobs:
-  test:
-    executor:
-      name: node/default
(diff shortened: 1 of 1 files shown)


Carla-Moz (approved):
LGTM

julienw:
The issue is caused by the CI running in a more recent version of node. This [hits a problem](https://github.com/facebook/jest/issues/11956) we found also in the profile code, and I used the same fix here (that is, adding `-w=4 --workerIdleMemoryLimit=1.5G` to the command, which works well with the resource_class `large` in circleci. The good news is that this problem has finally been fixed in v8 and we're now waiting for the fix to propagate to a version of node.

I wasn't sure about the version of node to use.  It looks like the default (that is using the current `lts`) is working currently so I think we can use this setup for now, and revisit if this is a problem in the future.

PR #535, opened 2023-09-07
PCF-290 Create memoized selector for populating all revisions dropdown

esanuandra:
This PR is a follow up for #516 to address the [comment](https://github.com/mozilla/perfcompare/pull/516#discussion_r1297358956) about adding memoized select for newRevisions in all revisions dropdown

[Jira ticket](https://mozilla-hub.atlassian.net/browse/PCF-290)

### DIFF
--- a/src/reducers/ComparisonSlice.ts
+++ b/src/reducers/ComparisonSlice.ts
@@ -91,9 +91,9 @@ export const selectProcessedResults = createSelector(
 );
 
 export const selectNewRevisions = createSelector(
-  (state: RootState) => state.selectedRevisions,
-  (selectedRevisions) => {
-    return selectedRevisions.new.map((item) => item.revision);
+  (state: RootState) => state.selectedRevisions.new,
+  (newSelectedRevisions) => {
+    return newSelectedRevisions.map((item) => item.revision);
   },
 );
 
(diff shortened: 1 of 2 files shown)


src/reducers/ComparisonSlice.ts
```diff
@@ -90,6 +90,13 @@ export const selectProcessedResults = createSelector(
   },
 );
 
+export const selectNewRevisions = createSelector(
+  (state: RootState) => state.selectedRevisions,
+  (selectedRevisions) => {
+    return selectedRevisions.new.map((item) => item.revision);
+  },
+);
```
julienw, line 98: I'd even suggest something like this:
```suggestion
export const selectNewRevisions = createSelector(
  (state: RootState) => state.selectedRevisions.new,
  (newSelectedRevisions) => {
    return newSelectedRevisions.map((item) => item.revision);
  },
);
```

so that the selector doesn't need to be recomputed if just the "old" revision changes.

julienw: I see that in the reducer they all change when one changes, so currently the behavior I describe above would still happen, but this can be fixed separately.

julienw (approved):
I didn't check the deploy preview, but the code looks good to me. So this is good to me with the suggested change!

PR #536, opened 2023-09-07
Lazy load the fake data instead of including it in the bundle

julienw:
I did a few passing-by changes (sorry!), but I isolated them into separate commits, so I believe it will be easier to look at the 5 commits separately. Also I left comments explaining some of the choices, hopefully that will make things easier for you!

[Deploy preview](https://deploy-preview-536--mozilla-perfcompare.netlify.app/#/compare-results?fakedata=true)

The production bundle goes from 861 kB to 780 kB.

Next step could be to lazy load the compare results page. I believe it's possible and easy-ish with the current version of react.


src/components/CompareResults/beta/ResultsView.tsx
```diff
@@ -64,44 +63,44 @@ function ResultsView(props: ResultsViewProps) {
...
         void searchByRevisionOrEmail(
-          repos[index] as Repository['name'],
+          reposArray[index] as Repository['name'],
           rev,
           'new',
         );
       });
     }
-  }, []);
+  }, [repos, revs, framework]);
```
julienw, line 91: I believe this will make the back button work properly, because now the search will be repeated as soon as the dependencies change.

Maybe the second part should be separated though, because it doesn't use `framework`... We can do that after sumair patch that touches also this area.

src/components/CompareResults/beta/RevisionSelect.tsx
```diff
@@ -30,9 +30,9 @@ const allRevisionsOption =
 
 const fakeRevisionsOptions = [
   allRevisionsOption,
-  'bb6a5e451dac',
-  '9d5066525489',
-  'a998c42399a8',
+  'bb6a5e451dace3b9c7be42d24c9272738d73e6db',
```
julienw, line 33: now this uses the same mechanism as "normal" queries, we have to use the full `new_rev` value.

src/thunks/compareResultsThunk.ts
```diff
@@ -44,3 +45,21 @@ export const fetchCompareResults = createAsyncThunk<
...
+
+export const fetchFakeResults = createAsyncThunk<
+  CompareResultsItem[],
+  FakeCommitHash,
+  { rejectValue: string }
+>(
+  'compareResults/fetchFakeResults',
+  async (commitHash, { rejectWithValue }) => {
+    try {
+      const module = (await import(`../mockData/${commitHash}`)) as {
```
julienw, line 57: This is where the magic happens.

src/reducers/CompareResults.ts
```diff
@@ -1,44 +1,56 @@
...
-      .addCase(fetchCompareResults.rejected, (state, action) => {
-        state.error = action.payload;
-        state.loading = initialState.loading;
-      });
+      .addCase(fetchCompareResults.pending, fetchPending)
+      .addCase(fetchFakeResults.pending, fetchPending)
+      .addCase(fetchCompareResults.fulfilled, fetchFulfilled)
+      .addCase(fetchFakeResults.fulfilled, fetchFulfilled)
+      .addCase(fetchCompareResults.rejected, fetchRejected)
+      .addCase(fetchFakeResults.rejected, fetchRejected);
```
julienw, line 52: I tried to use `addMatcher` to specify each reducer only once, but I found that the result were more messy.

src/__tests__/CompareResults/beta/ResultsView.test.tsx
```diff
@@ -178,10 +175,12 @@ describe('Results View', () => {
       />,
     );
 
-    const expandButtons = screen.getAllByTestId('expand-revision-button');
+    const expandButtons = await screen.findAllByTestId(
+      'expand-revision-button',
+    );
     await user.click(expandButtons[0]);
-    const expandedContent = await waitFor(() =>
```
julienw, line 183: I also used this opportunity to remove `waitFor` calls that can be replaced with a `find`.

src/__tests__/hooks/useFetchCompareResults.test.ts
```diff
@@ -23,7 +23,7 @@ describe('Tests useFetchCompareResults', () => {
       },
     } = renderHook(() => useFetchCompareResults(), { wrapper: StoreProvider });
     const spyOnFetch = jest.spyOn(global, 'fetch');
-    await dispatchFetchCompareResults(['fenix'], ['testRev'], '1');
+    dispatchFetchCompareResults(['fenix'], ['testRev'], '1');
```
julienw, line 26: and I removed some useless `await`. I kept the `async` on the function because we might need it in the future so I thought this would be easier like this.

src/__tests__/CompareResults/beta/ResultsTable.test.tsx
```diff
@@ -1,27 +1,30 @@
...
 import ResultsTable from '../../../components/CompareResults/beta/ResultsTable';
 import useProtocolTheme from '../../../theme/protocolTheme';
 import { renderWithRouter } from '../../utils/setupTests';
 import { renderHook, screen } from '../../utils/test-utils';
 
-jest.mock('react-router-dom', () => ({
-  ...jest.requireActual<typeof import('react-router-dom')>('react-router-dom'),
-  useSearchParams: () => [new URLSearchParams({ fakedata: 'true' })],
-}));
+function renderWithRoute(component: ReactElement) {
```
julienw, line 8: I'm not super happy with this name, but it's not bad either, tell me what you think!

beatrice-acasandrei (approved):
Thanks!

PR #538, opened 2023-09-08
Remove a syntax error in the CSS for the buttons, giving a warning in…

julienw:
… tests

I checked that there isn't any visible UI change as a result of this fix.


Carla-Moz (approved):
Thanks for catching this.

esanuandra (approved):

PR #539, opened 2023-09-11
Results: Enable "Download JSON" button (Follow up)

sumairq:
## This PR is a follow-up of  [#520](https://github.com/mozilla/perfcompare/pull/520) 

- Implemented a `fileName` selector.
- Used `createSelector()` and `React.memo()` to ensure the state is only recalculated when changed.
- Used `blob` instead of putting all the data inside the URL.
- used `disabled={blobUrl === null}` to ensure the button is disabled until we have data available.


julienw (commented):
I left a few comments, tell me what you think

src/reducers/ComparisonSlice.ts
```diff
@@ -33,6 +43,56 @@ const comparison = createSlice({
...
+    const transformedGroupEntries = Object.keys(groupedResults).map(
+      (header_name) => ({
+        [`${header_name} ${truncateHash(
+          groupedResults[header_name][0].new_rev,
+        )}`]: groupedResults[header_name],
+      }),
+    );
+
+    transformedGroups.push(...transformedGroupEntries);
+  });
```
julienw, line 73: I'm not a big fan of using `forEach` or `reduce` for complicated things, I think this makes things generally difficult to read. I tend to prefer simple loops.

In this case for example, I think we could group the first reduce + the second map into just one loop, and push the result directly to `transformedGroups` as a result.

But let's not change this here and now, if you want you can look at it in a separate PR.

src/reducers/ComparisonSlice.ts
```diff
@@ -33,6 +43,56 @@ const comparison = createSlice({
...
+        return grouped;
+      },
+      {},
+    );
+
+    const transformedGroupEntries = Object.keys(groupedResults).map(
+      (header_name) => ({
+        [`${header_name} ${truncateHash(
+          groupedResults[header_name][0].new_rev,
+        )}`]: groupedResults[header_name],
```
julienw, line 68: nit: can you please compute the key in a separate line, so that the resulting structure is easier to understand?

src/reducers/ComparisonSlice.ts
```diff
@@ -18,6 +19,15 @@ type Results = {
   revisionHeader: RevisionsHeader;
 };
 
+interface GroupedResults {
+  [key: string]: ResultObject[];
+}
```
julienw, line 24: This is just like `Record<string, ResultObject[]>`, so you could do:

```js
type ResultsGroupedByKey = Record<string, ResultObject[]>
```
And use `ResultsGroupedByKey` everywhere. Also I suggest to change the type name here, so that we understand better what's inside.

julienw (commented):
I left a few more comments, tell me what you think!

src/reducers/ComparisonSlice.ts
```diff
@@ -33,6 +43,56 @@ const comparison = createSlice({
...
+    if (
+      activeComparison ===
+      Strings.components.comparisonRevisionDropdown.allRevisions
+    ) {
+      return JSON.stringify(formatDownloadData(data));
+    } else {
+      return JSON.stringify(
+        formatDownloadData({
+          [activeComparison]: data[activeComparison],
+        }),
```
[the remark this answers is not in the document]
sumairq: I like the idea :100:

src/reducers/ComparisonSlice.ts
sumairq: I agree, it does not look easy to read ,but let's leave it for a future PR.

src/components/CompareResults/beta/DownloadButton.tsx
```diff
@@ -16,78 +19,42 @@ const styles = {
...
   return (
     <div className={styles.downloadButton}>
-      <Button
-        disabled={!results}
-        href={`data:text/json;charset=utf-8,${encodeURIComponent(
-          JSON.stringify(formatDownloadData(results)),
-        )}`}
-        download={fileName}
-      >
+      <Button href={blobUrl ?? ''} download={fileName}>
```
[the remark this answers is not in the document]
julienw: BTW it would be good to test that we actually get a blob url eventually, maybe you'll need to wait until it's rendered (maybe something like `await screen.findByText("Download JSON", { selector: "[href^=blob:]")`).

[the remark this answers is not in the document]
julienw: Still waiting for this :-) thanks

julienw:
hey @sumairq, I'm sorry, but I found an issue with the current way. The problem is that we're always generating the json even though the user won't always click on the download button. This gives a very noticeable lag when the data is big (it's not the only problem when the data is big of course).

I believe we should generate it lazily instead.

Here is a piece of code (I found it on stack overflow, full disclosure :D, but changed it a bit) that seems to work for me locally, but I haven't tried in our context:
```js
const jsonData = ... generate json data here;
const blob = new Blob([jsonData]);
const blobUrl = URL.createObjectURL(blob);             
var element = document.createElement('a');
element.setAttribute('href', blobUrl);
element.setAttribute('download', filename);
element.click();
URL.revokeObjectURL(blobUrl);
```

It's a bit difficult to test, but I guess you could do something such as:
```js
URL.createObjectURL = jest.fn().mockReturnValue('blob:value');
URL.revokeObjectURL = jest.fn();

// click on the button

expect(URL.createObjectURL).toHaveBeenCalled();
expect(URL.revokeObjectURL).toHaveBeenCalled();

delete URL.createObjectURL;
delete URL.revokeObjectURL;
```

julienw:
Maybe it would be easier to start over, I see you imported the commits from the beta branch instead of rebasing...

sumairq:
> Maybe it would be easier to start over, I see you imported the commits from the beta branch instead of rebasing...

Maybe we close this one.

PR #540, opened 2023-09-12
Specify testPathIgnorePattern in the config file instead of package.j…

julienw:
…son, so that we can now limit the running files

This moves the ignore param to the jest configuration.

After this change, it's easy to run only a subset of test files, like this for example:
```
npm run test -- Revisions
```
Previously this wasn't working.

Hopefully this will make it easier to work with the test files.

I checked that the same tests are run before and after this patch.


jest.config.ts
```diff
@@ -6,7 +6,7 @@ const config: Config.InitialOptions = {
   coveragePathIgnorePatterns: ['__tests__', 'index', 'resources', 'styles'],
   setupFiles: ['react-app-polyfill/jsdom'],
   setupFilesAfterEnv: ['<rootDir>/src/__tests__/utils/setupTests.ts'],
-  testMatch: ['<rootDir>/src/**/__tests__/**/*.{ts,tsx}'],
+  testPathIgnorePatterns: ['/node_modules/', '<rootDir>/src/__tests__/utils/'],
```
julienw, line 9: note: `/node_modules/` is the default value. I believe it wasn't used before, so maybe running the tests is slightly faster now.

beatrice-acasandrei (approved):

PR #541, opened 2023-09-13
Update links to the bugzilla on the home page

beatrice-acasandrei:


julienw:
I wonder if we should add a link to our channel on Matrix too

beatrice-acasandrei:
> I wonder if we should add a link to our channel on Matrix too (but this can be in a next patch)

https://mozilla-hub.atlassian.net/browse/PCF-314

PR #543, opened 2023-09-13
Remove RootState where unnecessary

beatrice-acasandrei:


julienw (approved):
Thanks for the cleanup

PR #545, opened 2023-09-14
Add commands to run all tests and fixes

julienw:
This make it easy to check "all the things" locally before pushing to a pull request.

In the future, this command will also be run automatically when pushing, but I keep this for a next PR.

The new commands are:
* `npm run test-all`
* `npm run fix-all`


package.json
```diff
@@ -40,7 +40,8 @@
     "test:coverage": "npm test -- --coverage",
     "test:watch": "npm test -- --watch",
     "test:update": "npm test -- -u",
-    "commit": "npm install npm run format && npm run lint:fix && npm test -- -u && npm run test:coverage",
```
julienw, line 43: I think this line wasn't working (`&&` was missing after `npm install`). So I just removed it. `npm fix-all` should be similar.

beatrice-acasandrei (approved):

PR #552, opened 2023-09-26
Results: Enable "Download JSON" button (Follow up)

sumairq:
## This PR is a follow-up of  [#520](https://github.com/mozilla/perfcompare/pull/520) 

- Implemented a `fileName` selector.
- Used `createSelector()` and `React.memo()` to ensure the state is only recalculated when changed.
- Used `blob` instead of putting all the data inside the URL.
- Blob is generated programatically when the user clicks the "Download JSON" button.

### DIFF
--- a/src/__tests__/CompareResults/ResultsView.test.tsx
+++ b/src/__tests__/CompareResults/ResultsView.test.tsx
@@ -302,4 +302,36 @@ describe('Results View', () => {
     const labelResult = labelFunction({ raw: { x: 5, y: 0, r: 10 } });
     expect(labelResult).toBe('5 ms');
   });
+
+  it('should make blobUrl available when "Download JSON" button is clicked', async () => {
+    const user = userEvent.setup({ delay: null });
+
+    const createObjectURLMock = jest.fn().mockReturnValue('blob:');
+    global.URL.createObjectURL = createObjectURLMock;
+    const revokeObjectURLMock = jest.fn();
+    global.URL.revokeObjectURL = revokeObjectURLMock;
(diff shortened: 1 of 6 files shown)


src/__tests__/CompareResults/ResultsView.test.tsx
```diff
@@ -302,4 +302,36 @@ describe('Results View', () => {
...
+      <ResultsView
+        protocolTheme={protocolTheme}
+        toggleColorMode={toggleColorMode}
+        title={Strings.metaData.pageTitle.results}
+      />,
+    );
+    const button = await screen.findByText('Download JSON');
+    await user.click(button);
+
+    expect(createObjectURLMock).toHaveBeenCalled();
```
julienw, line 335: nit: please check that revokeObjectUrlMock is called with the value returned by createObjectUrlMock.

src/components/CompareResults/DownloadButton.tsx
```diff
@@ -16,78 +20,42 @@ const styles = {
...
-      >
-        Download JSON
-      </Button>
+      <Button onClick={handleDownloadClick}>Download JSON</Button>
     </div>
   );
 }
 
-export default DownloadButton;
+export default React.memo(DownloadButton);
```
julienw, line 61: We probably don't need to memo it anymore! There's no effect, it's now a plain component with some connection to the state.

julienw (approved):
This looks good to me, and this is also much simpler as what we had before IMO!

I just left a couple comments, but once they're fixed we can merge this.

PR #553, opened 2023-10-01
Add accessibility test for results view component and related fixes

sumairq:
## This PR adds test for accessibility in results view component 
 
- Test results view to ensure it does not have accessibility violations.
- Implement fixes to the results view suggested by running the test.
Fixes include
- Add image role for "two overlaping circles" div 
- Add `titleAccess` prop for the grap icon within the link for treeherder graph.
- Use `searchType` prop to ensure the ids are unique for base and new repository dropdowns.


beatrice-acasandrei (approved):

src/components/Search/CompareWithBase.tsx
```diff
@@ -87,6 +87,7 @@ function CompareWithBase({ mode, view }: CompareWithBaseProps) {
           <div className='compare-card-tagline'>{strings.base.tagline}</div>
         </div>
         <div
+          role='img'
```
julienw, line 90: I think i'd use `aria-hidden="true"` for that one, it doesn't provide anything meaningful to the page, as the text besides  already has everything needed.

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -181,7 +181,7 @@ function RevisionRow(props: RevisionRowProps) {
           <div className='graph-link-button-container'>
             <IconButton aria-label='graph link' size='small'>
               <Link href={graphLink} target='_blank'>
-                <TimelineIcon />
+                <TimelineIcon titleAccess='link to treeherder graph' />
```
julienw, line 184: The title should contain information about the image itself, not about the link. I'd be interested to know what the axe error was in this case, can you please explicit? Then I may be able to suggest something different.

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -181,7 +181,7 @@ function RevisionRow(props: RevisionRowProps) {
           <div className='graph-link-button-container'>
             <IconButton aria-label='graph link' size='small'>
               <Link href={graphLink} target='_blank'>
```
julienw, line 183: I realize this now, but it's super weird to have a Link _inside_ a Button. Something to be looked at later...

src/components/CompareResults/RevisionRow.tsx
julienw: I'd suggest to use `aria-label` on the Link instead, as suggested in https://mui.com/material-ui/icons/#semantic-icons
On the Link we can provide a more contextual information, in that case `Go to Treeherder graph` maybe. We can probably remove the aria-label on the IconButton above, maybe even add `role="presentation"` so that it doesn't have any meaning to screen reader users.

julienw:
I'm not sure what you did here @sumairq but the patch seems to include unrelated change. Can you please try to clean it up?

julienw:
Sorry we couldn't get to this. I believe it's obsolete now.
I'm gonna close it but we could come back to it for reference in the future.

PR #554, opened 2023-10-02
Sort frameworks in framework dropdown list

sumairq:
## This PR adds a helper function to sort frameworks
It is a solution for https://bugzilla.mozilla.org/show_bug.cgi?id=1852292
sortedFrameworks

- `sortFrameworks()` function is implemented.
- Frameworks are sorted alphabetically

### DIFF
--- a/src/components/Search/FrameworkDropdown.tsx
+++ b/src/components/Search/FrameworkDropdown.tsx
@@ -22,13 +22,26 @@ import {
 } from '../../styles';
 import type { ThemeMode } from '../../types/state';
 import type { Framework } from '../../types/types';
-import { sortFrameworks } from '../../utils/helpers';
 
 interface FrameworkDropdownProps {
   mode: ThemeMode;
 }
 
 const strings = Strings.components.searchDefault.sharedCollasped.framkework;
+
+const sortFrameworks = (
+  frameworks: Record<Framework['id'], Framework['name']>,
+) => {
+  const unsortedArray = Object.entries(frameworks);
+
+  // Sort the array based on values
+
+  const sortedArray = unsortedArray.sort((a, b) => {
+    return a[1].localeCompare(b[1]);
+  });
+
+  return sortedArray;
+};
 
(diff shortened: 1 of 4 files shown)


julienw:
Would it work to make `frameworkMap` already sorted? In JS objects are sorted, so we could change the order directly here:

https://github.com/mozilla/perfcompare/blob/dc3c81fc5e9e00d04c2611d1559c599d94709b7c/src/common/constants.ts#L28-L39

julienw (changes requested):
see my comment above

beatrice-acasandrei:
> Would it work to make `frameworkMap` already sorted? In JS objects are sorted, so we could change the order directly here:
> 
> https://github.com/mozilla/perfcompare/blob/dc3c81fc5e9e00d04c2611d1559c599d94709b7c/src/common/constants.ts#L28-L39

Hi @julienw! I looked into this, changing the order directly is not working because there is an automatic sorting of numeric keys that is happening when using objects in Javascript (for [context](https://stackoverflow.com/questions/33351816/how-to-prevent-automatic-sort-of-object-numeric-property)). Another solution would be to change the keys to not be numeric anymore or use Map instead of an object. Imo we could accept Sumair's solution if we are not decided, changing the keys or using Map I think would affect other parts of the code.

src/components/Search/FrameworkDropdown.tsx
```diff
@@ -116,7 +117,7 @@ function FrameworkDropdown({ view, mode }: FrameworkDropdownProps) {
           onChange={(e) => void handleFrameworkSelect(e)}
           name='Framework'
         >
-          {Object.entries(frameworkMap).map(([id, name]) => (
+          {sortFrameworks(frameworkMap).map(([id, name]) => (
```
julienw, line 120: can we please compute this outside the component ? It will never change because frameworkMap will never change. It makes sense to keep it local to this file (as opposed to have it in the `constants` file) because it's used here only, but it makes no sense to compute this for every render.

julienw (approved):
thanks

PR #555, opened 2023-10-02
Minimal component for Compare over time

beatrice-acasandrei:
- [x] - Added minimal component for Compare over time but without the timerange, which will be addressed in [PCF-306](https://mozilla-hub.atlassian.net/browse/PCF-306)
- [x] - Changed the **search** state to **searchCompareWithBase** (now it also includes framework)
- [x] - Added new state **searchCompareOverTime** (includes **framework** and **timeRange**)
- [x] - Selecting from the dropdowns will update the state accordingly for fields **repository** and **framework**
- [ ] - WIP: Adding code coverage

Remaining tasks are:
https://mozilla-hub.atlassian.net/browse/PCF-306 
https://mozilla-hub.atlassian.net/browse/PCF-305
https://mozilla-hub.atlassian.net/browse/PCF-307

[PCF-306]: https://mozilla-hub.atlassian.net/browse/PCF-306?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Search/CompareOverTime.tsx
```diff
@@ -1,30 +1,106 @@
...
-        aria-label='a clock'
-      />
-    </div>
+        className={`compare-card-container content-base content-base--${base.class} ${styles.container} `}
+      >
+        <Divider className='divider' />
+        <div ref={formWrapperRef} className='form-wrapper'>
+          <SearchComponentCompareOverTime
+            searchType='new'
+            comparisonType='searchCompareOverTime'
```
esanuandra, line 84: Would be useful to move this `'searchCompareOverTime'` string in a constant?

src/components/Search/SearchDropdown.tsx
```diff
@@ -14,30 +14,41 @@ import {
...
   const size = view == compareView ? 'small' : undefined;
   const { handleChangeDropdown } = useHandleChangeDropdown();
-  const searchState = useAppSelector((state) => state.search[searchType]);
+  let searchState = useAppSelector(
+    (state) => state.searchCompareWithBase[searchType],
+  );
+  if (comparisonType == 'searchCompareOverTime') {
+    searchState = useAppSelector((state) => state.searchCompareOverTime.new);
+  }
   const { repository } = searchState;
```
julienw, line 43: The selector above should return `repository` directly instead of the full `searchState`.

src/components/Search/SearchResultsList.tsx
```diff
@@ -14,7 +14,9 @@ interface SearchResultsListProps {
 
 function SearchResultsList(props: SearchResultsListProps) {
   const { view, mode, searchType } = props;
-  const searchState = useAppSelector((state) => state.search[searchType]);
+  const searchState = useAppSelector(
+    (state) => state.searchCompareWithBase[searchType],
+  );
   const { searchResults } = searchState;
```
julienw, line 20: We need to be more clever. Currently this doesn't work properly, because the "new" entry in "compare with base" share the results with the "compare over time" widget. See for example this:

1. Open the dropdown for the "new" entry in "compare with base". Notice the entries.
2. Open the dropdown for the entry in "compare over time" => same entries
3. Change the repository in one of them, then open the 2 dropdowns again => entries changed in both of them

Instead, we should probably store the results of the fetch in a separate slice, that would be reused between all of the components. The results would be stored with a key for "repository".
Then from the components we can have a selector that would get the fetched value depending on the current selected repository, that should be straightforward.

I believe that it would be good to do this change in a separate PR before doing this one.

cc @Carla-Moz, this is something we discussed yesterday

julienw (commented):
I left a few comments.

I think that as a first step we should change how the "recent revisions" are fetched and stored, because the way it's done now doesn't work properly. Then there are a few other comments that could be done in some separate PR first too (eg moving some hooks back into their callers).

Finally I believe we need to stop passing a string indicating some view, but instead provide more "configuration" props if you will. In the end the component will be dumber and therefore easier to work with. We can change the existing code later, but at least it would be good to do it for the new code (in this case, it would be good to avoid the `searchType` prop).

Also if we want to land this, it would be good to have a flag so that the compare over time widget isn't rendered currently, using `useSearchParams` possibly.

Tell me what you think!

Carla-Moz:
Hi @beatrice-acasandrei, I'm doing a major refactor of CompareOverBase, the hooks, and reducers, so I think it might be better if you let me take over this work and continue from what you have. What do you think? Unless you really want to work on this. I agree with @julienw to pause this work until we improve the architecture.

Carla-Moz:
Closing this PR since a lot of what has been discussed has been implemented and it's now easier to create this comp.

PR #559, opened 2023-10-12
Compare Base Refactor 

Carla-Moz:
This PR refactors Compare With Base components and its children and closes [Refactor CompareWithBase Component](https://mozilla-hub.atlassian.net/browse/PCF-328?atlOrigin=eyJpIjoiYjVjZWEyNDE2MTcyNDUyNGEzNDdkMWNmZDZmNjI4M2UiLCJwIjoiaiJ) and fulfills a part of [Remove "if" conditions inside components by making them more configurable](https://mozilla-hub.atlassian.net/browse/PCF-336)

TLDR: Major changes include removing unecessary useEffects, useStates and improving the edit feature in the compare view. 

Key components: `CompareWithBase.tsx and SearchComponent.tsx,`

1.` CompareWithBase`
- [x] removed unnecessary `useEffect` 
- [x] Moved handling of the repository error message down to the `SearchComponent`
- [x] Removed the passing down of the `view` prop 

2.` SearchComponent` 
- [x] handled view based on url
- [x] extracted the edit and save/cancel buttons into their own components
- [x] implemented `createRef` to help with conditional rendering of the edit feature buttons based on the view and user actions


src/components/Search/SaveCancelButtons.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
```
julienw, line 1: Good idea to extract the buttons here!

src/components/Search/SearchComponent.tsx
```diff
@@ -87,33 +71,29 @@ function SearchComponent({
...
-      $nest: {
-        '&.MuiButtonBase-root': {
-          ...btnStyles.Secondary,
-        },
-      },
-    }),
-  };
 
-  const dispatch = useAppDispatch();
+  const { search } = useAppSelector((state) => state);
```
julienw, line 75: Selectors shouldn't return the full state, because this makes your component rerender each time any part of the state changes. Instead you should return the smallest part of the state you need, and even do the computation that returns exactly what you need inside the selector. It's OK to have several `useAppSelector` calls.

src/components/Search/SearchComponent.tsx
```diff
@@ -87,33 +71,29 @@ function SearchComponent({
...
-  const [editBtnVisible, toggleEditBtnVisible] = useState(true);
+  const location = useLocation();
+  const view = location.pathname == '/' ? searchView : compareView;
   const matchesQuery = useMediaQuery('(max-width:768px)');
   const containerRef = createRef<HTMLDivElement>();
+  const editButtonRef = createRef<HTMLButtonElement>();
+  const selectedRevisionsRef = createRef<HTMLDivElement>();
+  const isWarning =
+    (baseRepository === 'try' && newRepository !== 'try') ||
+    (baseRepository !== 'try' && newRepository === 'try');
```
julienw, line 96: It's not super clear to me that this is the right place, I wonder if `CompareWithBase` was better. Indeed `SearchComponent` is either `base` or `new`, so doing this check here doesn't feel right to me.

julienw (commented):
Thanks for this work Carla!
However I believe it's not going to the right direction.

Editing the class list directly using DOM doesn't sound a good idea. Indeed, with React editing DOM should be (as much as possible) a pure function with the state and props as inputs. You need to think: what variations are there in this component? Should this be a local state or a prop?

In this case, the SearchComponent could use a local state, and pass it down to its children components. It could also pass a callback that would change this local state.

The first thing to do is to define the various variations and states. IMO we have
1. Form is visible / not visible (when not visible the same information is displayed with an edit button) -- The initial value for this state would depend on a prop such as `isEditable=boolean` because it's different depending on where this component is used. -- The prop is passed through `CompareWithBase` too, it comes from `SearchContainer` and `ResultsView`.

* when isEditable=false: the form is visible as initial state (`const [formIsDisplayed, setFormIsDisplayed] = useState(!isEditable)`), the save/cancel buttons are not rendered. This state will never be changed as a result, because no button can change the `formIsDisplayed` state.
* when isEditable=true: the form isn't visible as the intitial state. An edit button is rendered if `isEditable && !formIsDisplayed`. Clicking the edit button toggles the state `formIsDisplayed`.
* The form is displayed if `formIsDisplayed` is true. The buttons are displayed if `formIsDisplayed && isEditable`. Clicking any of the buttons sets `formIsDisplayed` back to false. To achieve that, `SaveCancelButtons` should take callbacks as prop: `onSave` and `onCancel`, so that its parent can react on those. If the parent holds the state for the currently checked revisions (like explained below), then  the parent could also save the checked revisions, and `SaveCancelButtons` would be much simpler.
* when isEditable=true and form is not visible, the edit button is present. When pressed, the edit button toggles the `formIsDisplayed` state.


2. Which revisions are displayed -- the initial value for this state is passed as a prop because it's different depending on where this component is used (empty on the home page, selected revs in the search results page). Same thing as for 1: this comes from `SearchContainer` and `ResultsView`.
  Note that currently the state for this thing is stored in Redux, but I believe this makes things complicated, because we need to update it at the first render. It would be better to move this state to the component. But this should be done in a separate PR as a preparatory step, because it's not straightforward, with many small things here and there to be changed: especially `SearchListResultsItem` will need to pass up the click information to `SearchListResults` (for example, with a `onToggle` callback: `this.props.onToggle(searchType, item.revision, !isChecked)`) and then up to `SearchComponent`.


 
I'd be more than happy to do some pair programming with you tomorrow to help with this. Today unfortunately I can't but hopefully these explanations can give you some head start.

We also still have `if` conditions that compares the view, but instead we should have props to configure the component. But happy to look at this separate problem in a separate PR.
For example: maximum number of revisions that can be selected -- this depends on whether it's "new" or "base" and can be a prop passed from `CompareWithBase`.


Hope this helps, sorry this applies on most of the work here, but I believe using these ref doesn't improve the architecture.

Carla-Moz:
@julienw My initial approach was to manage the rendering of the edit buttons through SearchComponent's state and passing down props to the children. However, I was so fixated on bypassing conditional rendering and not using useState that I ended up using bad practice with DOM manipulation. Next time, I'll be more cautious about how I approach my fixes and read the docs more carefully to make sure my patches adhere to good practices, esp in React. Thanks for this.

beatrice-acasandrei (approved):

src/components/Search/SelectedRevisionItem.tsx
```diff
@@ -65,6 +67,13 @@ function SelectedRevisionItem({
   const selectedFramework = useAppSelector((state) => state.framework);
   const navigate = useNavigate();
   const prevRevRef = React.useRef<RevisionsList[]>([]);
+  const location = useLocation();
+  const view = location.pathname == '/' ? searchView : compareView;
+  //hide the close icon for the selected revisions in compare view
+  const iconClassName =
+    !formIsDisplayed && view !== searchView
```
julienw, line 74: We'll have to think a bit about how to remove this remaning condition using the view, I'm sure we can come up with something smart.

src/components/Search/SelectedRevisions.tsx
```diff
@@ -2,71 +2,72 @@ import { useEffect, useState } from 'react';
...
     const selectedRep = repoMap[item.repository_id];
     return selectedRep;
   });
+  const compareViewSelectedRevisions = useAppSelector(
+    (state) => state.selectedRevisions[searchType],
+  );
 
   useEffect(() => {
-    if (view === searchView || !editBtnVisible) {
+    if (view === searchView) {
```
julienw, line 57: This one too, but I have an idea for this one: the revisions and repositories should be passed as props.

julienw (commented):
This feels so much cleaner ! Thanks so much !

I've noticed 2 additional places we might want to improve, but nothing urgent here!

PR #561, opened 2023-10-17
Move thunk actions to the reducer files using them

julienw:
The goal for this PR is to simplify our file tree by merging the thunk actions to the reducer files that use them. This makes it easier to reason about them, because that's in one file.

While doing this I found the cause for [Bugzilla bug 1852287](https://bugzilla.mozilla.org/1852287) so I fixed this as well. Otherwise this is only some cleanup in preparation for changing how the search results for the search panel work.

I didn't touch the tests because they still work the same.

I recommend that you can look at the individual commits separately:
* commit 1: simple cleanup to use the createStore function when we create the store singleton
* commit 2: move the thunk actions for the comparison results
* commit 3: fix the issue I mentioned above
* commit 4: move the thunk actions for the Search panel
* commit 5: simple cleanup: rename the file for the CompareResults slice because it was the only one without "slice" in its name

I believe that there is no conflict with your work @Carla-Moz.


Carla-Moz (changes requested):

julienw:
Closing, this is deeply outdated and I believe we did some of this in other PRs, and some upcoming ones will do this more too.

PR #563, opened 2023-10-24
Remove a few hooks

beatrice-acasandrei:


src/hooks/useQueryParams.ts
```diff
@@ -1,15 +0,0 @@
-import React from 'react';
-
-import { useLocation } from 'react-router-dom';
-
-export function useQuery() {
-  const { search } = useLocation();
-
-  return React.useMemo(() => new URLSearchParams(search), [search]);
-}
```
julienw, line 9: It's worth mentioning that this was in React Router already :) https://reactrouter.com/en/main/hooks/use-search-params

julienw (approved):
Thanks!

PR #569, opened 2023-11-07
Results: (Bug Fix) In RevisionRow handle case when the platform icon is not present

sumairq:
### Bug: 
Selecting a testing framework other than talos along with selecting mozilla central repo for comparison breaks the app.

- In `RevisionRow` component we use `<plateformIcon/>` element to show the platform icon. On rare ocasion the icon is missing or unspecified however an invalid react element is still generated, breaking the app

### Solution:
Used a ternary operator to check if the `platformInfo.shortName` is present as this is a reliable value to check if the platform icon is  going to be a valid react element or not.
`{platformInfo.shortName ? <PlatformIcon /> : 'Unspecified'}`

If the platform icon is not present we handle it by adding "Unspecified" string. 
solution

**Julien's update:** I changed a few things to the original patch:
* I moved `getPlatformInfo` from `helpers.js` to `RevisionRow`
* I moved the logic about unspecified plaform to `getPlatformInfo`, and changed how we return the icon info from Components to Elements. This made the code in `RevisionRow` simpler.
* I created a new test file for `RevisionRow` to test for this behavior because I removed some old test file.
  * Probably some of the existing tests we have in `ResultsView` could move to this more targeted file later (I'm thinking about the tests about expanding rows) to make them faster.

### DIFF
--- a/src/components/CompareResults/RevisionRow.tsx
+++ b/src/components/CompareResults/RevisionRow.tsx
@@ -1,5 +1,6 @@
 import { useState } from 'react';
 
+import AppleIcon from '@mui/icons-material/Apple';
 import ExpandLessIcon from '@mui/icons-material/ExpandLess';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
 import FileDownloadOutlinedIcon from '@mui/icons-material/FileDownloadOutlined';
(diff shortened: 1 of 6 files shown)


src/components/CompareResults/RevisionRow.tsx
```diff
@@ -138,7 +138,7 @@ function RevisionRow(props: RevisionRowProps) {
       <TableRow className={`revisionRow ${styles.revisionRow}`}>
         <TableCell className='platform'>
           <div className='platform-container'>
-            <PlatformIcon />
+            {platformInfo.shortName ? <PlatformIcon /> : 'Unspecified'}
```
julienw, line 141: This looks good to me but the string should go to te Strings file please :-)

esanuandra (approved):

PR #571, opened 2023-11-13
Move dark mode to redux

beatrice-acasandrei:


src/reducers/Theme.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const theme = createSlice({
+  name: 'theme',
+  initialState,
+  reducers: {
+    updateThemeMode(
+      state,
+      action: PayloadAction<{
+        mode: string;
+      }>,
```
julienw, line 15: You can use a simpler payload:
```suggestion
      action: PayloadAction<string>,
```
and then
```js
    dispatch(
      updateThemeMode(storedMode === 'light' ? 'dark' : 'light'),
    );
```

julienw (commented):
This is a good start, but then it would be good to use the stored `mode` everywhere we pass the prop (instead of using the prop).

Carla-Moz (changes requested):
Thanks for all this work! However, the theme mode setting isn't preserved when the page is reloaded or starts a new session. Do we want users to always toggle the dark mode button or do we want to save their preferred settings? Or do we want to set this aside until we authentication? That way we can save users settings for theme mode?

julienw:
I think you could:
1. Get local storage to define the initialState of the dark mode slice
2. Set local storage in the component that dispatches the action

Because there's just one place that changes the mode, this is simple and good enough IMO.

src/reducers/Theme.ts
julienw, line 1: Please name the file ThemeSlice for consistency

julienw (commented):
The code looks OK to me, but I can see that the linter and the tests are failing.
For the tests that use `renderHook`, you should probably add the `wrapper` option to specify a Provider with a store.
In general we shouldn't use `renderHook` but rather test components directly. But let's not change this here.

src/theme/protocolTheme.ts
```diff
@@ -61,22 +63,24 @@ const getDesignTokens = (modeVal: ThemeMode) => ({
 });
 
 const useProtocolTheme = () => {
```
beatrice-acasandrei, line 65: @julienw because I am using `useAppDispatch()` here I am getting the following error
Screenshot 2023-11-20 at 15 37 10
Any idea on how to fix this?

julienw: I wrote what I think is the solution in my comment above: you need to specify the `<Provider>` to `renderHook` using the `wrapper` argument in the `renderHook` cool. You'll also need to create a store. Maybe it's good to create a shared function `renderHook` in `src/__tests__/utils/test-utils.tsx` like we have for `render` already?

PR #572, opened 2023-11-13
Use the new react router api to manage the URL and fetching external data

julienw:
Not finished yet, but I put this up so that we can look at the deploy preview.
It should be better/easier to look at the commits one by one.
I'll probably extract some of the commits (the first ones) to separate PRs to make it easier to review.

Done:
* [X] The compare results page use the URL parameters.
* [X] The Search form feeds the URL parameters.
* [X] The results shown in the Search dropdown come from the router too
* [X] Typescript passes.
* [X] The Search form itself isn't itself primed with the URL parameters (in the compare results page).
 
Not done or known bugs:
* [ ] Edit mode for the search form in the compare results page doesn't work.
* [ ] No clean up has been done on the existing redux store that we won't use anymore.
* [ ] No test has been updated 
* [ ] Handle errors gracefully
* [ ] "Loading" indicator in the CompareResults page (probably in another PR)
* [ ] Better behavior when clicking first in the search input -- currently there's a delay due to the fact that the request is sent on focus, while before it was primed before focus. We'll probably want some "optimistic design" with grey boxes while it's loading. Or find a way to prime the data in a better way than we used to do.
* [ ] Move from a HashRouter to a BrowserRouter (probably in another PR)

### DIFF
--- a/src/__tests__/hooks/useHandleChangeSearch.test.ts
+++ b/src/__tests__/hooks/useHandleChangeSearch.test.ts
@@ -32,9 +32,12 @@ describe('Tests useHandleSearchHook', () => {
       searchType,
       repository: 'try' as Repository['name'],
     };
-    const { result } = renderHook(() => useHandleChangeSearch(), {
-      wrapper: StoreProvider,
-    });
+    const { result } = renderHook(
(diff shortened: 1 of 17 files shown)


julienw:
This work is done.

PR #573, opened 2023-11-17
fix for close buttons in edit mode

Carla-Moz:
This  PR fulfills [Results: In edit mode, close buttons for selected revisions are broken](https://mozilla-hub.atlassian.net/browse/PCF-344). 

Note that the save and cancel bugs in edit mode will be taken care of in a separate PR. 

Also, I removed some fetch code in the Results View because I initially placed it in a separate, now deleted file but that will be taken care of in Julien's patch. I'll be glad to return it though.

### DIFF
--- a/src/hooks/useSelectedRevisions.ts
+++ b/src/hooks/useSelectedRevisions.ts
@@ -18,10 +18,13 @@ const useSelectRevision = () => {
     (state) => state.search.new.checkedRevisions,
   );
 
+  const selectedRevisionsState = useAppSelector(
+    (state) => state.selectedRevisions,
+  );
+
   const selectedRevisions = useAppSelector(
     (state) => state.selectedRevisions.revisions,
   );
-
   const addSelectedRevisions = () => {
     const newSelected = [...selectedRevisions];
     newSelected.push(...baseCheckedRevisions, ...newCheckedRevisions);
@@ -35,10 +38,11 @@ const useSelectRevision = () => {
     let revisionsForEdit = selectedRevisions;
     switch (searchType) {
       case 'base':
-        revisionsForEdit = [revisionsForEdit[0]];
+        revisionsForEdit = selectedRevisionsState.base;
+
         break;
       case 'new':
-        revisionsForEdit = revisionsForEdit.slice(1);
+        revisionsForEdit = selectedRevisionsState.new;
         break;
       default:
         throw new Error('Invalid search type');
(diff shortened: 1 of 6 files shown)


src/reducers/SelectedRevisionsSlice.ts
```diff
@@ -28,8 +28,16 @@ const selectedRevisions = createSlice({
       state,
       action: PayloadAction<{
         selectedRevisions: RevisionsList[];
+        isBaseDeletion: boolean;
       }>,
     ) {
+      if (action.payload.isBaseDeletion) {
+        return {
+          ...state,
+          revisions: action.payload.selectedRevisions.slice(1),
```
julienw, line 37: I think it's more correct to slice in the dispatch call rather than here. So that this code block looks more similar to the code below for deleting new revisions. What do you think?

src/hooks/useSelectedRevisions.ts
```diff
@@ -72,8 +76,20 @@ const useSelectRevision = () => {
 
   const deleteSelectedRevisions = (revision: RevisionsList) => {
     const newSelected = [...selectedRevisions];
+    if (newSelected[0].id === revision.id) {
+      //this is a base revision deletion
+      dispatch(
+        deleteRevision({
+          selectedRevisions: newSelected,
```
julienw, line 83: see below: I think it's more correct to call `slice` here. The code is then more similar to the code for deleting a new revision: change the selected revisions first here, and dispatch the resulting array then.

src/hooks/useSelectedRevisions.ts
```diff
@@ -72,8 +76,20 @@ const useSelectRevision = () => {
 
   const deleteSelectedRevisions = (revision: RevisionsList) => {
     const newSelected = [...selectedRevisions];
```
julienw, line 78: nit: this clone can happen after the if block: we need it for the `splice` operation, but not the `slice` operation (slice will always return a new array, while `splice` modifies the array).

julienw (approved):
This looks good to me, except the simple change I'm asking for.

In my opinion this comes from the fact that we currentmy store base and new revisions in the same array (actually we duplicate this value to base and new as well), but my upcoming patch should change this and we will have a cleaner state to work with as a result, hopefully.

src/reducers/SelectedRevisionsSlice.ts
Carla-Moz: Yes, we do store both in the same array and it's not great. I'm glad that will get simpler with your patch. Yeah, I'll slice in the dispatch. And when you land your patch, we can update accordingly if needed.

src/hooks/useSelectedRevisions.ts
Carla-Moz: Yes, agree, best to keep things consistent.

PR #575, opened 2023-11-21
Simplify the useHandleChangeSearch hook

julienw:
This contains 2 small commits, that are fairly independant so you might want to look at them separately. 

Commit 1:
1. Merges `handleFetch` and `searchByRevisionOrEmail` because one of them was just calling the other.
2. Removes the state that used to store the `searchType` (`new` or `base`). The value is already passed directly to the function that needs it so this was redundant. But also this was used to get the current repository for this `searchType`: now we pass this information directly to the function.
3. It also removes a useless/not working `isError` return value.

Commit 2 renames a function so that it reflects better what it does (in my opinion).

There should be no behavior change, so I didn't touch the tests except where needed.

Tell me what you think!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "react-chartjs-2": "^5.2.0",
         "react-dom": "^18.2.0",
         "react-redux": "^8.0.2",
-        "react-router-dom": "^6.3.0",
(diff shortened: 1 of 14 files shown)


Carla-Moz (approved):
Thanks for cleaning up this hook!

PR #576, opened 2023-11-21
Change a bit how the recent revisions dropdown handling works

julienw:
This contains 2 commits that can be looked at independently:

Commit 1:
* This changes the `mousedown` event handler that controls the hiding of the recent revisions dropdown when clicking elsewhere. Previously it also contained some code to display it, but that wasn't used... It also simplifies the check that the user clicks elsewhere.
* This simplifies the way that the search input notifies its parent about the focus event: instead of passing the `setFocused` function coming from `useState`, we just pass a callback that will be called when focus happens.

Commit 2: changes the state name about the fact that the dropdown is displayed or not, to better reflect that. For example we could have the input focused, press Escape, and then the dropdown would be hidden... but the focus would still be there. So the state was lying before.

There should be no behavior change again. So tell me what you think :-)

### DIFF
--- a/src/components/Search/SearchComponent.tsx
+++ b/src/components/Search/SearchComponent.tsx
@@ -92,7 +92,7 @@ function SearchComponent({
 
   const searchState = useAppSelector((state) => state.search[searchType]);
   const { searchResults } = searchState;
-  const [focused, setFocused] = useState(false);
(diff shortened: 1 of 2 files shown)


Carla-Moz (approved):
Tested this out and looks good! Thanks for the update and covering the displayDropdown case in Results view.

PR #577, opened 2023-11-21
Simplify how we pass the props to CompareWithBase

julienw:
This flats out the objects we used to pass some of the props to CompareWithBase. Indeed I realized that we were using useMemo to change the objects shapes, and all that. Instead I found it was easier to pass individual props.
Tell me what you think :-)


Carla-Moz (approved):
This looks way simpler and cleaner than before. Thanks for this. Just get the tests green and we're all good!

julienw:
One line in one file isn't covered, but this was there before the patch. The problem is that there's no test where the some "new" revisions are checked, this should probably be added to the tests for `SearchView`. I'll see if I can do this easily.

julienw:
I improved one of the existing tests and how the patch is covered completely \o/

PR #579, opened 2023-11-28
WIP: Save and Cancel Buttons in edit mode 

Carla-Moz:


src/reducers/SelectedRevisionsSlice.ts
```diff
@@ -23,26 +24,37 @@ const selectedRevisions = createSlice({
       state.base = [action.payload.selectedRevisions[0]];
       //returns array without first element
       state.new = action.payload.selectedRevisions.slice(1);
+      state.editModeRevisions = action.payload.selectedRevisions;
     },
-    deleteRevision(
+    saveUpdatedRevisions(
```
julienw, line 29: I think it would help the reader if each action had a comment explaining when it's called (for example: "when the user press the save button when the search form is in edit mode").

src/reducers/SelectedRevisionsSlice.ts
```diff
@@ -23,26 +24,37 @@ const selectedRevisions = createSlice({
...
+      //returns array without first element
+      state.new = action.payload.selectedRevisions.slice(1);
+      state.editModeRevisions = action.payload.selectedRevisions;
+    },
+    updateEditModeRevisions(
       state,
       action: PayloadAction<{
         selectedRevisions: RevisionsList[];
         isBaseDeletion: boolean;
+        isAddChecked: boolean;
```
julienw, line 45: having comments explaining what these 2 booleans are for would be useful.

My understanding:
* isBaseDeletion => the user deleted the base rev
* isAddChecked => I honestly don't know :D

src/reducers/SelectedRevisionsSlice.ts
```diff
@@ -23,26 +24,37 @@ const selectedRevisions = createSlice({
...
-        };
+        state.base = [];
+        state.editModeRevisions = action.payload.selectedRevisions;
+      }
+
+      if (action.payload.isAddChecked || !action.payload.isBaseDeletion) {
+        state.base = [action.payload.selectedRevisions[0]];
+        state.new = action.payload.selectedRevisions.slice(1);
+        state.editModeRevisions = action.payload.selectedRevisions;
       }
```
julienw, line 57: I don't understand well these 2 blocks.

From the rest of the code, I think these 3 cases are valid:
1. isBaseDeletion: true, isAddChecked: false
2. isBaseDeletion: false, isAddChecked: false
3. isBaseDeletion: false, isAddChecked: true

This case does not exist:
* isBaseDeletion: true, isAddChecked: true

Now if I do the bool operation on the 2 conditions in this reducer, here is the result for each of the 3 valid cases:

1. first condition: true, second condition: false
2. first condition: false, second condition: true
3. first condition: false, second condition: true

From this analysis, it looks like the second block is basically a `else` of the first condition, and that `isAddChecked` doesn't have any usefulness. Therefore this could be rewritten like this:

```js
      state.editModeRevisions = action.payload.selectedRevisions;

      if (action.payload.isBaseDeletion) {
        state.base = [];
      } else {
        state.base = [action.payload.selectedRevisions[0]];
        state.new = action.payload.selectedRevisions.slice(1);
      }
```
An alternative way would use an early return and remove the else.
Am I missing something?

I think I'd prefer 2 separate reducers instead of this boolean BTW. But I'm not sure how practical this is in the caller, so this is optional really.

julienw (commented):
I haven't finished the review but I think these initial comments can be useful already and you might be able to work on them, so I'm pushing them. I'll finish tomorrow.

src/reducers/SelectedRevisionsSlice.ts
Carla-Moz: Good idea.

src/reducers/SelectedRevisionsSlice.ts
Carla-Moz: I created 2 separate reducers for each action (deleting a revision and adding a revision when checked) to make this cleaner and easier.

julienw:
closing because it's duplicated by https://github.com/mozilla/perfcompare/pull/581 :-)

PR #580, opened 2023-11-28
PCF-346 Performance: layout is very slow with a lot of results because we're using a gigantic table

esanuandra:
This PR addresses [PCF-346](https://mozilla-hub.atlassian.net/browse/PCF-346)
It changes the structure from material ui table to grid layout.
Screenshot 2023-12-05 at 2 23 32 PM

[PCF-346]: https://mozilla-hub.atlassian.net/browse/PCF-346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/CompareResults/RevisionRow.tsx
```diff
@@ -66,130 +66,135 @@ function RevisionRow(props: RevisionRowProps) {
 
   const styles = {
     revisionRow: style({
+      display: 'grid',
+      gridTemplateColumns: '2fr 1fr 0.2fr 1fr 1fr 1fr 1fr 1fr 2fr 0.2fr',
```
julienw, line 70: can you please add a comment here to mention this should be kept in sync with the other gridTemplateColumns?

Ideally it would be great to share the CSS style, but I'm not sure how to do this with this CSS-in-JS library, so I'm fine with a comment for now.
Please add a comment in the other location too.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -30,45 +28,44 @@ function ResultsTable(props: ResultsTableProps) {
...
+        <div>
           <TableHeader themeMode={themeMode} />
           {processedResults.map((res, index) => (
-            <TableContent
-              themeMode={themeMode}
-              key={index}
-              header={res.revisionHeader}
-              results={res.value}
-            />
+            <div key={index}>
```
julienw, line 53: Is this div useful? I see that the TableContent component itself also has a div already as its root element.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -30,45 +28,44 @@ function ResultsTable(props: ResultsTableProps) {
...
+              <TableContent
+                themeMode={themeMode}
+                key={index}
```
julienw, line 56: you don't need the `key` here if you keep the `div` above.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -30,45 +28,44 @@ function ResultsTable(props: ResultsTableProps) {
...
       sx={customStyles}
+      role='table'
     >
       {loading ? (
         <Box display='flex' justifyContent='center' alignItems='center'>
           <CircularProgress />
         </Box>
       ) : (
-        <Table>
+        <div>
```
julienw, line 50: Is this div useful? The `<Paper>` above already adds a container

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -30,45 +28,44 @@ function ResultsTable(props: ResultsTableProps) {
   const styles = {
     tableContainer: style({
       backgroundColor: themeColor100,
+      borderCollapse: 'separate',
+      borderSpacing: `0 ${Spacing.Small}px`,
```
julienw, line 32: These 2 borderCollapse and borderSpacing are probably not used anymore

julienw (approved):
Thanks, this works very well! Great work!

I see there's a small visual difference in the space between rows, but I do actually prefer the new spacing ;-) It would be good to get @Carla-Moz's opinion on this too.

I also looked at the performance comparison with the beta branch on some random revisions on mozilla-central with the browsertime framework (usually producing large tables):
* initial rendering on some page was ~20 sec compared to ~47 sec on the beta branch. Still a long time, but this is already a significant improvement and there are more things to fix later (for example https://bugzilla.mozilla.org/show_bug.cgi?id=1867070, or my upcoming patch with react-render preventing some rerenders). I expect that the longer the page, the better the improvements, because I suspect that with the table we used to have an exponential render time.
* expanding a row is much much much much faster (instant, while with the previous table implementation it would take a few seconds in a long results page)

As a follow-up (not this PR), can you please add aria roles for tables / rows / rowgroup etc? See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Table_Role as a documentation. I see you added `table` role but not the others.

Thanks again

Carla-Moz (approved):
This is awesome. I think the new spacing between rows work well. Thank you!

src/components/CompareResults/ResultsTable.tsx
esanuandra: This `div` is useful because there is `TableHeader` component and the processedResults with `TableContent`. These components are shown based on the `loading` value.

julienw: But you could use a fragment instead, maybe?

```js
<>
  <TableHeader />
  {processedResults...}
</>
```

julienw: As a reminder, a fragment groups other components together when a single root is needed (like here), but otherwise doesn't render any DOM node to the page.
You might also have seen it as `<Fragment>` or `<React.Fragment>`, but `<>` is shorter.

PR #581, opened 2023-12-05
DRAFT: Updated save cancel

Carla-Moz:
The purpose of the Cancel button is to reset either the base or new revisions to their default selections pre and/or post any deletions or additions to the respective containers. Upon pressing cancel, the user must see the default selected revisions for said container.

TheSave button's purpose is to confirm any changes the user has made to the base or new revisions. It must show the user these changes. It must not, however, update the URL or change the results. This functionality must be reserved for the Compare button.

The following tickets cover the issues above which this PR will resolve:
[Cancel Btn PCF-356](https://mozilla-hub.atlassian.net/browse/PCF-356)
[Save Btn PCF-357](https://mozilla-hub.atlassian.net/browse/PCF-357)
[Compare button should apply the edits in results view PCF-310](https://mozilla-hub.atlassian.net/browse/PCF-310)

To fix these bugs, I added a temporary editModeRevisions to the selectedRevisions state; this array might change (re either deleted entirely or updated) post Julien's major patch. But for now, it will hold the updates the user has made in edit mode vs the default revisions.

On clicking Compare, editModeRevisions will replace the default revisions array, changing the url and results.


Carla-Moz:
@julienw I created a new branch since I was having git issues with my others. However, I did create the new aliases and will be sure to keep my history clean next time. This PR resolves you comments from the last one: https://github.com/mozilla/perfcompare/pull/581

Carla-Moz:
> I'm all for moving the 2 buckets "in progress" and "staging" out of Redux and inside the Search Component. Not necessarily now, but this sounds like a good moment, what do you think?

Yes, I don't mind doing so in this PR. The sooner the better. I like the idea of having the three different states because that's exactly what's going on here.

Carla-Moz:
> One example that come directly from the fact that we don't handle the 3 buckets properly: STR 1:
> 
>     1. Press Edit.
> 
>     2. Check another revision than the ones you currently have.
> 
>     3. Press save. (we should have 2 or 3 revisions depending on the initial count)
> 
>     4. Press Edit again.
> 
>     5. Press cancel.
> 
>     6. => notice that we're back at the state in 1, not the state in 3.

Yes, we go back to state 1 because the `Save` button doesn't set the "Commit" state in this scenario, it only sets the "In Progress" state which is erased upon hitting the `Cancel` button. The `Compare `button sets the "Commit" state. However, I do see how the above UX would confuse the user. Having the "Staging" state will correct that confusion. Therefore, the `Edit`  button will set the "In Progress" state, the `Save` button will set the "In Staging" state of the values so that the user can go back to state 3 instead of 1 should they hit the `Cancel` button. I'll map out the cases in the update of this PR to help with the next review.

Carla-Moz:
@julienw What do you think. I want to make sure I understand your proposal before I try implementing it. 

- We have 3 states: "In Progress", "Staging", "Committed". 
- The `CompareWithBase` Comp will handle the "In Progress" and "Staging" states
- `Redux` will handle the "Committed" state
- The `Edit` button sets the "In Progress" state
- The `Save` button sets the "Staging" state
- The `Cancel` button will return state to "Staging" post the user hitting the `Save` button; otherwise, it will return state to "In Progress"
- The `Compare` button will set the "Committed" state.

PR #582, opened 2023-12-14
Use a loader with the new react-router API to handle the compare results page

julienw:
**Post rebase comment:** I think this doesn't work well with the edit mode yet, but I want to look at this after this PR lands.

I believe it's now ready for a review.
I recommend to look at it commit by commit, because it's easier to understand the logic of how this works by looking it piece by piece.

Here is the general idea:
* All fetches to the treeherder API are done through React Router. When we define a route, we also define a [loader function](https://reactrouter.com/en/main/route/loader).
* This loader is responsible for fetching the data. All input values are in the URL.
* This data is then available to components using the hook [`useLoaderData`](https://reactrouter.com/en/main/hooks/use-loader-data).
* The [`<Form>`](https://reactrouter.com/en/main/components/form) component from ReactRouter is responsible for navigating and updating the URL. It works the same in both the SearchView and the ResultsView so the URL will be navigating the same way.

This fixes the following issues:
* The URL is now the authorative source to render a page. This means that we can share it easily. (note that it won't have the filtering options we have in other parts of the page, this is something we could discuss later, but I'm not sure for perf reasons)
* There's just one render, no more effects that will trigger rerenders.
* The flow of rendering a page is simpler.

And the following JIRA tickets:
[PCF-337 Store data fetches in a separate place in Redux](https://mozilla-hub.atlassian.net/browse/PCF-337)
[PCF-113 When pasting a hash to one of the inputs, I'm directly moved to the result page](https://mozilla-hub.atlassian.net/browse/PCF-313)
[PCF-324 URLs are often inconsistent with the current state](https://mozilla-hub.atlassian.net/browse/PCF-324)
[PCF-385 Refactor how the results view is rendered](https://mozilla-hub.atlassian.net/browse/PCF-385)

And helps with these in coordination with #595 :
[PCF-318 In Results view, on edit, the results get appended to the old results instead of replacing them](https://mozilla-hub.atlassian.net/browse/PCF-318)
[PCF-310 Compare button should apply the edits in results view](https://mozilla-hub.atlassian.net/browse/PCF-310)

Here are the known issues that are present after this patch, I think they're all fixable:
1. ReactRouter waits for the result of the `loader` (when the promise resolves) before navigating. This means that when clicking on "Compare", there's a visible waiting time without any user feedback before we end up on the next page. This should be fixable with the [Deferred Data](https://reactrouter.com/en/main/guides/deferred) concept in react router.
The visible wait is increased by the react rendering time, but that's another issue.
2. When there is an error, the page is still the default page for React Router, we could make a better one.
3. (previously existing issue) We use a HashRouter where we could use a BrowserRouter which would look more natural from a user point of view.
4. Edit mode doesn't work on the Search form.

(you might see others?)

I didn't try to support the old URLs, so we'll need to update existing references. But I'd like to change the HashRouter to a BrowserRouter first (probably I'll do that first after this PR).

I rewrote a lot of the tests in the "Fix tests" commit. I can split this one in smaller commits if you prefer.
For this specific commit, I tried to follow the guidelines from this blog post: https://kentcdodds.com/blog/write-fewer-longer-tests. So I merged a bunch of existing testcases into just one, trying to follow what a typical user would do. I also removed some tests about the internal redux state following this blog post: https://kentcdodds.com/blog/testing-implementation-details. Hopefully this will make our tests faster, easier to write and read, and less prone to break when changing some code.

Please tell me what you think!


src/__tests__/utils/test-utils.tsx
```diff
@@ -37,18 +37,22 @@ export function renderWithRouter(
...
+  // because this is a navigation. Instead we use `replaceState` that isn't a
+  // navigation and works also for us.
+  window.history.replaceState(null, '', route + search);
   return {
     ...render(
-      <Router location={history.location} navigator={history}>
-        {ui}
-      </Router>,
+      <RouterProvider
+        router={createBrowserRouter([{ path: route, element: ui, loader }])}
```
julienw, line 52: Note: I decided to use a browser router for tests because then it's easy to use window.location to check for the current location.

I also want to use a browser router for the app too later, but didn't want to implement it in this patch just yet.

src/components/Search/SelectedRevisionItem.tsx
```diff
@@ -76,6 +76,8 @@ function SelectedRevisionItem({
       className={`item-container item-${index} item-${searchType}`}
       data-testid='selected-rev-item'
     >
+      <input type='hidden' name={searchType + 'Rev'} value={item.revision} />
+      <input type='hidden' name={searchType + 'Repo'} value={repository} />
```
julienw, line 80: Note @Carla-Moz we need `searchType` here, just for naming the fields

src/__tests__/CompareResults/RevisionSelect.test.tsx
```diff
@@ -188,69 +130,4 @@ describe('Revision select', () => {
       }),
     ).toStrictEqual([]);
   });
-
-  it('Should render results for a998c42399a8', async () => {
```
julienw, line 192: I removed this one completely because this was essentially the same test and testing the same behavior as the previous test block.

src/__tests__/CompareResults/RevisionSelect.test.tsx
```diff
@@ -122,59 +101,22 @@ describe('Revision select', () => {
         selector: 'a',
       }),
     ).toStrictEqual([]);
-  });
-
-  it('Should render results for 9d5066525489', async () => {
```
julienw, line 127: I merged this into the previous one, so that all the initialization is removed.

src/components/CompareResults/loader.ts
```diff
@@ -115,34 +143,61 @@ async function fetchAllFakeCompareResults() {
...
+
+  const { baseRev, baseRepo, newRevs, newRepos, frameworkId } = checkValues({
+    baseRev: baseRevFromUrl,
+    baseRepo: baseRepoFromUrl,
+    newRevs: newRevsFromUrl,
+    newRepos: newReposFromUrl,
+    framework: frameworkFromUrl,
+  });
+
+  const resultsForAllRevs = await fetchCompareResultsOnTreeherder({
```
Carla-Moz: This is so much simpler and cleaner. Thanks so much for revamping this structure! Really liking this react-router API.

Carla-Moz (approved):
Hi @julienw, this looks amazing; I appreciate all the work you've done to revamp the architecture here on so many levels. It's a lot simpler and cleaner with the react router API. Hopefully merging will be easy post my review edit work. I've also taken note of how you structure your commits and their messages for the future.

PR #586, opened 2024-01-04
Fix the setup of jest fake timers so that they don't run unexpectedly

julienw:
This also removes the running of all timers at the end of some tests, that were working around the faulty setup.

I noticed this issue while working on another patch, and decided to extract it.

### DIFF
--- a/src/__tests__/utils/setupTests.ts
+++ b/src/__tests__/utils/setupTests.ts
@@ -92,7 +92,7 @@ afterEach(() => {
   jest.clearAllMocks();
   jest.resetAllMocks();
   jest.restoreAllMocks();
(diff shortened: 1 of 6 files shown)


src/__tests__/utils/setupTests.ts
```diff
@@ -65,7 +65,7 @@ afterEach(() => {
   jest.clearAllMocks();
   jest.resetAllMocks();
   jest.restoreAllMocks();
-  jest.runOnlyPendingTimers();
+  jest.clearAllTimers();
```
julienw, line 68: Instead of running the pending timers (moreover outside of act !!), we're just clearing them.

PR #584, opened 2023-12-21
Use fetch-mock-jest to mock fetch requests

julienw:

PR #587, opened 2024-01-04
Simplify how the render functions are defined and used in test files

julienw:
Before this patch, we have 2 different render functions: `render` and `renderWithRouter`. We also have a `StoreProvider` that's just `react-redux`'s `Provider` pre-configured, that we use in some hooks-related test files.

The 2 render functions are "created" in `beforeEach` and exposed by `setupTests`. Instead of "creating" them, in this patch they're now defined directly, and they can be imported directly from `test-utils` in the test files. The goal is that it's now easier to create new flavors for these functions, that I'd like to use in #582.

The `store` object is imported from `setupTests` instead of being passed as a parameter. I considered creating it in the  `render` functions (possibly as a default parameter, so that we could pass another store if needed) and returning it, but this would require changing all call sites, so I decided to keep it simple for now. We could always change it more later.

This is on top of #584 and #586 to minimize rebase conflicts, so take care to only look at the few un-labeled commits at the end of the commit list.

### DIFF
--- a/src/__tests__/CompareResults/RevisionRow.test.tsx
+++ b/src/__tests__/CompareResults/RevisionRow.test.tsx
@@ -1,9 +1,7 @@
-import { screen } from '@testing-library/react';
-
 import RevisionRow from '../../components/CompareResults/RevisionRow';
 import { Platform } from '../../types/types';
 import getTestData from '../utils/fixtures';
-import { render } from '../utils/setupTests';
+import { render, screen } from '../utils/test-utils';
 
 describe('<RevisionRow>', () => {
   it.each([
--- a/src/__tests__/Search/SearchView.test.tsx
+++ b/src/__tests__/Search/SearchView.test.tsx
@@ -1,12 +1,16 @@
-import { renderHook } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 
 import SearchView from '../../components/Search/SearchView';
 import { Strings } from '../../resources/Strings';
 import useProtocolTheme from '../../theme/protocolTheme';
 import getTestData from '../utils/fixtures';
-import { renderWithRouter } from '../utils/setupTests';
-import { screen, act, FetchMockSandbox } from '../utils/test-utils';
+import {
+  screen,
+  act,
+  renderWithRouter,
+  renderHook,
+  FetchMockSandbox,
+} from '../utils/test-utils';
 
 const protocolTheme = renderHook(() => useProtocolTheme()).result.current
   .protocolTheme;
--- a/src/__tests__/Search/SelectedRevision.test.tsx
+++ b/src/__tests__/Search/SelectedRevision.test.tsx
@@ -1,6 +1,4 @@
-import { renderHook } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { act } from 'react-dom/test-utils';
 
 import SearchView from '../../components/Search/SearchView';
 import { updateCheckedRevisions } from '../../reducers/SearchSlice';
@@ -8,8 +6,15 @@ import { Strings } from '../../resources/Strings';
 import useProtocolTheme from '../../theme/protocolTheme';
 import { InputType } from '../../types/state';
 import getTestData from '../utils/fixtures';
-import { renderWithRouter, store } from '../utils/setupTests';
-import { screen, within, FetchMockSandbox } from '../utils/test-utils';
+import { store } from '../utils/setupTests';
+import {
+  screen,
+  within,
+  renderHook,
+  renderWithRouter,
+  act,
+  FetchMockSandbox,
+} from '../utils/test-utils';
 
 const protocolTheme = renderHook(() => useProtocolTheme()).result.current
   .protocolTheme;
--- a/src/__tests__/hooks/useFetchCompareResults.test.tsx
+++ b/src/__tests__/hooks/useFetchCompareResults.test.tsx
@@ -1,6 +1,14 @@
+import { Provider } from 'react-redux';
+
 import useFetchCompareResults from '../../hooks/useFetchCompareResults';
-import { StoreProvider } from '../utils/setupTests';
+import { store } from '../utils/setupTests';
 import { renderHook, FetchMockSandbox } from '../utils/test-utils';
+
+function renderFetchCompareResultsHook() {
+  return renderHook(() => useFetchCompareResults(), {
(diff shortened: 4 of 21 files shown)


src/__tests__/CompareResults/RevisionRow.test.tsx
```diff
@@ -1,9 +1,7 @@
-import { screen } from '@testing-library/react';
```
julienw, line 1: note: I also took the opportunity to bring together all testing-library-related things, and import them all from `test-utils`.

src/__tests__/Search/SelectedRevision.test.tsx
```diff
@@ -1,15 +1,20 @@
-import { renderHook } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { act } from 'react-dom/test-utils';
```
julienw, line 3: note: it's incorrect to import `act` from `'react-dom/test-utils'` when using it with testing-library. We need to use the one exported by `testing-library` that does a little bit more stuff. It's reexported in `test-utils`.

src/__tests__/hooks/useFetchCompareResults.test.tsx
```diff
@@ -1,7 +1,15 @@
...
-import { StoreProvider } from '../utils/setupTests';
+import { store } from '../utils/setupTests';
 import { renderHook, FetchMockSandbox } from '../utils/test-utils';
 
+function renderFetchCompareResultsHook() {
+  return renderHook(() => useFetchCompareResults(), {
+    wrapper: ({ children }) => <Provider store={store}>{children}</Provider>,
+  });
+}
+
```
julienw: I decided to remove `StoreProvider` to replace it with this simple function instead. Tell me what you think.

src/__tests__/hooks/useHandleChangeSearch.test.tsx
```diff
@@ -1,13 +1,19 @@
...
+
 import useHandleChangeSearch from '../../hooks/useHandleChangeSearch';
 import { setInputError, updateSearchResults } from '../../reducers/SearchSlice';
 import { InputType, Repository } from '../../types/state';
 import getTestData from '../utils/fixtures';
-import { store, StoreProvider } from '../utils/setupTests';
+import { store } from '../utils/setupTests';
 import { renderHook, FetchMockSandbox } from '../utils/test-utils';
 
-jest.useFakeTimers();
```
julienw, line 11: It's done in beforeEach in the setup file.

src/__tests__/Search/SearchView.test.tsx
```diff
@@ -58,8 +57,11 @@ describe('Search View', () => {
 
   it('renders a skip link that sends the focus directly to search container', async () => {
     renderComponent();
-    await waitFor(() => userEvent.tab());
-    await waitFor(() => userEvent.keyboard('{Enter}'));
+
+    // set delay to null to prevent test time-out due to useFakeTimers
+    const user = userEvent.setup({ delay: null });
+    await user.tab();
+    await user.keyboard('{Enter}');
```
julienw, line 64: I'm not sure why, but this was making a _later_ test in this file fail.

This was incorrect too: `waitFor` should be used with things that throw like expectations. In this case I believe it was working around the fact that we use fake timers, because `waitFor` internally advance the fake timeouts. I'm not sure either why it was failing before this change, my guess is some conflicts with how testing-library's waitFor interacts with fake timers...

src/__tests__/Search/SearchView.test.tsx
```diff
@@ -1,21 +1,34 @@
...
+  (global.fetch as FetchMockSandbox).get(
+    'begin:https://treeherder.mozilla.org/api/project/try/push/',
+    {
+      results: testData,
+    },
+  );
+}
+
 function renderComponent() {
+  setupTestData();
```
Carla-Moz, line 31: I like and prefer this new implementation. Looks easier too.

src/__tests__/Search/SearchView.test.tsx
```diff
@@ -152,79 +140,78 @@ describe('Base Search', () => {
...
     const user = userEvent.setup({ delay: null });
     renderComponent();
 
     const searchInput = screen.getAllByRole('textbox')[0];
 
+    // We're running fake timers after each user action, because the input
+    // normally waits 500ms before doing requests. Because we want to test the
+    // requests, we have to run the timers to properly assess the result.
+    // They're all wrapped in "act" because they might also trigger state
+    // changes and rerenders.
```
Carla-Moz, line 160: Do we have to run all the timers post all user actions in our test. I understand why but I'm wondering if the above is an example or a standard we want to follow with current and future tests?

Carla-Moz (approved):
No major changes here requested. I left a few comments and a question. Thanks for this! I'm looking forward to implementing the new render functions in new tests.

src/__tests__/Search/SearchView.test.tsx
julienw: not always and all the time
but _in this case_ we need to run them so that the action happening after 500ms is actually triggered.

Another way would be to use `screen.findByXXX`, for example we could wait for the popup to be displayed, and then for the popup to disappear. I didn't think of that when I wrote it initially.
`screen.findByXXX` uses `act` already so we wouldn't need to use it explicitely ourselves, and it also runs fake timers automatically.
It's also closer to what a normal user would do (do an action, then wait for something happening).

Let me try to do it this way and if I can make it work easily I'll land it. Otherwise I'll keep it this way. What do you think?

src/__tests__/utils/test-utils.tsx
```diff
@@ -5,88 +5,53 @@ import { SnackbarProvider } from 'notistack';
...
-// override render method
-export type Render = ReturnType<typeof createRender>;
-export type RenderWithRouter = ReturnType<typeof createRenderWithRouter>;
-export {
-  createRender,
-  createRenderWithRouter,
-  createRouterWrapper,
-  createStoreProvider,
-};
+export { FetchMockSandbox } from 'fetch-mock';
```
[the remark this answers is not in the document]
julienw: Sounds good a good idea, I'll do it!

src/__tests__/Search/SearchView.test.tsx
Carla-Moz: I think for the sake of time you can land it and we can update later. But the latter is a good idea re `screen.findByXXX`.

PR #590, opened 2024-01-12
Move dark mode to redux #2

beatrice-acasandrei:


julienw:
The test problem is that a lot of tests are doing something like this:
```js
  const protocolTheme = renderHook(() => useProtocolTheme()).result.current
    .protocolTheme;
  const themeMode = protocolTheme.palette.mode;
```
Because `themeMode` is then passed to all components as a prop in the rest of the test.
This breaks because `useProtocolTheme` uses Redux now, so we'd have to specify a wrapper as the second parameter to `renderHook` ([see doc](https://testing-library.com/docs/react-testing-library/api/#renderhook)).

But... We don't need that anymore, because we don't pass `themeMode` to components anymore! :-)
So just remove all these lines (like eslint tells you), and everything should work :)
Fix tsc and eslint errors first, then let's see how the tests look like.

src/__tests__/utils/test-utils.tsx
```diff
@@ -6,15 +6,15 @@ import { Provider } from 'react-redux';
...
+import getProtocolTheme from '../../theme/protocolTheme';
 import { store } from './setupTests';
 
 type ChildrenProps = { children: React.ReactElement };
 
 type ThemeConfig = Partial<Theme> | null;
 export function render(ui: React.ReactElement, themeConfig?: ThemeConfig) {
   function Wrapper({ children }: ChildrenProps) {
-    const { protocolTheme } = useProtocolTheme();
+    const { protocolTheme } = getProtocolTheme('light');
```
beatrice-acasandrei, line 17: @julienw Here I'm not sure how I can avoid hardcoding if the approach from the last two commits is alright. Let me know what you think. Thanks.

julienw (commented):
I didn't try it but I like the way it looks like. I gave some comments about moving a little bit of code around but I think we can probably land it afterwards.
Tell me what you think!

src/reducers/ThemeSlice.ts
```diff
@@ -0,0 +1,19 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+const initialState = {
+  mode: localStorage.getItem('theme') || 'light',
```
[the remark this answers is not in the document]
julienw: Please forget about this "other suggestion" for now, because this needs to change all `useAppSelector` for the theme mode everywhere, and I'd rather move this PR forward than do that.
In the future we can encapsulate the "preferedMode computation" in a hook, but that's out of scope.

src/reducers/ThemeSlice.ts
```diff
@@ -0,0 +1,20 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+import { ThemeMode } from '../types/state';
+
+const initialState = {
+  mode: localStorage.getItem('theme') === 'dark' ? 'dark' : 'light',
```
julienw, line 6: nit: add `as ThemeMode` at the end, to force `mode` to use the type `ThemeMode` (currently it looks like it's infered as a generic string)
Alternatively you can type initialState explicitely:
```js
type State = {
  mode: ThemeMode
};
const initialState: State = {
  mode: ...
};
```

src/__tests__/utils/test-utils.tsx
julienw: I'm fine with that, especially that can be overriden by themeConfig. I'm not sure where themeConfig is used in tests currently (couldn't find it with a simple search), so if we need anything better in the future we can also revisit this parameter.

julienw (approved):
this looks good to me, thanks!
Just one simple comment and you can land it.

I'm so not looking forward to fixing my rebase conflicts 🙈, but it's for a good cause

PR #591, opened 2024-01-16
PCF-361 Add aria roles for tables / rows / rowgroup etc

esanuandra:
This is a follow-up PR to address https://github.com/mozilla/perfcompare/pull/580#pullrequestreview-1770187205

### DIFF
--- a/src/components/CompareResults/RevisionRow.tsx
+++ b/src/components/CompareResults/RevisionRow.tsx
@@ -168,37 +168,46 @@ function RevisionRow(props: RevisionRowProps) {
   };
   return (
     <>
-      <div className={`revisionRow ${styles.revisionRow} ${styles.typography}`}>
-        <div className='platform cell'>
+      <div
+        className={`revisionRow ${styles.revisionRow} ${styles.typography}`}
+        role='row'
+      >
+        <div className='platform cell' role='cell'>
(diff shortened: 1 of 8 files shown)


julienw:
Looking at the accessibility panel, the structure looks good to me!
I'm just not sure about the expanded row, if that should be a separate row. Ideally it should be inside the "parent" row, but not 100% sure how that should look in terms of nesting and roles. I think this could still go in with this structure and we can improve this part specifically later.

julienw:
I thought about that more over the night, now I think the expanded row shouldn't have the role=row, just make it a normal div. Later on we can work with `aria-expanded` and `aria-owns` but let's not focus on that now.

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -241,6 +250,7 @@ function RevisionRow(props: RevisionRowProps) {
           expanded ? 'expanded' : 'default'
         } ${stylesCard.container} `}
         data-testid='expanded-row-content'
+        role='row'
```
julienw, line 253: I would remove this role

julienw (approved):
ship it :)

It would be good to double check with a real screen reader but we can do this later when doing the a11y audit.

PR #595, opened 2024-01-18
Edit revision clean 2

Carla-Moz:
This PR closes the following Jira issues:

- [PCF-318 In Results view, on edit, the results get appended to the old results instead of replacing them](https://mozilla-hub.atlassian.net/browse/PCF-318)

- [PCF-354 Results: The revision(s) need(s) to be preselected in edit mode search dropdown for both base and new revisions ](https://mozilla-hub.atlassian.net/browse/PCF-354)

- [PCF-356 Results: in edit mode, clicking the cancel button doesn't return to default state after deleting revision(s)](https://mozilla-hub.atlassian.net/browse/PCF-356)

- [PCF-357 Results: in edit mode, clicking the save button does not update the state of the revisions correctly](https://mozilla-hub.atlassian.net/browse/PCF-357)

Please look at this commit - [e2e2d46](https://github.com/mozilla/perfcompare/pull/595/commits/e2e2d4667f2c93ff2959678ebf3bb85feaac3f1b) for relevant changes

The relevant files to look at for this review:
```
src/components/Search/CompareWithBase.tsx
src/components/Search/SearchComponent.tsx
src/components/Search/EditButton.tsx
src/components/Search/SaveCancelButtons.tsx
src/components/Search/SearchResultsList.tsx
src/components/Search/SearchResultsListItem.tsx
src/components/Search/SelectedRevisions.tsx
src/components/Search/SelectedRevisionItem.tsx
 
src/hooks/useCheckRevision.ts
src/hooks/useSelectedRevisions.ts
 
src/reducers/SelectedRevisionsSlice.ts
```

This review implements the "staging", "in progress", and "committed" states of the edit feature where the `CompareWithBase` handles the state of the selected revisions "in staging" and "in progress". Please look at the comments on this closed PR for reference: https://github.com/mozilla/perfcompare/pull/581

Tests and coverage will be handled in a separate PR.


src/components/Search/SearchResultsListItem.tsx
```diff
@@ -53,25 +68,51 @@ const styles = {
...
-  searchType,
+  isEditable,
+  isBase,
+  inProgress,
+  setInProgress,
 }: SearchResultsListItemProps) {
+  const searchType = isBase ? 'base' : 'new';
+  const { handleToggle } = useCheckRevision(isBase, isEditable);
+
   const isChecked: boolean = useAppSelector((state) =>
```
julienw, line 79: nit: name this `isCommittedChecked` so that the ternary below is clearer, and you can rename `isCheckedState` to `isChecked`.

src/components/Search/SelectedRevisionItem.tsx
```diff
@@ -29,46 +27,61 @@ import {
...
+  isBase,
+  mode,
+  inProgress,
+  setInProgress,
+  isWarning,
 }: SelectedRevisionItemProps) {
+  const searchType = isBase ? 'base' : 'new';
   const styles = SelectRevsStyles(mode);
-  const revisionHash = truncateHash(item.revision);
+  const revisionHash = item ? truncateHash(item.revision) : '';
```
julienw, line 63: Why this ternary, is it possible that item is absent?

src/components/Search/SelectedRevisionItem.tsx
```diff
@@ -138,7 +151,9 @@ function SelectedRevisionItem({
           name='close-button'
           aria-label='close-button'
           className={`${iconClassName} revision-action close-button`}
-          onClick={handleClose}
+          onClick={
+            isEditable ? handleRemoveEditViewRevision : handleRemoveRevision
```
julienw, line 155: Like suggested above, I would recommend to do this distinction in the parent instead of here, so that we don't have to pass `setInProgress`, and just have a callback `onRemove` on the item.

julienw (commented):
Thanks Carla, this is a lot of work!

For this PR I'd like that we focus on getting it landed, because it has quite a few structural changes and the PR is already big, I don't want this to bitrot.
I left some comments that we can do later, so you can read them but please ignore them otherwise for the sake of this PR. They're especially about the "checked revisions" state. I believe we will also see a bit clearer after my patch is in.

I also left some comments that I'd like that we do now: that's places where instead of passing simple callbacks, you passed directly react state setters. Because of this, you also have to pass a lot of other information so that the components can call them with everything needed, and they end up with too much responsibility.
Instead, these components should just "notify" their parents that something changed (the user clicked there, toggled this, etc), as an event if you will, and the parents will change the various states as an event handler. Think of this just like events in DOM.
I believe this shouldn't be too many changes, mostly moving things around.

I think there's also a small conceptual issue about how to deal with the various states in the 2 modes (editable/non editable), but we can deal with this in a more targeted PR later. The issue IMO is that you used `inProgressState` only when the component is editable, while I believe we can use it in both cases. I believe this is entangled with the "checked revisions" state I already mentioned above, and we can deal with both problems later.

First things first => let's get this PR landed!
I suggest that you squash all the existing commits into one first, that should also make the rebase after Beatrice's dark mode lands easier.
Then please add all the new changes to one separate commit.

.codecov.yml
```diff
@@ -0,0 +1,9 @@
+ignore:
+  - 'src/components/Search/CompareWithBase'
+  - 'src/components/CompareResults/ResultsView.tsx'
+  - 'src/components/Search/SaveCancelButtons.tsx'
+  - 'src/components/Search/SearchResultsListItem.tsx'
+  - 'src/components/Search/CompareWithBaseContext.tsx'
+  - 'src/components/Search/EditButton.tsx'
+  - 'src/hooks/useCheckRevision.ts'
+  - 'src/components/Search/SearchComponent.tsx'
```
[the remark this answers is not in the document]
Carla-Moz: Understood. I've removed it.

src/reducers/SelectedRevisionsSlice.ts
```diff
@@ -1,72 +1,31 @@
...
-  new: [],
+  baseCommittedRevisions: [],
+  newCommittedRevisions: [],
 };
 
 const selectedRevisions = createSlice({
   name: 'selectedRevisions',
   initialState,
   reducers: {
+    //when the user presses the "Compare" or "Cancel" (in edit mode) buttons
```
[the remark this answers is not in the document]
Carla-Moz: Yes, only on Compare. This is an old comment.

src/components/Search/SelectedRevisionItem.tsx
Carla-Moz: Yes, something before made the item absent, but that's all fixed so this irrelevant code.

src/components/Search/CompareWithBase.tsx
```diff
@@ -57,6 +111,86 @@ function CompareWithBase({
...
+    });
+  }, [baseRevs, newRevs]);
+
+  const toggleIsExpanded = () => {
+    setExpanded(!expanded);
+  };
+
+  const handleCancelBase = () => {
+    setInProgressBase({ ...revRepos, isInProgress: false });
+    dispatch(clearCheckedRevisionforType({ searchType: 'base' }));
```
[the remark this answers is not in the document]
Carla-Moz: I understand; I'll create a new ticket to handle this cleanup in a new PR.

PR #597, opened 2024-01-19
React-router use: instanciate the router outside of the react tree, and stop using hashes for our URLs

julienw:
Please look at the last commits only, as the first commits are #582.

This does these (fairly unrelated) things:
* Commit 1: instanciate the router outside of the react tree, like advised by the react router documentation, so that we don't regenerate one at each rerender. This should improve performance.
  This wasn't so easy before because we had to get the theme from a hook and pass it as a parameter, but thanks to #590 this is now possible.
* Commit 2: this uses [`createBrowserRouter`](https://reactrouter.com/en/main/routers/create-browser-router) instead of [`createHashRouter`](https://reactrouter.com/en/main/routers/create-hash-router).
  This makes more natural URLs without the `#` character. https://perf.compare/#/compare-results will be now just https://perf.compare/compare-results. Or in the deploy preview https://deploy-preview-597--mozilla-perfcompare.netlify.app/compare-results?fakedata=true.
  (A quick note here: I noticed that adding a `/` after `compare-results` doesn't work, and we don't even have a 404 page. I'll look at it later)
* Commit 3: for the browser router to work, we need to configure netlify though, and this commit does just that. See https://docs.netlify.com/routing/redirects/rewrites-proxies/#history-pushstate-and-single-page-apps for more information about that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -56,6 +56,7 @@
         "babel-loader": "^8.2.5",
         "babel-preset-jest": "^28.1.3",
         "case-sensitive-paths-webpack-plugin": "^2.4.0",
+        "copy-webpack-plugin": "^12.0.2",
         "core-js": "^3.24.0",
         "css-loader": "^6.7.1",
         "eslint": "^8.20.0",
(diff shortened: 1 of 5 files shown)


Carla-Moz (approved):
LGTM, glad to be rid of the hashes so thanks!

PR #598, opened 2024-01-22
PCF-362 Fix dark mode on results table

esanuandra:
This PR addresses [PCF-362 Fix dark mode on results table](https://mozilla-hub.atlassian.net/browse/PCF-362).

The table background is the same as the page background. Below there is a before and after.
Screenshot 2024-01-22 at 5 25 57 PM

Screenshot 2024-01-22 at 5 25 02 PM


julienw (approved):
Thanks, looks good to me!
but we should probably replace Paper with Box or Container, unless we intend to use the elevation property... and in that case we'll have to do something different than background: none.
A story for another day though.

PR #599, opened 2024-01-26
Edit Review Clean - 3

Carla-Moz:
Resolves the comments to simplify the child components with simple callbacks rather than state setters. See [here](https://github.com/mozilla/perfcompare/pull/595)
Squashed commits - cleaner history


src/components/CompareResults/ResultsView.tsx
```diff
@@ -84,18 +119,18 @@ function ResultsView(props: ResultsViewProps) {
...
+      //     reposArray[index] as Repository['name'],
+      //     rev,
+      //     'base',
+      //   );
+      //   void searchRecentRevisions(
+      //     reposArray[index] as Repository['name'],
+      //     rev,
+      //     'new',
+      //   );
+      // });
```
julienw, line 133: Please remove this commented code before landing

src/components/CompareResults/ResultsView.tsx
```diff
@@ -42,7 +77,7 @@ function ResultsView(props: ResultsViewProps) {
 
   const { dispatchFetchCompareResults, dispatchFakeCompareResults } =
     useFetchCompareResults();
-  const { searchRecentRevisions } = useHandleChangeSearch();
+  // const { searchRecentRevisions } = useHandleChangeSearch();
```
julienw, line 80: Please remove the commented code before landing

src/components/Search/SearchComponent.tsx
```diff
@@ -125,6 +120,34 @@ function SearchComponent({
     }
   };
 
+  const onCancel = () => {
+    if (isBaseComp) handleCancel(true);
+    if (!isBaseComp) handleCancel(false);
```
julienw, line 125: You could simply use:
`handleCancel(isBasecomp)`

Please don't bother changing this because as I wrote above, I believe we can get rid of `isBaseCamp` altogether, we'll do that in a next PR. I just wanted to point out that these if conditions were not necessary, rather you could use a more linear code.

julienw (approved):
This is much better, thanks for the update!

I think we can simplify this even further, but let's do that after this PR lands.
One thing I especially noticed is still this `isBase` boolean: it's not useful! Instead you can pass different callbacks to each of the `<SearchComponent>` instances. You also don't need to declare the callbacks formally, you can pass define directly in the prop if they're simple (for example if it's just one call to change a state).
But again, let's land this so that we can move forward, and we can do these small fixes in a next PR.

Let me also do a few comments that will make our team work better moving forward:
1. Instead of squashing all of your previous work like I suggested, so that I could look at the new changes only, you squashed your previous work and the first set of changes together. As a result I can't see the new changes only and I have to look at the full set again. Next time, please make sure to at least keep the new changes separate from the old changes.
2. Please look at writing better commit logs, better reflecting what's inside a specific commit. If a specific commit makes no sense by itself, maybe it's a good sign it should be squashed with others. Commit logs are really important to the reviewer, so that they can have a sense of what's happening in one commit.

src/components/Search/SearchComponent.tsx
Carla-Moz: True, not sure why I wrote it that. Must be tired. Will fix in next PR

PR #601, opened 2024-01-30
Simplify callbacks in Compare With Base child comps

Carla-Moz:
@julienw, I kept the `isBaseComp` for styling reasons. Otherwise, it's removed from any state updates.  In a new patch, I'll redo the styles so we can eliminate the prop completely.


julienw (approved):
Much better, thanks!

As for completely removing `isBaseComp` there are a few issues for now. Especially the "useCheckedRevisions" hook needs it still so we'll need to work on that first. But we'll have to see if that's more important than other work we have to do.

PR #607, opened 2024-02-23
PCF-378 Fix aria-label attributes Part 1

esanuandra:
This PR addresses [Fix aria-label attributes](https://mozilla-hub.atlassian.net/browse/PCF-378)

- [x] Add aria-label on Link components
- [x] Add/modify aria-label for some button components

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -99,11 +99,19 @@ function App() {
           <Alert className={Banner} severity='warning'>
             <div className='banner-text'>
               {strings.text}{' '}
-              <Link href={strings.href} target='_blank'>
+              <Link
+                href={strings.href}
+                target='_blank'
+                aria-label={'link to file bug in bugzilla'}
+              >
                 {strings.linkText}
               </Link>{' '}
               {contact.text}{' '}
-              <Link href={contact.href} target='_blank'>
+              <Link
+                href={contact.href}
+                target='_blank'
+                aria-label={'link to perfcompare matrix channel'}
+              >
                 {contact.linkText}
               </Link>
               .
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -99,19 +99,11 @@ function App() {
           <Alert className={Banner} severity='warning'>
             <div className='banner-text'>
               {strings.text}{' '}
-              <Link
-                href={strings.href}
-                target='_blank'
-                aria-label={'link to file bug in bugzilla'}
-              >
+              <Link href={strings.href} target='_blank'>
                 {strings.linkText}
               </Link>{' '}
               {contact.text}{' '}
-              <Link
-                href={contact.href}
-                target='_blank'
-                aria-label={'link to perfcompare matrix channel'}
-              >
+              <Link href={contact.href} target='_blank'>
                 {contact.linkText}
               </Link>
               .
--- a/src/components/CompareResults/RevisionHeader.tsx
+++ b/src/components/CompareResults/RevisionHeader.tsx
@@ -101,6 +101,7 @@ function RevisionHeader(props: RevisionHeaderProps) {
         <Link
           href={getTreeherderURL(header.new_rev, header.new_repo)}
           target='_blank'
+          aria-label={`link to revision jobs in treeherder for ${shortHash}`}
         >
           {shortHash}
         </Link>
--- a/src/components/CompareResults/RevisionHeader.tsx
+++ b/src/components/CompareResults/RevisionHeader.tsx
@@ -1,6 +1,7 @@
 import { Link } from '@mui/material';
 import { style } from 'typestyle';
 
+import { Strings } from '../../resources/Strings';
 import { Colors, Spacing } from '../../styles';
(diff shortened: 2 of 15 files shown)


src/components/CompareResults/RevisionRow.tsx
```diff
@@ -207,7 +207,11 @@ function RevisionRow(props: RevisionRowProps) {
...
               <IconButton aria-label='graph link' size='small'>
-                <Link href={graphLink} target='_blank'>
+                <Link
+                  href={graphLink}
+                  target='_blank'
+                  aria-label='link to revision graph'
+                >
                   <TimelineIcon />
                 </Link>
               </IconButton>
```
esanuandra, line 217: I think we need to remove one `aria-label` either from the `IconButton` or from the `Link`.

esanuandra, line 217: I think we need to remove one `aria-label` either from the `IconButton` or from the `Link`.

src/components/Search/SaveCancelButtons.tsx
```diff
@@ -38,7 +38,7 @@ export default function SaveCancelButtons({
         className={`cancel-save cancel-button ${cancelBtn.main} cancel-button-${searchType}`}
         role='button'
         name='cancel-button'
-        aria-label='cancel button'
+        aria-label='cancel edit'
```
julienw, line 41: These ones don't need any aria-label, as they already have a good text. 

[Rule 1 of ARIA: don't use ARIA](https://www.deque.com/blog/top-5-rules-of-aria/)

src/components/CompareResults/RevisionHeader.tsx
```diff
@@ -101,6 +101,7 @@ function RevisionHeader(props: RevisionHeaderProps) {
         <Link
           href={getTreeherderURL(header.new_rev, header.new_repo)}
           target='_blank'
+          aria-label={`link to revision jobs in treeherder for ${shortHash}`}
```
julienw, line 104: A tooltip could be useful here (instead of the aria-label) for also sighted users.

src/components/App.tsx
```diff
@@ -99,11 +99,19 @@ function App() {
...
               </Link>{' '}
               {contact.text}{' '}
-              <Link href={contact.href} target='_blank'>
+              <Link
+                href={contact.href}
+                target='_blank'
+                aria-label={'link to perfcompare matrix channel'}
+              >
                 {contact.linkText}
               </Link>
```
julienw, line 116: We don't need aria-label here, the text is already good.
(cf Rule 1 of ARIA)

src/components/Search/SelectedRevisionItem.tsx
```diff
@@ -93,6 +93,7 @@ function SelectedRevisionItem({
                 <Link
                   href={getTreeherderURL(item.revision, repository)}
                   target='_blank'
+                  aria-label={`link to revision jobs in treeherder for ${revisionHash}`}
```
julienw, line 96: As said above for a similar situation, we probably can use a title here, as this can be useful to sighted users.

julienw (changes requested):
Thanks for looking at this Andra!

I left a bunch of comments, tell me what you think!

In (not very) short:
* there are places where we don't need aria-label at all (I recommend that you look at the elements using the devtools' awesome accessibility panel that makes it possible to inspect all the accessibility properties in the page)
* there are places where a tooltip might be more appropriate, when the information can be useful for a sighted user. To keep simple for now, we can use the HTML `title` attribute that will give a tooltip automatically when the user hovers on the element for more than a few seconds. Please double check with the accessibility panel that this is properly exposed to assistive technologies.

Also I believe that we can improve how the `title` or `aria-label` are redacted. I think they can be shorter.

For example there are links where the aria-label  is "link to bla bla", but because it is a link, the assistive technology will already say to the user that it is a link! Rather just write as if there was text. For example "Remove revision" is good. For a link this coulod be "Go to XXX" instead of "Link to XXX".

I think we can use both aria-label or tooltip sometimes, as they don't have the same purpose always. A tooltip could give some more information where the aria-label would just be the blunt description. But as a first pass, let's keep it simple and use just one or the other, so that we don't risk saying different things.

Also it would be good that all text we add as tooltip or aria-label go to the Strings file.

julienw (approved):
Thanks, this looks better!

I made a few wording suggestions, tell me what you think! (also I'm not a super good english speaker so feel free to adjust or ask somebody who knows better)

If you apply them you can land the patch without another review.

But if you do want a last review, please add the new changes to a separate commit so that I can look at that one only :-)

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -235,7 +243,10 @@ function RevisionRow(props: RevisionRowProps) {
             onClick={toggleIsExpanded}
             data-testid='expand-revision-button'
           >
-            <IconButton aria-label='expand row' size='small'>
+            <IconButton
+              title={Strings.components.expandableRow.title}
```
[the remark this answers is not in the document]
esanuandra: right

PR #609, opened 2024-02-27
Added taskcluster packages

beatrice-acasandrei:


package.json
```diff
@@ -18,6 +18,8 @@
     "express": "^4.18.1",
     "fast-kde": "^0.2.1",
     "format": "^0.2.2",
+    "lodash.debounce": "^4.0.8",
+    "moment": "^2.30.1",
```
julienw, line 22: Probably added by mistake?

beatrice-acasandrei: I saw it being used to check if the credentials are expired. But probably not necessary at this point, could be installed when we add the logic for it.

julienw: It would be good to avoid "moment" because it's a quite big dependency, I think we can do similar checks by using simple timestamps.
Moment used to be useful to write dates and time in a locale friendly manner, but nowadays we have a lot of thing natively in browsers API (for example https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString).

"debounce" could be useful in some cases but probably not for this.

package-lock.json
```diff
@@ -25976,9 +26142,9 @@
       },
       "dependencies": {
         "semver": {
-          "version": "6.3.0",
-          "resolved": "https://registry.npmjs.org/semver/-/semver-6.3.0.tgz",
-          "integrity": "sha512-b39TBaTSfV6yBrapU89p5fKekE2m/NwnDocOVruQFS1/veMgdzuPcnOM34M6CwxW8jH/lxEa5rBoDeUwu5HHTw==",
+          "version": "6.3.1",
```
julienw, line 26145: It looks like you updated `semver` in the process, but that's not a big deal and I'm find with it.

julienw (approved):
Thanks!

PR #612, opened 2024-03-05
Request authorization code

beatrice-acasandrei:


src/logic/taskcluster.ts
```diff
@@ -0,0 +1,43 @@
...
+const getAuthCode = (defaultRootUrl: string) => {
+  const nonce = generateNonce();
+  localStorage.setItem('requestState', nonce);
+  localStorage.setItem('tcRootUrl', defaultRootUrl);
+
+  const params = new URLSearchParams({
+    // client_id: clientId,
+    client_id: treeherderClientId,
+    response_type: 'code',
+    // redirect_uri: redirectURI,
```
julienw, line 20: nit: please remove unused comments once we're done

src/types/types.ts
```diff
@@ -282,3 +282,10 @@ export type FakeCommitHash =
   | 'bb6a5e451dace3b9c7be42d24c9272738d73e6db'
   | '9d50665254899d8431813bdc04178e6006ce6d59'
   | 'a998c42399a8fcea623690bf65bef49de20535b4';
+
+export type UserCredentials = {
+  defaultRootUrl: {
```
julienw, line 287: I think you want a `Record<string, { expires, credentials }>`?

src/logic/taskcluster.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const checkTaskclusterCredentials = () => {
+  const defaultRootUrl = getRootUrl(prodFirefoxRootUrl);
+  const userCredentials = JSON.parse(
+    localStorage.getItem('userCredentials') as string,
+  ) as UserCredentials;
+
+  if (
+    !userCredentials ||
+    !userCredentials[defaultRootUrl as keyof UserCredentials]
```
julienw, line 37: you won't need the 'as keyof' if you fix the type as I suggest above.

src/logic/taskcluster.ts
```diff
@@ -0,0 +1,43 @@
...
+import { UserCredentials } from '../types/types';
+import { generateNonce } from '../utils/helpers';
+
+// For testing purposes
+const treeherderClientId = `treeherder-localhost-5000-client`;
+const treeherderRedirectURI = `http://localhost:5000/taskcluster-auth`;
+
+const getAuthCode = (defaultRootUrl: string) => {
+  const nonce = generateNonce();
+  localStorage.setItem('requestState', nonce);
```
julienw, line 13: Probably better to use `sessionStorage` (marginally better).
Can you add a small comment explaining why we need to store it (cf my other comment above)

src/logic/taskcluster.ts
```diff
@@ -0,0 +1,43 @@
...
+  localStorage.setItem('tcRootUrl', defaultRootUrl);
```
julienw, line 14: My understanding is that we store it so that we can use it in the callback, especially so that we can use the staging server if we initiated the request to the staging server. Is that right? Can you add a small comment for that?

julienw (commented):
I left a few comments, tell me what you think :-)

src/logic/taskcluster.ts
```diff
@@ -0,0 +1,67 @@
...
+  return window.crypto.randomUUID();
+};
+
+const getAuthCode = (tcParams: {
+  url: string;
+  redirectUri: string;
+  clientId: string;
+}) => {
+  const nonce = generateNonce();
+  sessionStorage.setItem('requestState', nonce); // The request state it's stored in sessionStorage so that it can be used in the callback
```
julienw, line 39: nit
```suggestion
  // The nonce is stored in sessionStorage so that it can be compared with the one received by the callback endpoint.
  sessionStorage.setItem('requestState', nonce);
```

src/logic/taskcluster.ts
```diff
@@ -0,0 +1,67 @@
...
+  sessionStorage.setItem('taskclusterUrl', tcParams.url);
```
julienw, line 40: nit: you can also add a comment here, such as
```
// The taskclusterUrl is also stored, so that the additional
// requests done by the callback endpoint use the same URL.
```

src/logic/taskcluster.ts
```diff
@@ -0,0 +1,67 @@
...
+    state: nonce,
+  });
+  const url = `${tcParams.url}/login/oauth/authorize/?${params.toString()}`;
+  window.open(url, '_blank');
+};
+
+export const checkTaskclusterCredentials = () => {
+  const taskclusterParams = getTaskclusterParams();
+  const userCredentials = JSON.parse(
+    sessionStorage.getItem('userCredentials') as string,
```
julienw, line 56: Note that the _userCredentials_ data should be in `localStorage`.

The difference with the other values is that this value is long-lived (several days/weeks) while the other values are short-lived (only used during the authentication flow).

src/logic/taskcluster.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const prodTaskclusterUrl = 'https://firefox-ci-tc.services.mozilla.com';
+export const stagingTaskclusterUrl =
+  'https://stage.taskcluster.nonprod.cloudops.mozgcp.net';
+export const stagingClientId = 'perfcompare-staging-client';
+export const tcClientIdMap: Record<string, string> = {
+  'https://perf.compare': 'production',
+  'https://beta--mozilla-perfcompare.netlify.app': 'beta',
+  'http://localhost:3000': 'localhost-3000',
+};
```
julienw, line 13: optional nit:
Instead of doing this small concatenation later:
```
    clientId: `perfcompare-${tcClientIdMap[window.location.origin]}-client`,
```
I'd just specify the full values directly here.

Advantages:
* more clarity about all the cliend_ids the project use
* easier to search from a client_id value
* more straightforward code

Disavantages:
* longer to type and to write

src/__tests__/CompareResults/ResultsView.test.tsx
```diff
@@ -150,4 +150,53 @@ describe('Results View', () => {
...
+      .get('begin:https://treeherder.mozilla.org/api/project/', {
+        results: [testData[0]],
+      });
+
+    const mockGetRandomValues = jest.fn().mockReturnValue(new Uint32Array(1));
+    Object.defineProperty(window, 'crypto', {
+      value: { getRandomValues: mockGetRandomValues },
+    });
+    window.open = jest.fn();
+    jest.spyOn(Storage.prototype, 'setItem');
```
julienw, line 172: You don't need to spy on this, you can just do things such as:
```js
expect(sessionStorage.taskclusterUrl).toBe("https://..");
expect(sessionStorage.requestState).toBe(fakeUuid);
```

julienw: I believe we don't even need to test exactly the content of `sessionStorage.requestState`, however we need to check that the URL passed to `window.open` contains it.
So maybe something like this:
```js
const windowOpenUrlString = window.open.mock.lastCall[0];
expect(windowOpenUrlString).toBe(...); // Like you did, I believe it's still good to check the full URL.

const windowOpenUrl = new URL(windowOpenUrlString);
expect(sessionStorage.requestState).toBe(windowOpenUrl.searchParams.get("state"));
expect(sessionStorage.taskclusterUrl).toBe(windowOpenUrl.origin);
```

Also because we don't mind about the exact value of the nonce, this allows us to not mock `randomUUID` and just use whatever data node crypto library gives us.

src/__tests__/CompareResults/ResultsView.test.tsx
```diff
@@ -150,4 +150,53 @@ describe('Results View', () => {
...
+        testCompareDataWithMultipleRuns,
+      )
+      .get('begin:https://treeherder.mozilla.org/api/project/', {
+        results: [testData[0]],
+      });
+
+    const mockGetRandomValues = jest.fn().mockReturnValue(new Uint32Array(1));
+    Object.defineProperty(window, 'crypto', {
+      value: { getRandomValues: mockGetRandomValues },
+    });
```
julienw, line 170: I believe we can use [Node version of window.crypto](https://nodejs.org/docs/latest-v18.x/api/webcrypto.html):

```js
import { webcrypto } from 'node:crypto';

...

Object.defineProperty(window, 'crypto', value: webcrypto);
```
(see https://github.com/jsdom/jsdom/issues/1612#issuecomment-1659591074 as well)

I think you can add this to `setupTests.ts` so that this is available for all tests.

src/logic/taskcluster.ts
```diff
@@ -0,0 +1,67 @@
...
+    redirect_uri: tcParams.redirectUri,
+    scope: 'hooks:trigger-hook:*',
+    state: nonce,
+  });
+  const url = `${tcParams.url}/login/oauth/authorize/?${params.toString()}`;
+  window.open(url, '_blank');
+};
+
+export const checkTaskclusterCredentials = () => {
+  const taskclusterParams = getTaskclusterParams();
```
julienw, line 54: We also need some code to handle the case when the origin isn't present (such as in deploy previews).
This could be as simple as:
```js
if (!taskclusterParams.clientId) {
  alert(`No clientId is configured for origin ${window.location.origin}, sorry!`);
  return;
}
```

This code works with the optional suggestion above `Instead of doing this small concatenation later ...`.
Otherwise you can do something such as:
```js
if (!(window.location.origin in tcClientIdMap)) {
  alert(`No clientId is configured for origin ${window.location.origin}, sorry!`);
  return;
}
```

Or even throw from `getTaskclusterParams` and catch from here (could be cleaner but maybe a little less easier to read).

Anyway, your choice, but please implement something like this :-)

julienw (commented):
at the first glance, the code looks correct!
let's make the tests and lint work now

src/utils/location.ts
```diff
@@ -0,0 +1,3 @@
+export const getLocationOrigin = () => {
```
julienw, line 1: nit: please add a comment explaining why this is in this file. (that we need it for tests, etc)

src/__tests__/CompareResults/ResultsView.test.tsx
```diff
@@ -164,12 +170,10 @@ describe('Results View', () => {
         results: [testData[0]],
       });
 
-    const mockGetRandomValues = jest.fn().mockReturnValue(new Uint32Array(1));
-    Object.defineProperty(window, 'crypto', {
-      value: { getRandomValues: mockGetRandomValues },
-    });
+    window.alert = jest.fn();
```
julienw, line 173: nit: it's cleaner to do:
```js
jest.spyOn(window, "alert").mockImplementation(() => {});
```
so that the original version is automatically restored at the end of the test.

src/__tests__/CompareResults/ResultsView.test.tsx
```diff
@@ -164,12 +170,10 @@ describe('Results View', () => {
...
+    const mockedWindowAlert = window.alert as jest.Mock;
     window.open = jest.fn();
```
julienw, line 175: same here

src/__tests__/CompareResults/ResultsView.test.tsx
```diff
@@ -34,6 +35,11 @@ function renderWithRoute(component: ReactElement) {
   });
 }
 
+jest.mock('../../utils/location', () => ({
+  getLocationOrigin: jest.fn(),
+}));
```
julienw, line 40: I believe you can just do
```
jest.mock('../../utils/location');
```
This automatically exports all exports with a mocked function.

julienw (commented):
I couldn't look super close but still wanted to give you this feedback. I can look at it closer once you fix these small nits!

src/__tests__/CompareResults/ResultsView.test.tsx
```diff
@@ -170,9 +168,9 @@ describe('Results View', () => {
         results: [testData[0]],
       });
 
-    window.alert = jest.fn();
+    jest.spyOn(window, 'alert').mockImplementation(() => {});
     const mockedWindowAlert = window.alert as jest.Mock;
-    window.open = jest.fn();
+    jest.spyOn(window, 'open').mockImplementation();
```
julienw, line 173: Interesting: the Jest documentation doesn't include this option, of calling `mockImplementation` without a parameter. This seems to work though... I'm a bit worried that this might fail in the future, but the risk is fairly small so I'm fine with keeping it this way.

Can you just make it the same for `window.alert` above? For consistency?

julienw (approved):
Let's land this and move forward!
Thanks

PR #613, opened 2024-03-07
PCF-399 Request token via https://firefox-ci-tc.services.mozilla.com/login/oauth/token using the authorisation code

esanuandra:
This fetches the token Bearer using an authorisation code. The token Bearer will be later used in another PR to fetch the access token.


src/components/TaskclusterAuth/TaskclusterCallback.tsx
```diff
@@ -0,0 +1,129 @@
...
+    const key = Object.keys(error);
+
+    errorMessage += ` ${key}: ${error[key]}`;
+  } else if (error) {
+    errorMessage += error;
+  }
+  return errorMessage ? errorMessage.trim() : error;
+};
+
+// from treeherder, TODO move this in another file/folder
```
beatrice-acasandrei, line 46: Maybe move it to src/logic/treeherder.ts?

src/components/TaskclusterAuth/TaskclusterCallback.tsx
```diff
@@ -0,0 +1,129 @@
...
+    };
+    console.log(`${rootUrl}/login/oauth/token`, options);
+    const response = await getData(`${rootUrl}/login/oauth/token`, options);
+    console.log('response ', response);
+
+    if (response.failureStatus) {
+      setErrorMessage(`errorMessage ${response.data}`);
+      return;
+    }
+
```
esanuandra, line 104: This is another extra verification. Maybe we can remove/move/replace this

src/__tests__/TaskclusterAuth/TaskclusterCallback.test.tsx
```diff
@@ -0,0 +1,21 @@
...
+        token_type: 'Bearer',
+      },
+    );
+
+    renderWithRouter(<TaskclusterCallback />, {
+      route: '/taskcluster-auth',
+      search: '?code=dwcygG5HQNaLiRe3RcTCbQ&state=OkCrH5isZncYqeJbRDelN',
+    });
+
+    expect(window.fetch).toHaveBeenCalledTimes(1);
```
esanuandra, line 19: could be more useful to use `toHaveBeenCalledWith`?

src/components/TaskclusterAuth/TaskclusterCallback.tsx
julienw: I'd move all the "gory" things related to requests to something such as logic/taskcluster.ts

src/components/TaskclusterAuth/TaskclusterCallback.tsx
```diff
@@ -0,0 +1,129 @@
...
+    if (code && requestState && requestState === state) {
+      void getCredentials(code);
+    } else {
+      if (errorMessage) {
+        setErrorMessage(
+          errorMessage + `We received error: ${errorMessage} from Taskcluster.`,
+        );
+      }
+    }
+  }, [code, state, requestState]);
```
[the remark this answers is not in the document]
julienw: My approach would be to throw a "real" Error when there is an error, and deal with it in an error boundary defined on the Route. See the documentation about that => https://reactrouter.com/en/main/route/error-element

It's pretty powerful because you can implement a completely different component to handle the error, possibly with a "try again" button or something like that. (for now just displaying the error would be good)

This makes it possible to focus on the happy case in the main component, but still handle errors separately.

src/components/TaskclusterAuth/TaskclusterCallback.tsx
```diff
@@ -0,0 +1,129 @@
...
+  return { data, failureStatus };
+};
+
+function TaskclusterCallback() {
+  const [searchParams] = useSearchParams();
+  const [errorMessage, setErrorMessage] = useState('');
+
+  const code = searchParams.get('code') as string;
+  const state = searchParams.get('state') as string;
+  const requestState = localStorage.getItem('requestState');
```
julienw, line 79: don't forget to check they are the same

src/components/TaskclusterAuth/TaskclusterCallback.tsx
```diff
@@ -0,0 +1,129 @@
...
+      setErrorMessage(errorMessage);
+      return;
+    }
+
+    // fetch token Bearer, it will be used to fetch the access_token
+    const options = {
+      mode: 'no-cors',
+      method: 'POST',
+      body: `grant_type=authorization_code&code=${taskclusterCode}&redirect_uri=${redirectURI}&client_id=${clientId}`,
+      headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
```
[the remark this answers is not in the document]
julienw: You can always try things in the command line: in the devtools' network monitor, you can "right click" the request, and use "copy as curl"; then you can use this command directly in the command line, and tweak it, etc.

src/components/TaskclusterAuth/TaskclusterCallback.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+    if (response.failureStatus) {
+      setErrorMessage(`errorMessage ${response.data}`);
+      return;
+    }
+
+    // TODO continue with fetchCredentials using response access_token Bearer
+  };
+
+  useEffect(() => {
```
julienw, line 108: instead of using an effect, I'd try to use a react router loader to do all the requests, and provide the component with just the result of this process.
More information about loaders: https://reactrouter.com/en/main/route/loader
Also see the example for the (more complex) compare results page: https://github.com/mozilla/perfcompare/blob/beta/src/components/CompareResults/loader.ts#L149-L150
As you can see, you can access the search params just as easily.

julienw (commented):
I left a few comments regarding the architecture for this work. Tell me what you think!

Also as a quick note, because I don't know if you're aware: for the next step, the "OPTIONS" request (as outlined in miro) is done automatically by the browser in a CORS context, you'll just have to do the GET. It's done automatically because we'll be adding the `Authorization` header.

Hope this helps!

src/components/TaskclusterAuth/TaskclusterCallback.tsx
```diff
@@ -0,0 +1,129 @@
+import { useEffect, useState } from 'react';
+
+import { useSearchParams } from 'react-router-dom';
+
+// TODO remove this after integrating the auth code logic
+localStorage.setItem('requestState', 'OkCrH5isZncYqeJbRDelN');
```
beatrice-acasandrei, line 6: We will use sessionStorage instead of localStorage because it is a transient value. (suggested by Julien)

julienw (approved):
Thanks, this looks reasonable! I didn't try it end to end yet but I trust you that you did it.

It would be good to move some of the code that's currently in `loader` to `logic/taskcluster`. 
I believe that the part that checks the state can stay in loader.ts.
Then the part that does the URLSearchParams + fetch can be in its own function, that takes as parameters `rootUrl` and `code`. It could be named `retrieveTaskclusterToken` for example. (The other one would be `retrieveTaskclusterCredentials`, if we follow the naming of their endpoints).

The retrieval of the right `clientId` can be done in that function directly too.

A good way to "smell" this, is that we should import actions (aka functions) from `logic/taskcluster`, but not low-level structures such as `tcClientIdMap` if that makes sense.

You can land this once you moved this over and checked that this works. No need to ask for a review again.
The test should stay just the same so it will be a good sign that this still works as expected.

Thanks again!

PR #616, opened 2024-03-08
Compare over time default state

Carla-Moz:
This PR begins the work on the Compare over Time component reflected in this ticket:
[Search: Add component for Compare over time](https://mozilla-hub.atlassian.net/browse/PCF-213)

Right now it covers only the first two checkboxes below.
The component is hidden until all work is completed.  Please add the `#comparetime` to the path to see and test it. Please let me know if there's anything I can add to make reviewing this simpler.  

- [x] Create default state
- [x] Create expanded state
- [ ] Handle checked and displayed selected revisions
- [ ] Time range dropdown
- [ ] Compare button functionality
- [ ] Edit feature for results page


src/styles/CompareCards.ts
```diff
@@ -23,6 +23,16 @@ export const CompareCardsStyles = (mode: string) => {
...
+    wrapper: {
+      $nest: {
+        '&.wrapper--withbase': {
+          marginBottom: `${Spacing.Large}px`,
+        },
+        '&.wrapper--overtime': {
+          marginBottom: `${Spacing.layoutXLarge - 12}px`,
+        },
+      },
+    },
```
julienw, line 35: I don't understand why we're doing overrides like this, instead of defining them directly in the component itself.
I only see inconvenients and no advantages.

Inconvenients:
* it's necessary to look in 2 different files to find the overrides.
* from CompareOverTime or CompareWithBase, it's not obvious that overrides exist at all.
* we need to introduce CSS classes that are used just for this. Later on if we remove these styles, the classes will probably stay on the component for no good reason. (that wouldn't be a big deal though).

Advantages:
* the advantage of adding things here is usually (IMO) when we _share_ styles. But that's not the case here as these styles are overrides, they're not shared. So this advantage doesn't apply for these styles.

I know that a lot of other styles are similar, but I think this should be changed eventually. Let's try to not add more this way... But please tell me what you think, especially if I'm missing something!

Additional question: why do you have this `-12` for compare over time? Can we use another constant somewhere? Or at the very least please add a comment about it to explain the reason.

src/components/Search/SearchOverTime.tsx
```diff
@@ -0,0 +1,178 @@
...
+      document.removeEventListener('mousedown', handleDocumentMousedown);
+    };
+  }, [handleDocumentMousedown]);
+
+  useEffect(() => {
+    document.addEventListener('keydown', handleEscKeypress);
+    return () => {
+      document.removeEventListener('keydown', handleEscKeypress);
+    };
+  });
```
julienw, line 106: If we want to run this at mount time, we need to pass an empty dependency list to `useEffect`. Otherwise this will be run at every render.

Even more correct would be to use `handleEscKeypress` as dependency, but then you'll need to use `useCallback` for it too, with an empty dependency list.
Or define `handleEscKeypress` directly inside the effect (like suggested above for `handleDocumentMousedown`.

src/components/Search/SearchOverTime.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+interface SearchProps {
+  selectLabel: string;
+  searchResults: RevisionsList[];
+  tooltip: string;
+  inputPlaceholder: string;
+  isEditable: boolean;
+}
+
+function SearchComponent({
```
julienw, line 31: nit: call this `SearchOverTime`

You can also directly write `export default function SearchOverTime` instead of adding the export at the end of the file. It makes it obvious that this is the exported function.

(I personally prefer named (aka non-default) exports for easier searching (because the exported and imported names are guaranteed to be the same), but because all other components are using default exports, let's keep it this way for now)

src/components/Search/SearchOverTime.tsx
```diff
@@ -0,0 +1,178 @@
...
+                ...(mode === 'light'
+                  ? DropDownItemRaw.Light
+                  : DropDownItemRaw.Dark),
+              },
+            },
+          },
+        },
+      },
+    },
+  });
```
julienw, line 71: I'm sure there are better ways ot handle that (for example MUI's `styleOverrides`), because contrary to the other override in  `styles`, this override here should probably be applied more broadly.
I'm not asking to solve this now, but it would be good to spend some time on this question later on.

Especially I'm not comfortable that we add these globally generic rules inside a component, and they can apply in code that is used in other components.

I'd love to hear what you think, because I'm sure you know more than me about this library.

src/components/Search/SearchOverTime.tsx
```diff
@@ -0,0 +1,178 @@
...
+      setDisplayDropdown(false);
+    }
+  };
+
+  useEffect(() => {
+    document.addEventListener('mousedown', handleDocumentMousedown);
+    return () => {
+      document.removeEventListener('mousedown', handleDocumentMousedown);
+    };
+  }, [handleDocumentMousedown]);
```
julienw, line 99: optional:
you could define `handleDocumentMousedown` directly in the effect, and use `displayDropdown` as the dependency for the effect. This would "group" together the event handler with its user.

src/components/Search/CompareOverTime.tsx
```diff
@@ -1,25 +1,106 @@
...
-        className='compare-card-img compare-card-img--time'
-        aria-label='a clock'
-      />
-    </div>
+        className={`compare-card-container content-base content-base--${
+          expanded ? 'expanded' : 'hidden'
+        } ${styles.container} `}
+      >
+        <Divider className='divider' />
+        <Form action='/compare-results' className='form-wrapper'>
```
julienw, line 86: What do you think about adding an aria-label to the form, so that you can target it more easily in tests and improve accessibility?

src/components/Search/CompareOverTime.tsx
```diff
@@ -1,25 +1,106 @@
...
-        <div className='compare-card-tagline'>{strings.overTime.tagline}</div>
+      <div
+        className={`compare-card-container compare-card-container--${
+          expanded ? 'expanded' : 'hidden'
+        } ${styles.container}`}
+        onClick={toggleIsExpanded}
+        data-testid='time-state'
+      >
+        <div className={`compare-card-text ${styles.cardText}`}>
+          <div className='compare-card-title'>{strings.overTime.title}</div>
```
julienw, line 72: Random thought for accessibility: this should be a header
=> https://mozilla-hub.atlassian.net/browse/PCF-392

(but I'm fine if we change both of them later)

julienw (changes requested):
Thanks for this work, this seems to work perfectly!

I left a few comments, especially about how the CSS is defined, but also a few other issues, including some things to do later, but that I'd like your input on.

Please add the new changes to one or more extra commits!

src/styles/CompareCards.ts
Carla-Moz: Yes, it makes sense to keep unique styles within the components. I don't mind doing that from now on. No prob. I'll make a spacing constant for 12. Not sure why it didn't exist before.

src/components/Search/SearchOverTime.tsx
Carla-Moz: Yes, thanks for catching that typo.

src/components/Search/CompareOverTime.tsx
Carla-Moz: I can do this now. No prob.

src/components/Search/SearchOverTime.tsx
Carla-Moz: I'll create a ticket to take care of this work. It needs an update.

src/components/Search/CompareButton.tsx
```diff
@@ -22,6 +26,7 @@ export default function CompareButton() {
       className={`compare-button ${styles.button}`}
       sx={{ textTransform: 'none !important' }}
       type='submit'
+      aria-label={name}
```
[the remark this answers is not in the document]
Carla-Moz: This will change the name of the buttons. Is that ok?

src/components/Search/CompareOverTime.tsx
Carla-Moz: I've change them now since it makes the tests easier.

Carla-Moz:
@julienw, please review the changes. I'll handle the style overrides refactor in a separate PR. Thanks for your review!!

julienw (approved):
Looks good to me, thanks!

I left a couple final comments before merging, tell me what you think!

src/components/Search/CompareOverTime.tsx
```diff
@@ -45,6 +43,12 @@ function CompareOverTime({ isEditable }: CompareWithTimeProps) {
     }),
   };
 
+  const wrapperStyles = {
+    wrapper: style({
+      marginBottom: `${Spacing.layoutXLarge - Spacing.xMedium}px`,
```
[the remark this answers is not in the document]
Carla-Moz: Just wanted to keep the spacing consistent with the design, but I don't mind removing it.

src/styles/Spacing.ts
```diff
@@ -3,6 +3,7 @@
 export enum Spacing {
   xSmall = 4,
   Small = 8,
+  xMedium = 12,
```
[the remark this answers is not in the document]
Carla-Moz: lol yes that was a stretch. I don't mind deleting it.

src/components/Search/SearchOverTime.tsx
```diff
@@ -81,29 +78,26 @@ function SearchComponent({
...
 
-  useEffect(() => {
     document.addEventListener('mousedown', handleDocumentMousedown);
     return () => {
       document.removeEventListener('mousedown', handleDocumentMousedown);
     };
-  }, [handleDocumentMousedown]);
+  }, [displayDropdown]);
 
   useEffect(() => {
```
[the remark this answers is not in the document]
Carla-Moz: Sure, I don't see why not and I have no probs doing so.

src/components/Search/CompareButton.tsx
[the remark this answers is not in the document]
Carla-Moz: Yes, I like the updated button names too. It is clearer.

PR #617, opened 2024-03-08
Rename RevisionsList to Changeset

julienw:
This was confusing me a lot. Indeed in the code base, the term "revisionsList" was used for both a Changeset and a list of Changesets (`RevisionsList[]`).
So I think the term Changeset is more idiomatic. This is how Firefox developers call this object.
For example on hg.mozilla.org: https://hg.mozilla.org/mozilla-central/pushloghtml or https://hg.mozilla.org/integration/autoland/rev/64e14543839c59fc9fc3d71a1cc60507bb9c7064
Note that 64e... is the changeset's revision, so it's exactly the same object we have here.

### DIFF
--- a/src/__tests__/utils/fixtures.ts
+++ b/src/__tests__/utils/fixtures.ts
@@ -1,7 +1,7 @@
-import type { CompareResultsItem, RevisionsList } from '../../types/state';
+import type { CompareResultsItem, Changeset } from '../../types/state';
 
 const getTestData = () => {
-  const testData: RevisionsList[] = [
+  const testData: Changeset[] = [
     {
       id: 1,
       revision: 'coconut',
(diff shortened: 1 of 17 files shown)


Carla-Moz:
> This was confusing me a lot. Indeed in the code base, the term "revisionsList" was used for both a Changeset and a list of Changesets (`RevisionsList[]`). So I think the term Changeset is more idiomatic. This is how Firefox developers call this object. For example on hg.mozilla.org: https://hg.mozilla.org/mozilla-central/pushloghtml or https://hg.mozilla.org/integration/autoland/rev/64e14543839c59fc9fc3d71a1cc60507bb9c7064 Note that 64e... is the changeset's revision, so it's exactly the same object we have here.

Do we know whether using `Changeset `is what the devs using this tool want, however? Why was `RevisionsList` used in the first place. Was there a reason?

Carla-Moz (commented):
Do we know whether using `Changeset` is what the devs using this tool want, however? Why was `RevisionsList` used in the first place. Was there a reason?

Carla-Moz (approved):

julienw:
thanks for the review!

PR #618, opened 2024-03-08
Do not store repository information in the CompareWithBase state

julienw:
Please look at the last commit only, as the first one is #617. Also of course I can rebase and change it if we decide to use another type name than `Changeset`.

This is a small simplification of how we deal with the repository information.
Indeed we were always computing them from the items before setting the state anyway, so it's easier to compute them when and where needed. This simplifies some code.

### DIFF
--- a/src/__tests__/Search/CompareWithBase.test.tsx
+++ b/src/__tests__/Search/CompareWithBase.test.tsx
@@ -23,7 +23,7 @@ function setUpTestData() {
     })
     .get('begin:https://treeherder.mozilla.org/api/perfcompare/results/', [])
     .get(
-      'begin:https://treeherder.mozilla.org/api/project/mozilla-central/push/?revision=coconut',
+      'begin:https://treeherder.mozilla.org/api/project/try/push/?revision=coconut',
       {
         results: [testData[0]],
       },
@@ -48,7 +48,7 @@ function renderWithCompareResultsURL(component: ReactElement) {
   return renderWithRouter(component, {
     route: '/compare-results/',
     search:
-      '?baseRev=coconut&baseRepo=mozilla-central&newRev=spam&newRepo=mozilla-central&framework=2',
+      '?baseRev=coconut&baseRepo=try&newRev=spam&newRepo=mozilla-central&framework=2',
     loader,
   });
 }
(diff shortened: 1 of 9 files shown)


src/__tests__/Search/CompareWithBase.test.tsx
```diff
@@ -23,7 +23,7 @@ function setUpTestData() {
     })
     .get('begin:https://treeherder.mozilla.org/api/perfcompare/results/', [])
     .get(
-      'begin:https://treeherder.mozilla.org/api/project/mozilla-central/push/?revision=coconut',
+      'begin:https://treeherder.mozilla.org/api/project/try/push/?revision=coconut',
```
julienw, line 26: I wasn't expecting some test change initially, but actually we had an inconsistency here: indeed the "coconut" revision in the test data is from the repository "try", not the repository "mozilla-central". So I had to change this test a little bit as well as the snapshots.

src/components/CompareResults/ResultsView.tsx
```diff
@@ -19,22 +19,15 @@ interface ResultsViewProps {
 }
 function ResultsView(props: ResultsViewProps) {
   const dispatch = useAppDispatch();
-  const {
-    baseRevInfo,
-    baseRepo,
```
julienw, line 24: We don't need this anymore but I think we'll need it later, when moving the repository dropdown value out of redux. That's why I still decided to keep it returned from the loader.

src/components/Search/CompareWithBase.tsx
```diff
@@ -24,67 +23,43 @@ interface CompareWithBaseProps {
...
-  });
+  const [newStagingRevs, setStagingNewRevs] = useState<Changeset[]>(newRevs);
 
   //the edit button will initialize the "in progress" state
   //and copy "stage" to "in progress" state
   const [baseInProgress, setInProgressBase] = useState<InProgressState>({
     revs: [],
-    repos: [],
     isInProgress: false,
   });
```
julienw, line 51: I wondered about splitting this in 2 different states, but finally decided to keep it like this for now.

src/components/Search/SelectedRevisionItem.tsx
```diff
@@ -51,6 +50,7 @@ function SelectedRevisionItem({
   const revisionHash = truncateHash(item.revision);
   const commitMessage = getLatestCommitMessage(item);
   const itemDate = new Date(item.push_timestamp * 1000);
+  const repository = repoMap[item.repository_id] ?? 'try';
```
julienw, line 53: computed "just in time" :-)

julienw:
Thanks for the review!

For context, this makes things easier to use the information coming from the "loader".

PR #619, opened 2024-03-14
Split the "In progress" composite states into two different states: the revisions array, and an "In progress" boolean

julienw:
I hesitated doing it in #618, and finally I found this to be useful for my other coming patches.
Indeed I find it more convenient to be able to change the in progress revisions independently from the boolean.

There should be no difference in what the users see.


Carla-Moz (approved):
This looks good to me. It makes easier to handle the in progress boolean in other components without having to deal with the revs.

PR #620, opened 2024-03-14
Rename isEditable to hasNonEditableState

julienw:
This one is optional and mostly cosmetic, but I find it easier to reason with.

Indeed, the forms in SearchView and ResultsView are both editable! The difference is that the one is ResultsView has a "non editable" state, or a "staging" state, that the user can revert to.
Therefore I find that the term `hasNonEditableState` makes more sense.
Tell me what you think! I'm also happy to rename it or just abandoning this change.

As part of this change, I also added a pretty big comment as an explanation of the 3 states of the CompareWithBase form. I'd like to keep that one at least.

### DIFF
--- a/src/components/Search/SearchComponent.tsx
+++ b/src/components/Search/SearchComponent.tsx
@@ -30,7 +30,7 @@ import SearchResultsList from './SearchResultsList';
 import SelectedRevisions from './SelectedRevisions';
 
 interface SearchProps {
-  isEditable: boolean;
+  hasNonEditableState: boolean;
   isWarning: boolean;
   isBaseComp: boolean;
   searchResults: Changeset[];
@@ -48,7 +48,7 @@ interface SearchProps {
 }
 
 function SearchComponent({
-  isEditable,
+  hasNonEditableState,
   isBaseComp,
   searchResults,
   displayedRevisions,
@@ -89,7 +89,7 @@ function SearchComponent({
   });
 
   const [displayDropdown, setDisplayDropdown] = useState(false);
-  const [formIsDisplayed, setFormIsDisplayed] = useState(!isEditable);
+  const [formIsDisplayed, setFormIsDisplayed] = useState(!hasNonEditableState);
 
   const handleDocumentMousedown = useCallback(
     (e: MouseEvent) => {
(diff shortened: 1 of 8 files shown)


src/components/Search/SearchComponent.tsx
```diff
@@ -169,11 +169,11 @@ function SearchComponent({
...
           id={`${searchType}_search-dropdown`}
           className={`${searchType}-search-dropdown ${styles.dropDown} ${
-            isEditable ? 'small' : ''
-          } ${isEditable ? compareView : ''}-base-dropdown`}
+            hasNonEditableState ? 'small' : ''
+          } ${hasNonEditableState ? compareView : ''}-base-dropdown`}
         >
           <SearchDropdown
-            isEditable={isEditable}
+            compact={hasNonEditableState}
```
Carla-Moz, line 176: Why are you calling it `compact` here?

src/components/Search/SearchDropdown.tsx
```diff
@@ -17,19 +17,19 @@ import { fetchRecentRevisions } from '../../thunks/searchThunk';
...
 
 //handle in progress repos here if necessary
 function SearchDropdown({
-  isEditable,
+  compact,
   selectLabel,
   searchType,
 }: SearchDropdownProps) {
-  const size = isEditable == true ? 'small' : undefined;
+  const size = compact ? 'small' : undefined;
```
Carla-Moz, line 32: Ok, I see why.

Carla-Moz (approved):
I have no qualms changing it to` nonEditableState.` Also, thanks for providing an explanation for the `CompareOverBase` component.

src/components/Search/SearchDropdown.tsx
julienw: I should have provided an explanation about that in the PR.

I try to think at the component level: what is the component doing, and what does it need as parameter to drive its behavior? In this case, this property controls how it is displayed, and its compacity. That's I renamed it to `compact`. The fact that the form is editable makes no sense at this specific component level.

julienw: Of course this doesn't change the logic at all! But this is the idea of splitting the responsibilities between the components (and the props).

julienw:
Thanks for the review!

PR #624, opened 2024-03-15
Rename most callback props used in CompareWithBase so that they look like events

julienw:
This is another quite cosmetic PR!

This mostly changes how some props that take callbacks are called.
There are several goals:
1. more consistent naming across components, especially for the "onRemoveRevision" callback
2. make clear what we consider as being "events" (eg: toggle, edit, etc)


src/components/Search/SelectedRevisions.tsx
```diff
@@ -22,29 +22,22 @@ function SelectedRevisions({
...
-    removeCheckedRevision(item);
-  };
   const removeRevision = (item: Changeset) => {
     if (hasNonEditableState) {
-      onEditRemoveAction(item);
+      onRemoveRevision(item);
     } else {
-      handleRemoveRevision(item);
+      removeCheckedRevision(item);
     }
```
julienw, line 41: This will be further simplified also in a later PR.

beatrice-acasandrei (approved):

PR #626, opened 2024-03-19
Refactor how Edit mode works in the SearchView by completely remove the state checkedRevisions

julienw:


src/components/Search/CompareWithBase.tsx
```diff
@@ -159,12 +159,46 @@ function CompareWithBase({
...
 
-  const handleSearchResultsToggleNew = (toggleArray: Changeset[]) => {
-    setInProgressNewRevs(toggleArray || []);
+  const handleSearchResultsToggleNew = (item: Changeset) => {
+    const newNewRevs = handleItemToggleInChangesetList(
+      item,
+      3,
+      newInProgressRevs,
+    );
+    setInProgressNewRevs(newNewRevs);
```
esanuandra: I had to scroll up to see what are `1` and `3`.

julienw: very good remark!
you know what, I'll change `handleItemToggleInChangesetList` to take an option object, so that the parameters are clearer

julienw: I added a commit https://github.com/mozilla/perfcompare/pull/626/commits/1e7894533c5b79f22943c161813c80400986bc0d, tell me what you think!

src/components/Search/CompareWithBase.tsx
```diff
@@ -129,87 +122,85 @@ function CompareWithBase({
...
   const handleRemoveRevisionNew = (item: Changeset) => {
     const revisionsNew = [...newInProgressRevs];
     revisionsNew.splice(newInProgressRevs.indexOf(item), 1);
     setInProgressNewRevs(revisionsNew);
   };
 
-  const handleSearchResultsToggleBase = (toggleArray: Changeset[]) => {
-    setInProgressBaseRevs(toggleArray || []);
-    setInProgressBase(true);
+  const handleItemToggleInChangesetList = ({
```
julienw, line 161: It will probably be worth extracting this again, because Carla will want to reuse this in CompareOverTime. I'm not sure yet if this should be a hook or just a utility function.
I can do it after #629.

src/components/Search/CompareWithBase.tsx
```diff
@@ -70,23 +70,19 @@ function CompareWithBase({
   const [expanded, setExpanded] = useState(true);
   const { enqueueSnackbar } = useSnackbar();
 
-  //the "committed" base and new revisions initialize the staging state
+  // The "committed" base and new revisions initialize the staging state.
+  // These states are snapshots of selections, that the user can either save (by
+  // pressing the "Save" button) or revert to (by pressing the Cancel button).
```
Carla-Moz, line 74: I think these comments are somewhat confusing. See comment below.

src/components/Search/CompareWithBase.tsx
```diff
@@ -70,23 +70,19 @@ function CompareWithBase({
...
-  const [baseInProgress, setInProgressBase] = useState(false);
-
-  const [newInProgressRevs, setInProgressNewRevs] = useState<Changeset[]>([]);
-  const [newInProgress, setInProgressNew] = useState(false);
-
-  const [displayedRevisionsBaseRevs, setDisplayedRevisionsBaseRevs] =
+  // The "committed" base and new revisions initialize the "in progress" state
+  // too. These states hold the data that are displayed to the user. They always
+  // contain the displayed data. That's also the ones that will be committed if
+  // the user presses the "Compare" button.
```
Carla-Moz, line 81: I thought the "staging" states become "committed" when the user presses the "Compare" button?

Shouldn't pressing "Save" be a requirement before pressing the "Compare" button ? Unless you changed that logic too? I remembered we talked about how unnecessary that extra step was. Did you change it?

When the user presses the "Save" button, the "inprogress" states overwrite the "staging" states? 

Also, this bring up another question. Should the "Compare" button be active once the results page mounts? Wouldn't it make sense for it to be active when the user presses the "Edit" button?

Carla-Moz (commented):
Hi Julien, I had some questions regarding the comments for this commit.

src/components/Search/CompareWithBase.tsx
julienw: Thanks for the great questions! I can see where the confusion comes from.

My goal for this patch is that no behavior was to be changed. We discussed in the past that the "Staging" state was  unnecessary and I'm all for it, but wanted to defer any behavior change to after this refactoring.


> I thought the "staging" states become "committed" when the user presses the "Compare" button?
> 
> Shouldn't pressing "Save" be a requirement before pressing the "Compare" button ? Unless you changed that logic too? I remembered we talked about how unnecessary that extra step was. Did you change it?

I didn't change this in _this_ patch, but indeed I changed it in a previous patch. I think that was in #582, more specifically in the change to `SelectedRevisionItem.tsx` in https://github.com/mozilla/perfcompare/pull/582/commits/cd04c4aed944171efb71cd75908b017d8f3646c6#diff-3db985901d6838f35e9001408344d9ed36d6c3377ee8e0ccb8d0d0f38bad9c29R64: the revisions that are displayed are the ones that are submitted when the user presses `Compare`.

Now that I think about it, I should probably have added these hidden fields in `CompareWithBase` so that it would be easier to control what is actually submitted; and therefore I could have used the "staging" data instead of the "in progress" data.

So currently, the "staging" data is only used as a snapshot (at least that's how I think about it): the user can "edit", then press "cancel" to go back at the previously "save"d snapshot. Or "save" to save a snapshot. I believe it's not so useful, especially that with the router now, the user can press the browser's back button after "compare" to come back at the previous state.

I'd like to not change this behavior in this patch, but for the near future there are IMO 2 possible paths (both of them in another patch though):

1. I'd be fine reverting this to the previous approach (i.e. "Staging" is used when the user presses "Compare"), to stay faithful to my previous goal "do no change behavior just yet", so that we can discuss this further.
2. Remove both the staging state and the "save" button (keeping Edit button though).

My preference is solution 2.

> 
> When the user presses the "Save" button, the "inprogress" states overwrite the "staging" states?

Yes!

> 
> Also, this bring up another question. Should the "Compare" button be active once the results page mounts? Wouldn't it make sense for it to be active when the user presses the "Edit" button?

Yeah, definitely, it makes sense that it is active only after the user did at least one edit.

I can file a JIRA ticket for this and the previous suggestion, actually I believe it makes sense that they are both grouped in one patch.

Tell me what you prefer!

Carla-Moz: This sounds good for now. Yes, I agree. Let's defer any new changes to a new patch so it's clearer. Thank you!

Carla-Moz: I'm also drawn to solution 2.

Carla-Moz (approved):

src/components/Search/CompareWithBase.tsx
```diff
@@ -99,7 +99,7 @@ function CompareWithBase({
     (baseRepository !== 'try' && newRepository === 'try');
 
   const possiblyPreventFormSubmission = (e: React.FormEvent) => {
-    const isFormReadyToBeSubmitted = baseRevs.length > 0;
+    const isFormReadyToBeSubmitted = baseInProgressRevs.length > 0;
```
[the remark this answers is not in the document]
julienw: Great questions!

> The "staging" states in SearchView are pretty much useless, right?

Yes, because there's no edit mode, so that's just "in progress" really. There's even no "commited" state because when clicking "compare" the user goes to the compare results page.

> We only use the "staging" states in regards to the "Cancel" action.

Yes that's true!

> Why not use the "commited" state values instead of the "staging" state values to overwrite the "inprogress" state when the user presses the "Cancel" button? This way, I don't think we need the "staging" states. What do you think?

Think of these steps:

1. The user presses edit.
2. They change the revisions.
3. They press "save".
4. They press edit again.
5. They change the revisions.
6. They press "cancel".

=> What revisions should be displayed? What does the user expect? The ones from before step 1, or the ones from after step 3?
I think (in the current situation) that it should the ones after step 3. That's why we have the staging state.
Does that make sense?

_But_ we seem to agree that we don't need all this complexity and we'll get rid of this "save" and this staging state anyway :-)

julienw:
Thanks for the review!

PR #629, opened 2024-03-25
Fix that the wrong item is unselected in some situations

julienw:
Please look at the last commit only.

Here is the STR:
1. Go to https://beta--mozilla-perfcompare.netlify.app/
2. Pick a base rev, and 3 "new revs".
3. Press "Compare"
4. Reload the page (important)
5. Edit the new revs
6. Try to uncheck the first one from the search results list.

=> result: the last one is unchecked instead of the first one

When unchecking one element, the executed operation to find the index is:
```js
revs.indexOf(item)
```
Where `revs` contains the currently selected items, and `item` is the item we just clicked. 
The problem is that the object `item` isn't inside `revs`... an identical object is present, but not the exact same object with the same reference. As a result `indexOf` doesn't find the object... and returns `-1`.
Then `-1` is passed to `splice`, which helpfully removes... the last element. Tada.

Instead I changed the logic to use `findIndex` with a callback that compares the revision id instead. Then this fixes the problem accordingly.

This may be a regression from #582.

I didn't do a test for this because I thought this is an edge case. But I can do one if you prefer.

### DIFF
--- a/src/components/Search/CompareWithBase.tsx
+++ b/src/components/Search/CompareWithBase.tsx
@@ -148,13 +148,25 @@ function CompareWithBase({
   };
 
   const handleRemoveRevisionBase = (item: Changeset) => {
+    // Currently item seems to be the same object than the one stored in
+    // baseInProgressRevs, but it might change in the future. That's why we're
+    // comparing the ids instead of using indexOf directly.
+    const indexInBaseChangesets = baseInProgressRevs.findIndex(
+      (rev) => rev.id === item.id,
+    );
     const revisionsBase = [...baseInProgressRevs];
-    revisionsBase.splice(baseInProgressRevs.indexOf(item), 1);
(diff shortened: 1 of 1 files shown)


src/components/Search/CompareWithBase.tsx
```diff
@@ -148,13 +148,25 @@ function CompareWithBase({
   };
 
   const handleRemoveRevisionBase = (item: Changeset) => {
```
beatrice-acasandrei, line 150: Wondering if this logic can be simplified for removing the base revision, since we only allow one.

julienw: yeah, I believe we can have a behavior that's more like radio buttons: instead of warning we could just unselect the old one and select the new one...

beatrice-acasandrei (approved):

PR #631, opened 2024-03-28
Fix lint command

esanuandra:


package.json
```diff
@@ -35,7 +35,7 @@
     "dev": "webpack serve --config webpack/webpack.config.js --env env=development --open",
     "start": "node server.js",
     "build": "webpack --config webpack/webpack.config.js --env env=production",
-    "lint": "eslint src/**/*.{js,ts,tsx}",
+    "lint": "eslint 'src/**/*.{js,ts,tsx}'",
```
julienw, line 38: There's a subtle issue: on windows single quotes don't work as expected.
Therefore I suggest this instead:
```suggestion
    "lint": "eslint \"src/**/*.{js,ts,tsx}\"",
```

(we do a similar thing in the profiler for stylelint... but not eslint curiously)

julienw (approved):
Interesting! For me the previous version works too, but let's do this

package.json
esanuandra: done

PR #633, opened 2024-04-02
Move the repository information from Redux to a local state in CompareWithBase

julienw:
The goal is to make `SearchContainer` and all other components in the tree more self-contained that they were before.

This also moves SearchViewInit to CompareWithBase. But this is only
temporary because this component will be removed later.

### DIFF
--- a/src/components/Search/CompareWithBase.tsx
+++ b/src/components/Search/CompareWithBase.tsx
@@ -10,10 +10,11 @@ import { style } from 'typestyle';
 import { useAppSelector } from '../../hooks/app';
 import { Strings } from '../../resources/Strings';
 import { CompareCardsStyles, SearchStyles, Spacing } from '../../styles';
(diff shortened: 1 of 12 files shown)


src/components/Search/CompareWithBase.tsx
```diff
@@ -201,6 +206,10 @@ function CompareWithBase({
 
   return (
     <Grid className={`wrapper--withbase ${wrapperStyles.wrapper}`}>
+      <SearchViewInit
+        repositoryBase={baseRepository}
+        repositoryNew={newRepository}
+      />
```
julienw, line 212: I moved it here because this needs the repository information. But in a future patch this component will be removed.

julienw:
Rebased on top of latest beta branch!

Carla-Moz (approved):
LGTM and makes sense. Thanks!

PR #635, opened 2024-04-02
 Extract a component `SearchInputAndResults` for the `SearchInput` and its dropdown 

julienw:
Extracting these 2 components into a separate comment makes sense in terms of component responsibilities. Indeed the SearchInput controls its Dropdown. Using a component to contain both makes it possible for this component to keep all the necessary state to control this behavior This is also common code for both `SearchOverTime` and `SearchView`, so it makes sense to extract it.

Please look at the two last commits only.
Commit 1: this extracts the common code to a separate component
Commit 2: this changes this component to make it more generic, so that it works the same for both CompareWithBase and CompareOverTime, without relying on class names. There are a bunch of snapshot changes because this adds a new container to the pair `SearchInput` / `SearchResultsList`, if you look at them by disabling whitespace changes it is much easier to see these changes.

### DIFF
--- a/src/__tests__/Search/__snapshots__/CompareOverTime.test.tsx.snap
+++ b/src/__tests__/Search/__snapshots__/CompareOverTime.test.tsx.snap
@@ -85,51 +85,55 @@ exports[`Compare Over Time renders correctly in Search View: Initial state for t
         id="new_search-input--time"
       >
         <div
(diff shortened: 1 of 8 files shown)


Carla-Moz (commented):
This also makes the `SearchComponent `and `SearchOverTime `easier to read and understand. Thanks!

PR #636, opened 2024-04-08
For the base revision, clicking another item will select it and unselect the previous one

julienw:
This was suggested by @beatrice-acasandrei in https://github.com/mozilla/perfcompare/pull/629#discussion_r1544456012 and I found this is a good idea.

With this patch, the logic for the base revision toglging is changed compared to the one for the new revisions. For the base revision, now if the user selects a different item, we'll automatically unselect the previous one instead of displaying an error.

Thanks to this change and that the logic is now different, I could also simplify the logic for the new revision toggle handling. Especially the function `handleSearchResultsToggleBase` doesn't call `enqueueSnackbar` anymore, and is now a simple function without a side effect, and so it could easily be extracted to be reused for `CompareOverTime` as a simple function to a simple module, and we don't need all the react hooks mechanism.

@Carla-Moz @beatrice-acasandrei @esanuandra please tell me what you think of this new behavior!


julienw:
The missing coverage is on some code branch that shouldn't happen, but that I prefered to have a if condition for. So I can't test it (hopefully :D).

beatrice-acasandrei (approved):

Carla-Moz (approved):
LGTM! Thanks

PR #637, opened 2024-04-11
Compare overtime time dropdown

Carla-Moz:
This PR fulfills the time range dropdown task in the Compare over Time epic. 
Please see [Time range dropdown](https://mozilla-hub.atlassian.net/browse/PCF-306)

This PR also includes a minor change fixing the repository dropdown labels for the Revisions which was incorrectly named "Base". I did this while adding coverage for the time range dropdown.


src/components/Search/TimeRangeDropdown.tsx
```diff
@@ -0,0 +1,129 @@
...
+          labelId='select-timerange-label'
+          className='dropdown-select'
+          variant='standard'
+          onChange={(e) => void handleTimeRangeSelect(e)}
+          name='timerange'
+        >
+          {Object.entries(timeRangeMap).map(([value, text]) => (
+            <MenuItem
+              value={value}
+              key={text}
```
julienw, line 117: optional nit: I'd use the `value` as the key. But I don't think this matters much really in this case.

src/components/Search/SearchOverTime.tsx
```diff
@@ -72,12 +72,9 @@ export default function SearchOverTime({
       <Grid
         item
         xs={2}
-        className={`$new-search-dropdown ${styles.dropDown} label-edit-wrapper`}
+        className={`new-search-dropdown ${styles.dropDown} label-edit-wrapper`}
```
julienw, line 75: 😁

src/components/Search/SearchOverTime.tsx
```diff
@@ -72,12 +72,9 @@ export default function SearchOverTime({
...
       >
-        <InputLabel
-          id='select-repository-label--time'
-          className='dropdown-select-label'
-        >
+        <InputLabel id={selectLabel} className='dropdown-select-label'>
```
julienw, line 77: It doesn't seem correct to use the label as the id. What happens if there are spaces for example? 

Do we really need an id though? Can we just remove it?

src/components/Search/SearchComponent.tsx
```diff
@@ -94,10 +94,7 @@ function SearchComponent({
           styles.dropDown
         } label-edit-wrapper`}
       >
-        <InputLabel
-          id='select-repository-label'
-          className='dropdown-select-label'
-        >
+        <InputLabel id={selectLabel} className='dropdown-select-label'>
```
julienw, line 97: Same remark as below about the id. In this was this works because this is just one word, but if we change it in the future this might lead to problems.
I understand you wanted to make it unique. But maybe we can just remove it. Do we use it anywhere?
What do you think?

julienw (commented):
This looks good to me!
My one concern is how we store the new state. I've seen you used Redux, but I think this is taking the wrong trend, that is we try to remove from Redux the things that are not used throughout the app, and _especially_ the things that will come from the URL in the results view. Indeed synchronizing redux with the URL (and vice versa) is painful.

Therefore I think it's better, following the trend we try to take here, to not use Redux to store this state. Can you please change this to store the time range value in CompareOverTime instead?

Apart from that this looks reasonable to me!
Thanks!

src/components/Search/TimeRangeDropdown.tsx
```diff
@@ -0,0 +1,129 @@
...
+          ...(mode === 'light'
+            ? ButtonsLightRaw.Dropdown
+            : ButtonsDarkRaw.Dropdown),
+        },
+      },
+    }),
+  };
+
+  const dispatch = useAppDispatch();
+  const timeRangeValue = useAppSelector((state) => state.timeRange.value);
```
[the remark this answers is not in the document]
Carla-Moz: Is there a ticket to move framework information to CompareWithBase? I didn't know we wanted to remove everything out of redux. Yes, I'll move it to CompareOverTime. Shouldn't be a problem, same with framework.

src/components/Search/SearchOverTime.tsx
Carla-Moz: I'll create a unique id name aside from the select label to account for the possibility of spaces in the name. And yes, we want an` id` for the label to match the select `labelID.` Please see https://mui.com/material-ui/react-select/

src/components/Search/SearchComponent.tsx
Carla-Moz: Same reply as above.

PR #640, opened 2024-04-26
Compare overtime checked and displayed revisions

Carla-Moz:
This PR fulfills the requirements of [Handle checked and displayed selected revisions]( https://mozilla-hub.atlassian.net/browse/PCF-305). This is the last remaining part of the Compare over time component. The hash to hide it is removed.  The "Compare" button will remain disabled ([_see this commit_](https://github.com/mozilla/perfcompare/pull/640/commits/1b8119c055392f093f066e4eade248a69ac310cb)) until the next PR to add react-router config to fetch the data. 

**New File Created:** In this [_commit_](https://github.com/mozilla/perfcompare/pull/640/commits/bac019b1ff83b13d953d73d51366eb1453f69938), I created a new file called `useSearchResults` because both the Revisions component in "With base" and "Overtime" handle the search results toggle the same way, including showing the error when the user attempts to select more than 3 revisions. This required updating the `CompareWithBase` file. 

**Style Tweaks:** In this [_commit_](https://github.com/mozilla/perfcompare/pull/640/commits/a83b8f27c675cd09487bccf23a48f01788367908),  I did a small update on `styles/SearchContainerStyles `and `styles/Documents` to better reflect the overflow of the search results dropdown.

Update test coverage in `CompareOvertTime.test:` please check this [_commit_](https://github.com/mozilla/perfcompare/pull/640/commits/a83b8f27c675cd09487bccf23a48f01788367908). 

_**Side note:**_ I talked to Dasha about toggling the display of the "Compare with Base" and "Compare Over time" component in which clicking on one hides the content below their respective headers so they function like radio buttons. This is not handled in this PR but something I think would be helpful to do in a future PR.


julienw:
> Side note: I talked to Dasha about toggling the display of the "Compare with Base" and "Compare Over time" component in which clicking on one hides the content below their respective headers so they function like radio buttons. This is not handled in this PR but something I think would be helpful to do in a future PR.

Yeah I don't like the current situation much, happy to improve it in any way.

src/__tests__/Search/CompareOverTime.test.tsx
```diff
@@ -194,4 +185,63 @@ describe('Compare Over Time', () => {
...
+    const checkbox = (await screen.findAllByTestId('checkbox-0'))[0];
+    await user.click(checkbox);
+    expect(checkbox).toHaveClass('Mui-checked');
+    const removeButton = document.querySelectorAll('[title="remove revision"]');
+    expect(removeButton[0]).toBeInTheDocument();
+    expect(screen.getAllByTestId('selected-rev-item')[0]).toBeInTheDocument();
+
+    await user.click(removeButton[0]);
+
+    expect(document.body).toMatchSnapshot();
```
julienw, line 207: optional: Maybe it would be good to run a snapshot for just the compare over time component. This would give smaller snapshots, probably faster tests, and easier to review snapshots.

julienw (changes requested):
This looks mostly good to me! The behavior is right, and the code looks reasonable.
I'm just not sure about this new hook file, see below for more information about that.

I'd like to have a double look after the changes, that's why I've put "request changes", but nothing crazy!

src/components/Search/CompareButton.tsx
```diff
@@ -23,6 +27,7 @@ export default function CompareButton({ label }: CompareButtonProps) {
       className={`compare-button ${styles.button}`}
       sx={{ textTransform: 'none !important' }}
       type='submit'
+      disabled={isDisabled}
```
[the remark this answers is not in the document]
Carla-Moz: This is only temporary. I plan to enable it after the functionality is complete to fetch the data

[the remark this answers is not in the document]
Carla-Moz: But thanks for the resources

src/styles/Document.ts
```diff
@@ -4,3 +4,7 @@ cssRule('html, body', {
   margin: 0,
   padding: 0,
 });
+
+cssRule('#root', {
+  overflow: 'hidden',
+});
```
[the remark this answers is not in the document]
Carla-Moz: It's to stop the unintended vertical overflow on the root div for the home page. I'll add a comment.

src/hooks/useSearchResults.ts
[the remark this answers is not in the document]
Carla-Moz: If it's overkill, we can repeat the code in both files. I'm not tied to creating a separate file for it.

src/styles/CompareCards.ts
```diff
@@ -66,16 +66,17 @@ export const CompareCardsStyles = (mode: string) => {
           boxShadow: '0px 2px 4px rgba(58, 57, 68, 0.2)',
         },
         '&.content-base': {
+          visibility: 'hidden',
           minHeight: '0',
           height: '0',
           flexWrap: 'nowrap',
-          overflow: 'hidden',
```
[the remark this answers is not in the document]
Carla-Moz: Yes, I try not to do style updates but it's related to the Time Component and it was small so I didn't think it would be too difficult to review.

src/styles/CompareCards.ts
```diff
@@ -66,16 +66,17 @@ export const CompareCardsStyles = (mode: string) => {
           boxShadow: '0px 2px 4px rgba(58, 57, 68, 0.2)',
         },
         '&.content-base': {
+          visibility: 'hidden',
           minHeight: '0',
           height: '0',
```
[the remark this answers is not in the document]
Carla-Moz: No, this is part of a smooth animation for the dropdowns rather than just appearing and disappearing. I hadn't finished it. But I'll cover this in a separate PR. So let's leave as is for now.

src/__tests__/Search/CompareOverTime.test.tsx
```diff
@@ -204,7 +205,7 @@ describe('Compare Over Time', () => {
 
     await user.click(removeButton[0]);
 
-    expect(document.body).toMatchSnapshot();
+    expect(formElement).toMatchSnapshot();
```
julienw, line 208: 👍

src/components/Search/CompareWithBase.tsx
```diff
@@ -161,6 +155,34 @@ function CompareWithBase({
     setInProgressNewRevs(newStagingRevs);
   };
 
+  const handleItemToggleInChangesetList = ({
```
julienw, line 158: hey carla, sorry if I wasn't clear in my comment.

What I was refering to "overkill" was specifically the use of a hook. But I started the comment with :

> Extracting handleItemToggleInChangesetList and handleRemoveNewRevision is a good idea because they're quite complex functions and duplicating them doesn't feel right.

My suggestion was to extract them to a separate file as normal function.

But duplicating them is not too bad either.
Some people say that duplicating only twice is OK, and only from 3 times they start to think about reducing the duplication.
"The rule of three" => https://stackoverflow.com/a/2298357/5169882 
The rationale is that (as I said also in other occasions) keeping the related code in the same file makes it easier to read, while moving it elsewhere requires more hops to understand what's going on (hence my focus on naming things correctly if we do that).

So.... bottom line is, let's keep it that way. But if we ever need to copy paste this again, we'll move them out to another place.

src/styles/Document.ts
```diff
@@ -4,3 +4,8 @@ cssRule('html, body', {
   margin: 0,
   padding: 0,
 });
+
+//to prevent the vertical overflow of the root element
```
julienw, line 8: The comment you added paraphrase the CSS rule, so it's not super useful in its current shape.
A useful comment will have  a _why_ or _when_. Can you add in which case this happens? What is the result to the user if we don't have this rule?

I sincerely don't understand why we need this :-) When I remove it I don't see a difference.

src/components/Search/CompareOverTime.tsx
```diff
@@ -21,32 +21,29 @@ const stringsNew =
...
-  //temporary hash to hide the component until functionality is complete
-  const location = useLocation();
-  const hash = location.hash;
-
-  //temporary styles to hide the component based on hash
-  const containerStyles = {
-    container: style({
-      display: hash === '#comparetime' ? 'block' : 'none',
-    }),
-  };
```
julienw, line 49: should we keep this toggle until the loader is done?

julienw (approved):
Thanks, this looks good to me.

2 small points:
* I still don't understand why we need overflow: hidden on the root.
* I believe we should keep the toggle still, because the results page isn't ready yet.

src/components/Search/CompareWithBase.tsx
Carla-Moz: Okay, I hear you. Yes, let's leave it for now and I'll extract to a new file within the Search folder if it comes up again.

src/styles/Document.ts
Carla-Moz: Screenshot 2024-05-02 at 08 50 40

Honestly, it's just reducing an excess of vertical flow to the page (the scroll bar responds appropriately, scrolling further down). I know that's repeating myself but I'm not sure how else to say it. But if I can't convince you, I can remove it and just update the body background color, etc. I don't know why it ended up this gray color in the first place.

src/components/Search/CompareOverTime.tsx
Carla-Moz: Okay, I can leave it and remove the disable rule to the button until the loader part is done.

PR #642, opened 2024-04-29
Remove the recent revisions state from redux

julienw:
The goal for this patch is to remove the recent revisions state from redux, replacing it by a local state in `SearchInputAndResults.tsx`.

This also removes the hook `useHandleChangeSearch` and the thunk `searchThunk` and instead moved most of their code directly to `SearchInputAndResults`.

The debouncing behavior was made somewhat more generic and extracted to a separate helper file. I decided to keep this super simple like it was before, so I didn't use an external dependency and implemented it simply.

Note that some of the code in `searchThunk` was already implemented in `logic/treeherder` in a previous patch so I could reuse it directly.

This also resolves a bug with the recent revisions handling between the Compare With Base and the Compare Over Time forms:
1. Changing the repository `to mozilla-central` in the `CompareOverTime` form
2. Focus the "new" input in `CompareWithBase` (the "repository" dropdown is still `try` in `CompareWithBase`)
3. Notice the revisions for `mozilla-central` are displayed.

The issue was that we were using the same state for both inputs. This is solved by this patch thanks to the separate state.

You can look at the commits separately, they should be fairly straightforward when taken one by one.


src/components/Search/SearchInputAndResults.tsx
```diff
@@ -51,6 +53,49 @@ export default function SearchInputAndResults({
...
+        typeof e === 'string'
+          ? e
+          : e instanceof Error
+          ? e.message
+          : `Unknown error: ${String(e)}`;
+      setSearchError(strError || 'An error has occurred');
+      setRecentRevisions(null);
+    }
+  };
+
```
Carla-Moz, line 122: Makes a lot of sense to move this fetch work in the SearchInputAndResults file.

Carla-Moz (approved):
This was pretty amazing work and quite easy to follow. Thanks for breaking up the patch into easy-to-follow commit as you did. I'll learn to apply this myself in future PRs to make it easier for the reviewers. Aside from that, I don't see anything requiring any changes. Thank you!

julienw:
Thanks for the review!

PR #648, opened 2024-05-13
PCF-400 Query https://firefox-ci-tc.services.mozilla.com/login/oauth/credentials to verify token and request credentials

esanuandra:
This PR fetches the Taskcluster credentials (with an access token that has expiration date) using access token Bearer.


julienw (commented):
hey andra, thanks for this work!
here are a few comments, I hope they make sense.

In addition to the comments on the code itself, there's another missing bit IMO.

Here is the current workflow:
1. The user is in tab A and clicks on the "login" button in the modal
2. open a new tab B to taskcluster
3. in tab B the user logs in on task cluster
4. in tab B taskcluster redirects on our callback
5. in tab B perfcompare requests and retrieves the access token to the taskcluster server
6. in tab B perfcompare requests and retrieves the user credentials to the taskcluster server
7. in tab B perfcompare stores the token to localStorage
8. ... the missing link

For step 8, we can add an event listener on the "storage" event in tab A so that we're notified with the user credentials change. We can add it in step 1.
(see MDN: https://developer.mozilla.org/en-US/docs/Web/API/Window/storage_event)

I'm thinking that if step 5 stores the access token to the localStorage, then step 6-7 could be done in tab A... it might be a better design actually, because (as I write below in another comment) in the future I'd like that we don't need to open a tab if we already have an unexpired access token.

So here is how the workflow could look like (bold for changes):
1. The user is in tab A and clicks on the "login" button in the modal.
2. **adds an event listener for "storage"**, and opens a new tab B to taskcluster
3. in tab B the user logs in on task cluster
4. in tab B taskcluster redirects on our callback
5. in tab B perfcompare requests and retrieves the access token to the taskcluster server
6. **in tab B perfcompare stores the access token to localStorage, then closes the tab with `window.close`**
7. **in tab A, perfcompare is notified from the "storage" event.
8. **in tab A** perfcompare requests and retrieves the user credentials to the taskcluster server
9. **in tab A** perfcompare stores the token to localStorage
10. **in tab A** perfcompare does the retrigger request

I don't request to do this workflow change in this PR, but it could be good to do it in a next PR, if you don't mind.

Tell me what you think (and if I'm missing anything)!

src/types/types.ts
```diff
@@ -297,3 +297,8 @@ export type UserCredentials = Record<
   string,
   { expires: string; credentials: { clientId: string; accessToken: string } }
 >;
+
+export type TokenBearer = Record<
+  string,
+  { access_token: string; token_type: 'Bearer' }
+>;
```
julienw, line 304: I think these 2 types are incorrect, they're not `Record<string, XXX>`, but directly `XXX`.
```js
export type UserCredentials = { expires: string; credentials: { clientId: string; accessToken: string } };

export type TokenBearer = { access_token: string; token_type: 'Bearer' };
```

A `Record` is like a Map (but using an object), also called a dictionary. This isn't what it is here.

esanuandra: I just noticed that the Treeherder implementation stores the userCredentials in this way using the url.
```
userCredentials           "https://firefox-ci-tc.services.mozilla.com": {
                                       expires: ... 
                                       credentials: {
                                              clientId: ...
                                              accessToken: ...
                                       }
                           }
```

I believe that's why the Record for userCredentials. Should we still go with removing the Record for userCredentials?

src/types/types.ts
```diff
@@ -297,3 +297,13 @@ export type UserCredentials = Record<
   string,
   { expires: string; credentials: { clientId: string; accessToken: string } }
```
julienw, line 298: nit: reuse CredentialsReponse here

src/logic/token-storage.ts
```diff
@@ -0,0 +1,11 @@
+import { TokenBearer } from '../types/types';
+
+export function storeTokenBearer(token: TokenBearer) {
```
julienw, line 3: nit: if we use a record for credentials-storage, we should use a Record here as well.

src/logic/taskcluster.ts
```diff
@@ -58,6 +62,7 @@ const getAuthCode = (tcParams: {
 export const checkTaskclusterCredentials = () => {
   const taskclusterParams = getTaskclusterParams();
   const locationOrigin = getLocationOrigin();
+  console.log(taskclusterParams);
```
julienw, line 65: nit: remove this log

src/logic/token-storage.ts
julienw, line 1: I think i'd move this code to credentials-storage.ts, having 2 different files sounds a bit cumbersome. What do you think?

src/__tests__/TaskclusterAuth/TaskclusterCallback.test.tsx
```diff
@@ -29,4 +34,42 @@ describe('Taskcluster Callback', () => {
...
+    renderWithRouter(<TaskclusterCallback />, {
+      route: '/taskcluster-auth',
+      search: '?code=dwcygG5HQNaLiRe3RcTCbQ&state=OkCrH5isZncYqeJbRDelN',
+      loader,
+    });
+    mockedGetLocationOrigin.mockImplementation(() => 'http://localhost:3000');
+
+    expect(
+      await screen.findByText(/Getting Taskcluster credentials/),
+    ).toBeInTheDocument();
```
julienw, line 73: You should probably check if the things you expect are in storage.
and also if the fetch have been called as you expect (`expect(window.fetch).toHaveBeenCalledWith(...)`)

src/__tests__/TaskclusterAuth/TaskclusterCallback.test.tsx
```diff
@@ -29,4 +34,42 @@ describe('Taskcluster Callback', () => {
...
+      'taskclusterUrl',
+      'https://firefox-ci-tc.services.mozilla.com',
+    );
+
+    renderWithRouter(<TaskclusterCallback />, {
+      route: '/taskcluster-auth',
+      search: '?code=dwcygG5HQNaLiRe3RcTCbQ&state=OkCrH5isZncYqeJbRDelN',
+      loader,
+    });
+    mockedGetLocationOrigin.mockImplementation(() => 'http://localhost:3000');
```
julienw, line 69: nit: please move this mock setup before the render call, to avoid possible future issues.

julienw (commented):
Thanks!
here are a few more comments to fix before landing, thanks again for your work Andra

src/components/TaskclusterAuth/loader.ts
```diff
@@ -20,11 +28,22 @@ export function loader({ request }: { request: Request }) {
     );
   }
 
-  const tokenBearer = retrieveTaskclusterToken(rootUrl, taskclusterCode);
+  const tokenResponse = await retrieveTaskclusterToken(
+    rootUrl,
+    taskclusterCode,
+  );
+
+  storeToken({ [tokenResponse.token_type]: tokenResponse });
```
julienw, line 36: This should look like this:
```suggestion
  storeToken({ [rootUrl]: tokenResponse });
```

but please look at my other comments as well, as I believe the signature will be changed and therefore this will look like this:
```suggestion
  storeToken(rootUrl, tokenResponse);
```

src/logic/credentials-storage.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function retrieveToken(): TokenBearer {
+  return JSON.parse(
+    localStorage.getItem('tokenBearer') as string,
+  ) as TokenBearer;
+}
+
+export function storeUserCredentials(credentials: UserCredentials) {
+  localStorage.setItem('userCredentials', JSON.stringify(credentials));
+}
```
julienw, line 16: The goal for storing a Record is that we can add separate entries for different URLs. Therefore I believe the implementation should probably look lile this:
```suggestion
export function storeUserCredentials(rootUrl: string, credentials: UserCredentialsResponse) {
  const allCredentialsAsString = localStorage.userCredentials;
  const allCredentials = (allCredentialsAsString ? JSON.parse(allCredentialsAsString) : {}) as UserCredentials;
  allCredentials[rootUrl] = credentials;
  localStorage.userCredentials = JSON.stringify(allCredentials);
}
```

Similar for for storeToken.

src/logic/credentials-storage.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function retrieveUserCredentials(): UserCredentials {
+  return JSON.parse(
+    localStorage.getItem('userCredentials') as string,
+  ) as UserCredentials;
+}
```
julienw, line 22: Likewise, the retrieve functions could look lile this:
```suggestion
export function retrieveUserCredentials(rootUrl: string): UserCredentialsResponse | null {
  const allCredentialsAsString = localStorage.userCredentials;
  if (!allCredentialsAsString) {
    return null;
  }
  
  const allCredentials =  JSON.parse(allCredentialsAsString) as UserCredentials;
  return allCredentials[rootUrl] ?? null;
}
```

and similarly for the other retrieve function.

julienw (commented):
Looking at this with fresh eyes, here are some last minute changes, sorry about that.

Also have you seen my longer comment in https://github.com/mozilla/perfcompare/pull/648#pullrequestreview-2071055291 ? Any comment about that? Thanks

src/types/types.ts
```diff
@@ -293,7 +293,20 @@ export type FakeCommitHash =
...
+export type TokenBearerResponse = {
+  access_token: string;
+  token_type: 'Bearer';
+};
+
+export type TokenBearer = Record<string, TokenBearerResponse>;
+
+export type TokenResponse = TokenBearerResponse;
+
+export type UserToken = Record<string, TokenResponse>;
```
julienw, line 312: it looks like you have the same types twice:
TokenBearerResponse == TokenResponse
TokenBearer == UserToken

Please pick just one pair (I'd suggest TokenBearerResponse / TokenBearer)

Optional nit: it would be good to move the directionary types to `credentials-storage.ts` because they're used only there, they're not global.

Optional nit: Rename:

CredentialsResponse => UserCredentials
UserCredentials => UserCredentialsDictionary
TokenBearerResponse => TokenBearer
TokenBearer => TokenBearerDictionary

(with the Dictionary types in credentials-storage.ts as suggested above, and the other types kept here)

julienw (approved):
Let's land this and move forward!

src/types/types.ts
esanuandra: Thank you! I was confused about how to keep them.

PR #649, opened 2024-05-13
base title fix with new tests

Carla-Moz:
My last patch had a bug for the base component title. I also added two tests to make sure the titles are always correct.


julienw (approved):
Thanks, looks much better :-)

I should have seen that in your previous patch 🤦 The existing snapshots were showing the change properly actually, so we might not need the extra test. Up to you if you want to keep it though :)

PR #653, opened 2024-05-17
Donot render non expanded rows

Carla-Moz:
In this PR, the non expanded rows are not rendered unless `expanded` is true. Ran the profiler and it reduces the jank significantly especially in my compare-over-time patch. Closes https://mozilla-hub.atlassian.net/browse/PCF-439

### DIFF
--- a/src/components/CompareResults/RevisionRow.tsx
+++ b/src/components/CompareResults/RevisionRow.tsx
@@ -262,15 +262,14 @@ function RevisionRow(props: RevisionRowProps) {
           </div>
         </div>
       </div>
-
-      <div
-        className={`content-row content-row--${
-          expanded ? 'expanded' : 'default'
-        } ${stylesCard.container} `}
-        data-testid='expanded-row-content'
-      >
-        {expanded && <RevisionRowExpandable result={result} />}
-      </div>
+      {expanded && (
+        <div
+          className={`content-row ${stylesCard.container}`}
+          data-testid='expanded-row-content'
+        >
+          <RevisionRowExpandable result={result} />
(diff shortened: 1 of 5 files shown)


src/styles/ExpandableRow.ts
```diff
@@ -6,14 +6,12 @@ export const ExpandableRowStyles = () => {
...
       flexDirection: 'row',
       flexWrap: 'wrap',
       cursor: 'pointer',
       transition: 'border-radius 0.4s ease-in-out',
       $nest: {
         '&.content-row': {
-          display: 'none',
           cursor: 'default',
           $nest: {
             '&.content-row--expanded': {
```
julienw, line 17: now you can merge the values for `.content-row--expanded` with `.content-row` above, and remove the rule `.content-row--expanded`. Indeed when displayed it will always be in the expanded state now.

Carla-Moz: Yes, I intended to do that. Thanks for catching

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -269,7 +269,7 @@ function RevisionRow(props: RevisionRowProps) {
         } ${stylesCard.container} `}
         data-testid='expanded-row-content'
       >
-        <RevisionRowExpandable result={result} />
+        {expanded && <RevisionRowExpandable result={result} />}
```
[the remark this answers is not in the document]
Carla-Moz: Yup, makes sense!

beatrice-acasandrei (approved):

PR #654, opened 2024-05-23
added new tests

Carla-Moz:
-`OverTimeResultsView` test
-handled no revisions selected error case for compare over time `Compare` button 
-handled redirect to compare-over-time-results page

### DIFF
--- a/src/__tests__/CompareResults/OverTimeResultsView.test.tsx
+++ b/src/__tests__/CompareResults/OverTimeResultsView.test.tsx
@@ -33,20 +33,12 @@ jest.mock('../../utils/location');
 // const mockedGetLocationOrigin = getLocationOrigin as jest.Mock;
 
 describe('Results View', () => {
-  it('renders correctly', async () => {
-    renderWithRoute(
-      <OverTimeResultsView title={Strings.metaData.pageTitle.results} />,
-    );
-
-    expect(document.body).toMatchSnapshot();
-  });
-
   it('The table should match snapshot and other elements should be present in the page', async () => {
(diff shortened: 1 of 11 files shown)


src/__tests__/CompareResults/OverTimeResultsView.test.tsx
```diff
@@ -0,0 +1,53 @@
...
+  });
+
+  it('The table should match snapshot and other elements should be present in the page', async () => {
+    renderWithRoute(
+      <OverTimeResultsView title={Strings.metaData.pageTitle.results} />,
+    );
+    expect(await screen.findByRole('table')).toMatchSnapshot();
+
+    const link = screen.getByRole('link', { name: /link to home/i });
+    expect(link).toBeInTheDocument();
```
julienw, line 51: (optional) I would combine this with the previous testcase

also I'm not sure the 2 snapshots are useful, maybe one is enough?
```ts
// make sure everything is rendered (esp when we'll use deferred later)
await screen.findByRole('table')
// check the full page
```

The goal is to keep the test run fast enough; currently they're a bit slow, I think because we render a lot and we do a lot of snapshots -- but we should profile the tests to be sure :-) (later)

julienw (approved):
looks good to me, thanks for the additional tests, esp for the compare over time page!

src/__tests__/CompareResults/OverTimeResultsView.test.tsx
Carla-Moz: Yeah that makes sense - Thanks!

PR #655, opened 2024-05-24
React router loader for subtests view

beatrice-acasandrei:
Testing links:
[Subtests from comparison with base rev](https://deploy-preview-655--mozilla-perfcompare.netlify.app/subtestsCompareWithBase?baseRev=42f8b13731a592d1a25e662692cbd82cee5012e0&baseRepo=mozilla-central&newRev=80ae2c7908e4cc558336a6f6a78097d9c2c9317d&newRepo=mozilla-central&framework=1&parentSignature=4769486)
[Subtests from comparison over time ](https://deploy-preview-655--mozilla-perfcompare.netlify.app/subtestsCompareOverTime?baseRepo=mozilla-central&newRev=d5738a6a3f878a3a8eb8c4c5d7e544d11b6b344a&newRepo=mozilla-central&framework=1&parentSignature=4769486&interval=172800)
(Julien's note: I replaced localhost with the deploy preview domain in your links)

### DIFF
--- a/src/__tests__/utils/fixtures.ts
+++ b/src/__tests__/utils/fixtures.ts
@@ -142,6 +142,7 @@ const getTestData = () => {
       is_improvement: true,
       is_regression: false,
       is_meaningful: true,
+      parent_signature: '4769486',
     },
     {
       base_rev: 'coconut',
@@ -187,6 +188,7 @@ const getTestData = () => {
       is_improvement: false,
       is_regression: true,
       is_meaningful: true,
+      parent_signature: '4769486',
     },
     {
       base_rev: 'coconut',
@@ -232,6 +234,7 @@ const getTestData = () => {
       is_improvement: false,
       is_regression: false,
       is_meaningful: true,
+      parent_signature: '4769486',
     },
     {
       base_rev: 'coconut',
@@ -277,6 +280,7 @@ const getTestData = () => {
       is_improvement: false,
       is_regression: false,
       is_meaningful: true,
+      parent_signature: '4769486',
     },
   ];
 
@@ -326,6 +330,7 @@ const getTestData = () => {
       is_improvement: true,
       is_regression: false,
       is_meaningful: true,
+      parent_signature: '4769486',
     },
     {
       base_rev: 'coconuts',
@@ -371,6 +376,7 @@ const getTestData = () => {
       is_improvement: true,
       is_regression: false,
       is_meaningful: true,
+      parent_signature: '4769486',
     },
     {
       base_rev: 'coconuts',
@@ -416,6 +422,7 @@ const getTestData = () => {
       is_improvement: true,
       is_regression: false,
       is_meaningful: true,
+      parent_signature: '4769486',
     },
     {
       base_rev: 'coconuts',
@@ -461,6 +468,7 @@ const getTestData = () => {
       is_improvement: true,
       is_regression: false,
       is_meaningful: true,
+      parent_signature: '4769486',
     },
     {
       base_rev: 'coconuts',
@@ -506,6 +514,7 @@ const getTestData = () => {
       is_improvement: true,
       is_regression: false,
       is_meaningful: true,
+      parent_signature: '4769486',
     },
     {
       base_rev: 'coconuts',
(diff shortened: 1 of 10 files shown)


src/logic/treeherder.ts
```diff
@@ -92,6 +109,78 @@ export async function fetchCompareOverTimeResults({
...
+    } else {
+      throw new Error(
+        `Error when requesting treeherder: (${response.status}) ${response.statusText}`,
+      );
+    }
+  }
+
+  return response.json() as Promise<CompareResultsItem[]>;
+}
+
```
julienw, line 183: All the functions here are very similar, only the searchParams is different in all these functions.
We could probably extract a (non exported) function taking the `searchParams` as a parameter, doing the fetch, so that all the functions would be smaller.
Optional for this PR, this could be a follow-up.

src/__tests__/utils/fixtures.ts
```diff
@@ -142,6 +142,7 @@ const getTestData = () => {
       is_improvement: true,
       is_regression: false,
       is_meaningful: true,
+      parent_signature: '4769486',
```
julienw, line 145: Tell me if I'm wrong, but my guess is that `parent_signature` is returned only for subtests.
Therefore I'm not sure we should have them by default, rather we should have it only for subtests (so when you'll add tests, you'll add them recursively using `map` when defining the mock).

src/mockData/9d50665254899d8431813bdc04178e6006ce6d59.ts
```diff
@@ -45,6 +45,7 @@ export const comparisonResults: CompareResultsItem[] = [
     is_improvement: false,
     is_regression: false,
     is_meaningful: true,
+    parent_signature: '4769486',
```
julienw, line 48: same remark as for the fixtures

julienw (commented):
Thanks!
I have a few comments, tell me what you think

src/__tests__/utils/fixtures.ts
julienw: see the discussion with the type

src/types/state.ts
```diff
@@ -91,6 +91,7 @@ export type CompareResultsItem = {
   is_regression: boolean;
   is_meaningful: boolean;
   more_runs_are_needed: boolean;
+  parent_signature: string;
```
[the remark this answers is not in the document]
julienw: Then you can remove the property from the fixtures and mocks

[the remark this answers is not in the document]
beatrice-acasandrei: @julienw we need the `parent_signature` to retrieve the subtests results and we send it as an url param. That's why we need to always have access to it when getting the regular comparison results.

src/__tests__/utils/fixtures.ts
beatrice-acasandrei: We need access to the parent_signature value so we can make a request to get the subtests results, otherwise we can't know which subtests compare results belong to a certain regular compare result (if that makes sense).

src/types/state.ts
```diff
@@ -91,7 +91,7 @@ export type CompareResultsItem = {
   is_regression: boolean;
   is_meaningful: boolean;
   more_runs_are_needed: boolean;
-  parent_signature: string;
+  parent_signature: number | null;
```
beatrice-acasandrei, line 94: I corrected a mistake I made regarding the response of the endpoint.

src/__tests__/utils/fixtures.ts
```diff
@@ -142,7 +142,7 @@ const getTestData = () => {
       is_improvement: true,
       is_regression: false,
       is_meaningful: true,
-      parent_signature: '4769486',
+      parent_signature: null,
```
beatrice-acasandrei, line 145: I think it makes more sense to leave them as `null` for now - when adding tests we can adjust.

src/types/state.ts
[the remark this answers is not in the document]
julienw: it's just not clear to me when it's null and when it's present, if you can try to make this clearer in a comment that would be lovely.

src/logic/treeherder.ts
beatrice-acasandrei: I made a separate [PR](https://github.com/mozilla/perfcompare/pull/663) for this.

julienw (approved):
Let's move forward with that, we can always adjust later a bit if needed!

PR #657, opened 2024-05-28
Rework styles for buttons using the palette from Material UI

julienw:
This PR started as a simple nit PR but ended up doing some more. This changes the button using the "primary" and "secondary" colors, I didn't touch to the dropdown styles though because I didn't need that... maybe later ;-)

There are 4 commits then:
* Commit 1: Small nit regarding the compare button
* Commit 2: Use MUI's palette for primary buttons
* Commit 3: Use MUI's palette for secondary buttons
* Commit 4: Update tests

More comments in the commits themselves.

There should be no visible change with this PR besides the Download Button. That one now works with dark mode, and also gains a box shadow that should be removed with #665.

### DIFF
--- a/src/components/Search/CompareButton.tsx
+++ b/src/components/Search/CompareButton.tsx
@@ -1,27 +1,18 @@
 import Button from '@mui/material/Button';
-import { style } from 'typestyle';
-
-import { useAppSelector } from '../../hooks/app';
-import { ButtonStyles } from '../../styles';
 
 interface CompareButtonProps {
   label: string;
 }
 
 export default function CompareButton({ label }: CompareButtonProps) {
-  const mode = useAppSelector((state) => state.theme.mode);
-  const btnStyles = ButtonStyles(mode);
-
-  const styles = {
-    button: style({ ...btnStyles.Primary }),
-  };
-
   return (
     <Button
       id='compare-button'
-      className={`compare-button ${styles.button}`}
-      sx={{ textTransform: 'none !important' }}
+      color='primary'
       type='submit'
+      sx={{
+        height: 32,
+      }}
     >
       {label}
     </Button>
--- a/src/styles/Buttons.ts
+++ b/src/styles/Buttons.ts
@@ -14,16 +14,6 @@ const sharedDropDownBtnStyles = {
 
 //BUTTONS LIGHT
 export const ButtonsLightRaw = {
-  Secondary: {
-    color: Colors.PrimaryText,
-    backgroundColor: Colors.SecondaryDefault,
-    '&:hover': {
-      backgroundColor: Colors.SecondaryHover,
-    },
-    '&:active': {
-      backgroundColor: Colors.SecondaryActive,
-    },
-  },
   Dropdown: {
     ...sharedDropDownBtnStyles,
     backgroundColor: Colors.SecondaryDefault,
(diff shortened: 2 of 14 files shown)


src/components/Search/CompareButton.tsx
```diff
@@ -1,28 +1,19 @@
...
-  const styles = {
-    button: style({ ...btnStyles.Primary }),
-  };
-
   return (
     <Button
       id='compare-button'
       variant='contained'
-      className={`compare-button ${styles.button}`}
-      sx={{ textTransform: 'none !important' }}
```
julienw, line 24: Initially I wanted to just remove this textTransform that was overriding some other CSS in theme/components (see below).

Note that the text transform override for buttons is already globally set in https://github.com/mozilla/perfcompare/blob/3b2e359b0561746b0f0f94e17d8d6a408ae9001c/src/theme/typography.ts#L11-L13 (as it should be).

src/components/Search/CompareButton.tsx
```diff
@@ -1,28 +1,19 @@
...
+      color='primary'
       type='submit'
+      sx={{
+        height: 32,
+      }}
```
julienw, line 16: This was in the shared button styles that I removed from the button as well.
I think the sizing is usually a "local" css, in that it can be different in all different cases. But if you disagree please tell me.
For example maybe the protocol theme says that it should always be 32 px high. If necessary I think we could implement that globally using global classes on MuiButton too.

src/theme/components.js
```diff
@@ -46,12 +46,6 @@ const components = {
             },
           },
         },
-        '&.compare-button': {
-          lineHeight: '1.4375em',
-          padding: '16.5px 14px',
-          textTransform: 'uppercase',
-          marginBottom: '30px',
-        },
```
julienw, line 54: These styles weren't useful, indeed when I removed them nothing changed.

src/styles/Buttons.ts
```diff
@@ -12,27 +12,8 @@ const sharedDropDownBtnStyles = {
...
 
-const sharedButtonStyles = {
-  padding: `${Spacing.xSmall}px ${Spacing.Medium}px !important`,
-  height: `${Spacing.xLarge}px`,
-  margin: '0 !important',
-};
-
 //BUTTONS LIGHT
 export const ButtonsLightRaw = {
-  Primary: {
```
julienw, line 23: This wasn't used anywhere anymore. It's easier to use `color="primary"` which will use the palette Carla defined in `protocolTheme`. From my testing the colors were the same, even hover and active, including in dark mode.

I tried to do the same for Secondary as well, but there the colors were different especially for hover and active, so I didn't want to include the change in this PR.

src/components/Search/CompareButton.tsx
Carla-Moz: Yeah, these global buttons styles like height should be in protocol theme so that we maintain the` buttons `standard set by Dasha in the[ figma](https://www.figma.com/design/LRV76asSVOULlsWbRNZYp0/PerfCompare?node-id=2835-113706&m=dev) file.

Carla-Moz (approved):
LGTM, you can add the btn height to the protocol theme if you wish or later. I'm good with either.

julienw:
> LGTM, you can add the btn height to the protocol theme if you wish or later. I'm good with either.

OK thanks for the feedback, I'll do it in a separate PR because there might be changes needed for other buttons too.

PR #658, opened 2024-05-28
Filter results by title, platform, revision and other things

julienw:
[Deploy preview](https://deploy-preview-658--mozilla-perfcompare.netlify.app/compare-results?fakedata)

You can find 4 commits that can be looked independently (again :D)

Commit 1: very simple refactoring where I move the very simple ResultsHeader into ResultsMain.
Commit 2: very simple change adding an aria-label to search inputs to make them easier to target in tests. It's useful for this feature adding another search box, so that the new tests will be able to target the new search box easily. This is also good for accessibility.
Commit 3: this implements the search functionality.
Commit 4: this adds new tests.
Commit 5: remove the apply button.

Tell me what you think!

image

~~As I mentioned before, I'm not sure we'll want to keep the Apply button but I still kept it for now.~~
Update: after all I decided to remove it (last commit).
I also tweaked the placeholder so that we see more of it (updated the 3rd commit). Esp I removed the "Search by" from the placeholder. Here is the new screenshot:
image


src/components/CompareResults/ResultsMain.tsx
```diff
@@ -33,17 +36,35 @@ function ResultsMain(props: { results: CompareResultsItem[][] }) {
     }),
   };
 
+  const filteredResults = useMemo(
```
julienw, line 39: Honestly I wasn't sure whether to put that here or in ResultsTable. I decided to put it here so that I wouldn't have to move the searchTerm down the component chain, but this is debatable. We could filter _and_ process in `processResult` so that we'd loop the results once less.

Also `searchTerm` could be in redux in the future, which for example would make it easier to highlight the search term in the results.

But I decided to keep it simple for now.

julienw:
Rebased!

src/components/CompareResults/SearchInput.tsx
```diff
@@ -1,53 +1,86 @@
...
-      <Button disabled>Apply</Button>
-    </div>
+    <Box
+      component='form'
+      aria-label={inputStrings.label}
+      onSubmit={onSubmit}
+      sx={{ display: 'flex', marginRight: 'auto' }}
+    >
+      <TextField
+        placeholder={inputStrings.placeholder}
```
esanuandra, line 49: What do you say about using a title or something similar so that the user can see the options he can search by?
The placeholder is not fully visible. I could only see `title, platform and revision` when testing. I wasn't sure if I can search by `options`.

julienw: Yeah I was wondering about adding a title too... It might be good to ask the a11y team about that, as a follow-up. Currently it's in an aria-label but it could easily be a title instead...

I think we should make everything the user might want to search for work so that they don't have to wonder about that. I'll replace the placeholder with "Search results".

Carla-Moz (approved):

src/resources/Strings.tsx
```diff
@@ -128,6 +128,10 @@ export const Strings = {
         text: 'All revisions',
       },
     },
+    searchResultsInput: {
+      placeholder: 'title, platform, revision or options',
```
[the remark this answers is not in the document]
julienw: I just did that (increased the width), and replaced by "Search results" like Dasha suggested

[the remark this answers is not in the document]
julienw: BTW I've put the width to 50% instead of a precise px value.
Ideally we should use MUI grids for this (https://mui.com/material-ui/react-grid2/) but I didn't want to change this now... but this could make it easier for us to support smaller screens, where we're not great at at the moment.

julienw:
Thanks for the review!

PR #659, opened 2024-05-29
added coverage for overTimeLoader

Carla-Moz:
Also fixed a bug re getting interval info from url while writing tests for overTimeLoader.

### DIFF
--- a/src/components/CompareResults/overTimeLoader.ts
+++ b/src/components/CompareResults/overTimeLoader.ts
@@ -42,12 +42,12 @@ function checkValues({
     );
   }
 
-  if (framework === null) {
-    framework = 1; // default to talos so that manually typing the URL is easier
+  if (interval === null) {
+    throw new Error('The parameter interval is missing.');
   }
 
-  if (interval === null) {
-    interval = 86400; // default to 1 day
+  if (framework === null) {
+    framework = 1; // default to talos so that manually typing the URL is easier
   }
 
   const frameworkId = +framework as Framework['id'];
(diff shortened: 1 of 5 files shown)


src/components/CompareResults/overTimeLoader.ts
```diff
@@ -26,10 +26,30 @@ function checkValues({
...
+    );
+  }
+
   if (framework === null) {
     framework = 1; // default to talos so that manually typing the URL is easier
   }
 
+  if (interval === null) {
+    interval = 86400; // default to 1 day
+  }
```
julienw, line 51: for compare over time, shouldn't this be an error actually?

src/components/CompareResults/overTimeLoader.ts
```diff
@@ -46,9 +66,6 @@ function checkValues({
     );
   }
 
-  if (interval === null) {
-    interval = 86400; // default to 1 day
-  }
   const intervalValue = +interval as TimeRange['value'];
```
julienw, line 69: I just noticed that the return values below do not return these values, but instead they return hardcoded values :) This should be fixed (in this PR or in another PR as you wish)

julienw (approved):
thanks for the additional coverage

src/components/CompareResults/overTimeLoader.ts
Carla-Moz: That's silly. I don't know why I did that. I must have been testing something and forgot to revert it. I'll fix it in this PR. Thanks for catching that!

PR #661, opened 2024-05-30
change the prop name to make it less confusing

Carla-Moz:
The prop `hasNonEditableState` is confusing because of the double negative when using the false boolean. Let's make it positive, `hasEditableState`, so there's less cognitive load.

### DIFF
--- a/src/components/CompareResults/ResultsView.tsx
+++ b/src/components/CompareResults/ResultsView.tsx
@@ -44,7 +44,7 @@ function ResultsView(props: ResultsViewProps) {
         <LinkToHome />
 
         <CompareWithBase
-          hasNonEditableState={true}
+          hasEditableState={true}
           baseRev={baseRevInfo ?? null}
           newRevs={newRevsInfo ?? []}
           frameworkIdVal={frameworkId}
--- a/src/components/CompareResults/ResultsView.tsx
+++ b/src/components/CompareResults/ResultsView.tsx
@@ -44,7 +44,7 @@ function ResultsView(props: ResultsViewProps) {
         <LinkToHome />
 
         <CompareWithBase
-          hasEditableState={true}
+          hasEditButton={true}
           baseRev={baseRevInfo ?? null}
           newRevs={newRevsInfo ?? []}
           frameworkIdVal={frameworkId}
--- a/src/components/Search/SearchComponent.tsx
+++ b/src/components/Search/SearchComponent.tsx
@@ -23,7 +23,7 @@ import SearchInputAndResults from './SearchInputAndResults';
 import SelectedRevisions from './SelectedRevisions';
 
 interface SearchProps {
-  hasNonEditableState: boolean;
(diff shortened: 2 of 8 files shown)


src/components/CompareResults/ResultsView.tsx
```diff
@@ -43,7 +43,7 @@ function ResultsView(props: ResultsViewProps) {
         <LinkToHome />
 
         <CompareWithBase
-          hasNonEditableState={true}
+          hasEditableState={true}
```
julienw, line 46: shouldn't the boolean pass to false here?

src/components/Search/SearchComponent.tsx
```diff
@@ -83,7 +83,7 @@ function SearchComponent({
     },
   });
 
-  const [formIsDisplayed, setFormIsDisplayed] = useState(!hasNonEditableState);
+  const [formIsDisplayed, setFormIsDisplayed] = useState(!hasEditableState);
```
julienw, line 86: if you invert the meaning, shouldn't you also change how you use it? In this case removing the `!`?

src/components/Search/SearchComponent.tsx
```diff
@@ -101,7 +101,7 @@ function SearchComponent({
           </Tooltip>
         </InputLabel>
         {/**** Edit Button ****/}
-        {hasNonEditableState && !formIsDisplayed && (
+        {hasEditableState && !formIsDisplayed && (
```
julienw, line 104: similar here and below, probably need to add `!`

julienw (commented):
I'm all in favor of finding a new name if the current name is confusing. I'm not myself a big fan of the current one either.

But the problem is that even when "false" it has an editable state... indeed the user can edit it!
When the prop is true, it adds this "in between" state where the user has to press "edit" to move to edit mode. 
But when the prop is false, there's no "in between" state and it's always editable.
Hence the name... this adds this "non editable" state.

Therefore it's not really a "double negative" like you mention, it is actually a positive attribute.
When set to "true", this _adds_ another state. When set to "false" this state doesn't exist.


Again I'm in favor of finding a different name because I understand how it can be confusing. But I don't think `hasEditableState` is less confusing.

What do you think of something more concrete such as `hasEditButton`?

src/components/CompareResults/ResultsView.tsx
julienw: oops disregard this comment and others, I misunderstood the patch initially.

Carla-Moz: Actually the definition of edit is amending a draft, so in the `Search` page the user is _creating_ the committed state and in the `Results` page they are _editing_ the committed state when they press the edit button. However, you're right; it's not a double negative. I was mistaken. Semantics and grammar aside, I agree to name it `hasEditButton`, lol.

julienw (approved):
Thanks :-)

PR #662, opened 2024-05-30
pass down repo info and adjusted tests

Carla-Moz:
In the same way that we pass down the rev information for the selected revs in the results list, it makes sense to pass down the repo info so the repo drop-down is already set to the committed repo in the Results view edit section

### DIFF
--- a/src/components/Search/SearchComponent.tsx
+++ b/src/components/Search/SearchComponent.tsx
@@ -173,7 +173,7 @@ function SearchComponent({
       {displayedRevisions && (
         <Grid
           className='d-flex'
-          data-testid={`${isBaseComp ? 'base' : 'new'}-selected-revision`}
+          data-testid={`${searchType}-selected-revision`}
         >
           <SelectedRevisions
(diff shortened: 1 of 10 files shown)


src/components/Search/SearchComponent.tsx
```diff
@@ -171,7 +171,10 @@ function SearchComponent({
       </Grid>
       {/***** Selected Revisions Section *****/}
       {displayedRevisions && (
-        <Grid className='d-flex'>
+        <Grid
+          className='d-flex'
+          data-testid={`${isBaseComp ? 'base' : 'new'}-selected-revision`}
```
julienw, line 176: nit: reuse `searchType` directly instead of using the ternary

julienw (approved):
Makes sense, thanks for the change!

src/components/Search/SearchComponent.tsx
Carla-Moz: True, forgot about that

PR #663, opened 2024-05-30
Small refactor on treeherder.ts

beatrice-acasandrei:


julienw (changes requested):
As discussed on Matrix, there's value in keeping the high level functions for their strongly typed parameters, but making them call the underlying function with the right parameters.
Please request review again once this is ready :)

Carla-Moz (approved):
This looks a lot cleaner and makes sense to me. Thanks!

julienw (approved):
Thanks for the cleanup!

PR #664, opened 2024-05-30
Fix the bad layout happening sometimes in the compare results

julienw:
This removes some unused style as well as some clashing styles.

### DIFF
--- a/src/components/CompareResults/ResultsMain.tsx
+++ b/src/components/CompareResults/ResultsMain.tsx
@@ -17,7 +17,6 @@ function ResultsMain(props: { results: CompareResultsItem[][] }) {
   const styles = {
     container: style({
       backgroundColor: themeColor100,
-      maxWidth: '950px',
       margin: '0 auto',
       marginBottom: '80px',
     }),
--- a/src/components/CompareResults/ResultsTable.tsx
+++ b/src/components/CompareResults/ResultsTable.tsx
@@ -2,12 +2,11 @@ import { useMemo } from 'react';
 
 import Box from '@mui/material/Box';
 import CircularProgress from '@mui/material/CircularProgress';
-import Paper from '@mui/material/Paper';
 import { style } from 'typestyle';
 
 import { useAppSelector } from '../../hooks/app';
 import { Strings } from '../../resources/Strings';
-import { Colors, Spacing } from '../../styles';
+import { Spacing } from '../../styles';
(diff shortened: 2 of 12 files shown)


src/theme/components.js
julienw, line 1: These styles didn't seem to be used.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -60,13 +59,7 @@ function processResults(results: CompareResultsItem[]) {
 const allRevisionsOption =
   Strings.components.comparisonRevisionDropdown.allRevisions.key;
 
-const customStyles = {
-  boxShadow: 'none',
```
julienw, line 64: instead of overrinding boxShadow I simply used a Box (which is really like a div expect it's more integrated in mui... we could probably use a div here actually, not sure).

I tried `<Paper elevation={0}` but this also changed the background in dark mode, so this wasn't working as expected.

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -17,7 +17,6 @@ function ResultsMain(props: { results: CompareResultsItem[][] }) {
   const styles = {
     container: style({
       backgroundColor: themeColor100,
-      maxWidth: '950px',
```
julienw, line 20: This was clashing with mui's default behavior on `<Container>`; they use `maxWidth: 1200px` for screens bigger than that, and otherwise use a fluid layout.
Because this is the behavior we have generally I decided to keep it, and removed this fixed `maxWidth`.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -85,22 +78,17 @@ function ResultsTable(props: { results: CompareResultsItem[][] }) {
   // TODO Implement a loading UI through the react-router defer mechanism
   const loading = false;
 
-  const themeColor100 =
-    themeMode === 'light' ? Colors.Background100 : Colors.Background100Dark;
-
   const styles = {
     tableContainer: style({
-      backgroundColor: themeColor100,
```
julienw, line 93: this backgroundColor was clashing with the `background: none` above. Instead I removed both of them to keep it transparnet, so that we see the one from ResultsMain instead.

Carla-Moz (approved):
Thanks for the styles cleanup and fixing the janky table width.

PR #665, opened 2024-05-30
Use IconButton for nicer "edit" and "remove revision" icons, and set default appearance for all buttons

julienw:
It's probably easier to look at commits separately.

* Commit 1: uses IconButton instead of Button for some of our icons -- namely the "edit" button and the "remove revision" button. This makes these buttons much nicer in the process, especially on hover. Note that this was needed for setting the variant by default in commit 2, that's why I included it in this PR too.

* Commit 2: set some defaults for all buttons: variant = contained so that they look like button, and disableElevation to remove box shadows. I think that the shared styles do some of "variant = contained" too and we might want to simplify them later, but I didn't want to do that now. (see https://mui.com/material-ui/customization/theme-components/#theme-default-props about this -- we could also add `disableRipple` if we want to)

* Commit 3: remove the "variant" property where appropriate now that it's set globally.

* Commit 4: update test snapshots

To summarize visible changes:
* removes the box shadows on buttons like the design says
* makes the "remove revision" icon a bit bigger
* makes the hover style for both the "edit" icon and "remove revision" icon nicer, and their hit size bigger
*  Some spacings are slightly different too but I think I managed to keep these changes minimal.

I think that the only debatable change is making the variant contained by default. It can still be overriden on a case by case basis by using `variant="text" or "outlined"`.

Please tell me what you think!

[Deploy preview](https://deploy-preview-665--mozilla-perfcompare.netlify.app/compare-results?baseRev=b05a24f158503b7ea175520ef383fa005a4c41b7&baseRepo=mozilla-central&newRev=fd0f25542804074c0a38ba6c7f273a36bf658127&newRepo=mozilla-central&framework=1)
[Production version](https://beta--mozilla-perfcompare.netlify.app/compare-results?baseRev=b05a24f158503b7ea175520ef383fa005a4c41b7&baseRepo=mozilla-central&newRev=fd0f25542804074c0a38ba6c7f273a36bf658127&newRepo=mozilla-central&framework=1)

Before:
image
After:
image

### DIFF
--- a/src/__tests__/Search/__snapshots__/CompareOverTime.test.tsx.snap
+++ b/src/__tests__/Search/__snapshots__/CompareOverTime.test.tsx.snap
@@ -11,7 +11,7 @@ exports[`Compare Over Time renders correctly in Search View: Initial state for t
     class="MuiGrid-root component_f14z0jsa css-plh9qo-MuiGrid-root"
   >
     <div
-      class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-2 new-search-dropdown dropDown_fl6uoag label-edit-wrapper css-13eijkh-MuiGrid-root"
+      class="MuiGrid-root MuiGrid-item MuiGrid-grid-xs-2 new-search-dropdown dropDown_f1qectas label-edit-wrapper css-13eijkh-MuiGrid-root"
     >
       <label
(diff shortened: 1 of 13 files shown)


Carla-Moz (approved):
Thanks the buttons styles cleanup! Along with removing the box shadows.

PR #666, opened 2024-05-31
Use the global overrides to define the default height of buttons

julienw:
As discussed in https://github.com/mozilla/perfcompare/pull/657#pullrequestreview-2089537267, I added the height CSS property with value 32px for all contained buttons to the global theme configuration.


src/theme/components.js
```diff
@@ -17,6 +17,9 @@ const components = {
       variant: 'contained',
     },
     styleOverrides: {
+      contained: {
```
julienw, line 20: I decided to apply it only for "contained" buttons. I thought that other types of buttons, if we use them, would possibly be different. Happy to move it to "root" if you prefer though.

julienw: Also I wonder if maxHeight is better, what do you think?

Carla-Moz: yes, maxHeight is better.

Carla-Moz (approved):

src/theme/components.js
julienw: Thanks for the feedback!
So....
I tried maxHeight, but this has some issues.

See the download button for example:
image
We definitely want it to take the full height. I tried to play with flex to make it take the full height even with this maxHeight, but this wouldn't work. I'd have to override it.

If I use minHeight instead, then I'll have similar problems with the filtering options (because they'll be buttons):
image

Now I believe `height` is easier because it's more easily overridable. It's even overriden "by default" when using Flex, because it's used as the flex-basis and then the flex algorithm makes it grow/shrink, while max/minHeight are very strict.

So in the end I kept it with "height"!

PR #667, opened 2024-06-01
remove save btn and move cancel

Carla-Moz:
Also removed styles related to save cancel pair; updated search component and compare button files; removed staging state since save is no longer needed

**_Resolved comments guide:_**
The revised PR does the following:

Please see[ resolved comments](https://github.com/mozilla/perfcompare/pull/667/commits/3c4eeba58886d10f789d3e15df0c4324f3ccee1d) commit.

- [x] Combined `Cancel` and `Compare` into one component `CancelAndCompareButtons`
- [x] Removed unnecessary props from `CancelAndCompareButton` and added `HasCancelButton` to both `CompareWithBase` and `CompareOverTime`
- [x] moved the cancel and compare styles from global styles folder to the component instead. 
- [x] moved button padding styles to contained override styles for the buttons

Update:
I don't think I made it clear that I would handle the global edit button in a separate PR but I've gone ahead and handle it in this PR based on the comments. I apologize for the extra review. 

This PR now closes:

- [x] [Remove the save button in the edit mode]( https://mozilla-hub.atlassian.net/browse/PCF-415)
- [x] [Results: Replace edit buttons with one global edit button](https://mozilla-hub.atlassian.net/browse/PCF-409)


src/theme/components.js
```diff
@@ -18,6 +18,9 @@ const components = {
     },
     styleOverrides: {
       root: {
+        '&.btn-all': {
+          padding: '4px 16px',
+        },
```
julienw, line 23: If this should be added to all buttons, better add it directly to `root` without this extra class.
See how I did it in https://github.com/mozilla/perfcompare/pull/666

src/styles/CompareCards.ts
```diff
@@ -125,31 +127,29 @@ export const SearchStyles = (mode: string) => {
...
     },
+    cancelCompareBtn: {
+      display: 'flex',
+      alignItems: 'center',
+      $nest: {
+        '.cancel-button': {
+          marginRight: `${Spacing.Small}px`,
+        },
+      },
+    },
```
julienw, line 138: can we use this opportunity to style the cancel button directly on the componen rather than with these classes?

src/components/Search/CompareButton.tsx
```diff
@@ -1,20 +1,48 @@
...
 interface CompareButtonProps {
   label: string;
+  hasNonEditableState: boolean;
+  onCancel: () => void;
+  onSetDisplayForm: (display: boolean) => void;
+  formIsDisplayed: boolean;
 }
 
-export default function CompareButton({ label }: CompareButtonProps) {
+export default function CompareButton({
```
julienw, line 15: What do you think of renaming to "CancelAndCompareButton"?

src/components/Search/CompareButton.tsx
```diff
@@ -1,20 +1,48 @@
...
+import { useAppSelector } from '../../hooks/app';
+import { SearchStyles } from '../../styles';
+import CancelButton from './CancelButton';
+
 interface CompareButtonProps {
   label: string;
+  hasNonEditableState: boolean;
+  onCancel: () => void;
+  onSetDisplayForm: (display: boolean) => void;
+  formIsDisplayed: boolean;
```
julienw, line 12: Try to think about the "point of view" of this component.

All it needs to do its work is:
* should the cancel button be displayed => prop `hasCancelButton: boolean`
* was the cancel button clicked => onCancel

Therefore you can remove `hasNonEditableState` and `formIsDisplayed` (its caller can use them to pass `hasCancelButton` of course), and also `onSetDisplayForm` because the caller will handle this in its `onCancel` callback.

src/components/Search/CancelButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+export default function CancelButtons({ onCancel }: CancelButtonsProps) {
+  return (
+    <Button
+      className={`cancel-button btn-all`}
+      name='cancel-button'
+      color='secondary'
+      onClick={onCancel}
+    >
+      {cancel}
+    </Button>
```
julienw, line 21: it probably can be moved directly into CancelAndCompareButton, I'm not sure having it in a separate component is worth it.

src/styles/CompareCards.ts
```diff
@@ -187,8 +187,8 @@ export const SearchStyles = (mode: string) => {
       left: '220px',
       $nest: {
         '&.big': {
-          left: '190px',
-          minWidth: '510px',
+          left: '212px',
+          minWidth: `${850 - repoDropdownWidth - Spacing.Default}px`,
```
julienw, line 191: note: I looked at the style of this recently, I was surprised how the style worked (with some absolute positioning etc instead of flex). I thought this was because of the dropdown. But let's not change that now anyway.

src/components/Search/CompareWithBase.tsx
```diff
@@ -80,14 +80,7 @@ function CompareWithBase({
...
   const [frameWorkId, setframeWorkValue] = useState(frameworkIdVal);
 
-  // The "committed" base and new revisions initialize the staging state.
-  // These states are snapshots of selections, that the user can either save (by
-  // pressing the "Save" button) or revert to (by pressing the Cancel button).
-  const [baseStagingRev, setStagingBaseRev] = useState<Changeset | null>(
-    baseRev,
-  );
-  const [newStagingRevs, setStagingNewRevs] = useState<Changeset[]>(newRevs);
-
```
julienw, line 90: 🥳

src/components/Search/SearchComponent.tsx
```diff
@@ -101,12 +96,13 @@ function SearchComponent({
           </Tooltip>
         </InputLabel>
         {/**** Edit Button ****/}
-        {hasNonEditableState && !formIsDisplayed && (
+        {/* temporary until global edit button PR */}
+        {hasNonEditableState && !formIsDisplayed && isBaseComp && (
```
julienw, line 100: nit: I'd rather have a boolean prop `showEditButton`

src/components/Search/SearchComponent.tsx
```diff
@@ -101,12 +96,13 @@ function SearchComponent({
...
           <EditButton
             isBase={isBaseComp}
             onEditAction={() => {
               onEdit();
-              setFormIsDisplayed(true);
+              onSetDisplayForm(true);
```
julienw, line 105: If the state goes up, I think this could go in the onEdit callback (in the caller) directly, and we don't need `onSetDisplayForm` at all.

julienw:
I wasn't sure how ready the PR was but still went through it and gave a few comments.

I thought we agreed that we didn't want to change edit/cancel that much for the MVP (just removing the Save button), but well I'm not opposed.

src/styles/CompareCards.ts
Carla-Moz: Yes, that makes sense. Keep forgetting to do that. Will remember for next time. Thanks.

src/components/Search/CompareButton.tsx
Carla-Moz: Yes, I agree.

src/components/Search/CompareButton.tsx
Carla-Moz: Right, right. Rushed through this too quickly Thanks!

src/components/Search/SearchComponent.tsx
Carla-Moz: Yes, the name has been changed in my last PR. I'm going to handle the edit button in a new PR since this one is already crowded.

Carla-Moz:
> I wasn't sure how ready the PR was but still went through it and gave a few comments.
> 
> I thought we agreed that we didn't want to change edit/cancel that much for the MVP (just removing the Save button), but well I'm not opposed.

I decided to go ahead and do this because while I was working on the edit feature for CompareOverTime, it didn't make sense to replicate an edit feature that we would revamp so I wanted it out of the way. Both components share many of the same edit comps, config, etc.

src/styles/CompareCards.ts
Carla-Moz: Yes, we can deal with this in a separate cleanup PR.

src/components/Search/SearchComponent.tsx
```diff
@@ -101,12 +94,13 @@ function SearchComponent({
           </Tooltip>
         </InputLabel>
         {/**** Edit Button ****/}
-        {hasEditButton && !formIsDisplayed && (
+
+        {/* temporary until global edit button PR */}
+        {hasEditButton && !formIsDisplayed && isBaseComp && (
```
julienw, line 99: It would be better to have another boolean controlling this.
But the best name for this boolean would be `hasEditButton`, right? 😅 and it clashes with the other one, and frankly I don't want to nitpick about that anymore. So let's keep it this way now.

src/components/Search/SearchComponent.tsx
```diff
@@ -101,12 +94,13 @@ function SearchComponent({
...
           <EditButton
             isBase={isBaseComp}
             onEditAction={() => {
               onEdit();
-              setFormIsDisplayed(true);
             }}
```
julienw, line 104: optional:
```
onEditAction={onEdit}
```

src/components/Search/CompareWithBase.tsx
```diff
@@ -290,15 +280,14 @@ function CompareWithBase({
...
           />
           <SearchComponent
             {...stringsNew}
             isBaseComp={false}
             hasEditButton={hasEditButton}
             isWarning={isWarning}
             displayedRevisions={newInProgressRevs}
-            onSave={handleSaveNew}
-            onCancel={handleCancelNew}
             onEdit={handleEditNew}
```
julienw, line 291: there's no need for onEdit on this SearchComponent now, you can pass an empty function `()=>{}` and remove `handleEditNew`.

src/components/Search/CompareWithBase.tsx
```diff
@@ -99,6 +92,8 @@ function CompareWithBase({
     useState<Changeset[]>(newRevs);
   const [baseRepository, setBaseRepository] = useState(baseRepo);
   const [newRepository, setNewRepository] = useState(newRepo);
+  const [formIsDisplayed, setFormIsDisplayed] = useState(!hasEditButton);
+  const [hasCancelButton, setHasCancelButton] = useState(false);
```
julienw, line 96: optional:
I think this state about cancelButton isn't useful and can be simply computed from these 2 other states:
```js
const hasCancelButton = formIsDisplayed && hasEditButton
```

src/components/Search/CompareWithBase.tsx
```diff
@@ -137,25 +132,22 @@ function CompareWithBase({
...
-  const handleSaveBase = () => {
-    setStagingBaseRev(baseInProgressRev);
-  };
-  const handleSaveNew = () => {
-    setStagingNewRevs(newInProgressRevs);
+  const handleCancel = () => {
+    setInProgressNewRevs(newRevs);
+    setInProgressBaseRev(baseRev);
+    setFormIsDisplayed(false);
+    setHasCancelButton(false);
```
julienw, line 139: I believe we also need to set it to false when the form is submitted. Curiously this didn't look so strange with the previous design, but now I really feel like it's needed.

My suggestion therefore is to:
1. rename `possiblyPreventFormSubmission` to `onFormSubmit`
2. In that function, add the code to hide the form.

Tell me what you think!

julienw (approved):

I only have a few nits, tell me what you think. You can land without an additional review after fixing these!

src/components/Search/CompareWithBase.tsx
Carla-Moz: Yes, I planned to handle this in a new PR.

src/components/Search/SearchComponent.tsx
Carla-Moz: Yes, I left this here, but I'll just go ahead and add the Edit Entry global button in this PR too.

src/components/Search/SearchComponent.tsx
Carla-Moz: I'll simply remove it from the search component. That's why I said it was temporary until the edit button PR.

src/components/Search/CompareWithBase.tsx
Carla-Moz: Yes, this makes sense. I thought of this as well.

src/__tests__/Search/CompareWithBase.test.tsx
```diff
@@ -74,15 +74,18 @@ function getEditButtons() {
...
-    const target = screen.getByText(targetRevision);
+    const target = screen.getAllByText(targetRevision)[index];
     container = within(target.closest('li') ?? document.body);
   }
 
   // eslint-disable-next-line testing-library/prefer-screen-queries
-  return container.getByRole('button', {
+  return container.getAllByRole('button', {
     name: 'remove revision',
   });
```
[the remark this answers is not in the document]
Carla-Moz: Noted!

src/__tests__/Search/CompareWithBase.test.tsx
```diff
@@ -276,140 +276,16 @@ describe('Compare With Base', () => {
...
-    // Click the remove button for the "ridden on a horse" item.
-    await user.click(getRemoveRevisionButton(/ridden on a horse/));
-    expect(
-      screen.queryByText(/What, ridden on a horse?/),
-    ).not.toBeInTheDocument();
-
-    // Click the cancel button
-    await user.click(getCancelButton());
-    // Oh look, it's back!
-    expect(screen.getByText(/What, ridden on a horse?/)).toBeInTheDocument();
```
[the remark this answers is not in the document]
Carla-Moz: These tests test the save button flow. We cover the cancel button flow before these tests:

Please search:

```
 it('should move back to the previously selected base and new revisions when Cancel is clicked', async () => {
    renderWithCompareResultsURL(
      <ResultsView title={Strings.metaData.pageTitle.results} />,
    );
```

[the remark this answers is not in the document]
Carla-Moz: If you don't think it covers all edit feature flow, please open a jira ticket and I'll handle it in another PR.

PR #668, opened 2024-06-04
Inject MaterialUI styles first so that typestyle styles can override them

julienw:
It's a small change that makes it possible to override MaterialUI styles with CSS modules (typestyle generates CSS modules).
See the documentation for this: https://mui.com/material-ui/integrations/interoperability/#css-modules

From my testing It doesn't seem to change anything at the moment, but I tested locally that this was working (I tried changing the background-color of CompareButton using `typestyle` for example).

I believe that this issue is why we have so many specific styles in https://github.com/mozilla/perfcompare/blob/beta/src/theme/components.js: because we couldn't override them in components directly or in https://github.com/mozilla/perfcompare/tree/beta/src/styles, we had to inject them there.

I'm still not super happy about the fact we use `typestyle` and instead we should probably [embrace material UI customization mechanism](https://mui.com/material-ui/customization/how-to-customize/), but this is a good stop gap so that we can move forward more easily in this last sprint.


src/__tests__/utils/test-utils.tsx
```diff
@@ -17,17 +22,19 @@ export function render(ui: React.ReactElement, themeConfig?: ThemeConfig) {
...
-          maxSnack={3}
-          autoHideDuration={6000}
-          action={(snackbarKey) => (
-            <SnackbarCloseButton snackbarKey={snackbarKey} />
-          )}
-        >
-          <Provider store={store}>{children}</Provider>
-        </SnackbarProvider>
-      </ThemeProvider>
+      <StyledEngineProvider injectFirst>
```
julienw, line 25: I added it in the tests for consistency, but it doesn't look like the snapshots look at the styles anyway...

Carla-Moz (approved):
Ah, this is great. Thanks!

PR #669, opened 2024-06-05
added base repository to compare over time component

Carla-Moz:
This PR closes: [Search: Add base repository component to Compare Over Time and update design](https://mozilla-hub.atlassian.net/browse/PCF-442)

- [x] Adds the base repository to the Compare Over Time component to capture baseRepo selection
- [x] Updates the OverTimeLoader.ts doc to include baseRep selection and adds a new error when baseRepo is missing
- [x] Updates SearchOverTime and CompareOverTime to match new design created by Dasha: moves the TimeRangeDropdown. 
- [x] Updates treeherder.ts to use baseRepo selection. 

Screenshot 2024-06-05 at 10 34 49

The widths of the dropdowns differ from the current Figma values but this will be cleaned up in future polishing PRs.

### DIFF
--- a/src/components/Search/SearchDropdown.tsx
+++ b/src/components/Search/SearchDropdown.tsx
@@ -20,6 +20,7 @@ interface SearchDropdownProps {
   searchType: InputType;
   repository: Repository['name'];
   labelIdInfo: string;
+  isOverTimeBase: boolean;
   onChange: (val: Repository['name']) => unknown;
 }
 
@@ -29,6 +30,7 @@ function SearchDropdown({
   searchType,
   repository,
   labelIdInfo,
+  isOverTimeBase,
   onChange,
 }: SearchDropdownProps) {
   const size = compact ? 'small' : undefined;
@@ -78,6 +80,7 @@ function SearchDropdown({
           className='dropdown-select'
           variant='standard'
           onChange={(e) => void handleRepoSelect(e)}
+          name={`${isOverTimeBase ? 'baseRepo' : ''}`}
         >
           {Object.keys(repoMap).map((key) => (
             <MenuItem
--- a/src/components/Search/SearchDropdown.tsx
+++ b/src/components/Search/SearchDropdown.tsx
@@ -20,7 +20,7 @@ interface SearchDropdownProps {
   searchType: InputType;
   repository: Repository['name'];
   labelIdInfo: string;
-  isOverTimeBase: boolean;
+  name?: string;
   onChange: (val: Repository['name']) => unknown;
 }
 
@@ -30,7 +30,7 @@ function SearchDropdown({
   searchType,
   repository,
   labelIdInfo,
-  isOverTimeBase,
+  name,
   onChange,
 }: SearchDropdownProps) {
   const size = compact ? 'small' : undefined;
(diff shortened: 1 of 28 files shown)


src/components/Search/SearchDropdown.tsx
```diff
@@ -20,6 +20,7 @@ interface SearchDropdownProps {
   searchType: InputType;
   repository: Repository['name'];
   labelIdInfo: string;
+  isOverTimeBase: boolean;
```
julienw, line 23: Suggestion, use an optional string prop instead:

```suggestion
  name?: string;
```

Then you don't need to specify it when not needed, and you can use it directly below.

Carla-Moz: Right! Much better. Thanks

src/components/Search/SearchOverTime.tsx
```diff
@@ -17,38 +18,69 @@ import {
...
+  const stringsBase =
+    Strings.components.searchDefault.overTime.collapsed.baseRepo;
+  const stringsNew =
+    Strings.components.searchDefault.overTime.collapsed.revisions;
 
   const [formIsDisplayed, setFormIsDisplayed] = useState(!hasEditButton);
 
+  const baseRepoStyles = style({
+    marginBottom: `${Spacing.Large}px`,
+    justifyContent: 'space-between',
```
julienw, line 65: I'm surprised this needs to be specified as CSS instead of using properties on the `<Grid>` component.

src/components/Search/SearchOverTime.tsx
```diff
@@ -17,38 +18,69 @@ import {
...
+    $nest: {
+      ' > div': {
+        flexGrow: 1,
```
julienw, line 68: My understanding is that you can use `<Grid item xs>` to make an item in the grid automatically grow.
See https://mui.com/material-ui/react-grid/#auto-layout

julienw (approved):
I don't have much to say about the code logic and organization, this is pretty good work!

My only concern is about the CSS part. It looks like you're fighting MUI's Grid API instead of embracing it. That is you're mixing using Grid with using CSS flex properties, where I believe the Grid API itself could handle that.
I think this is something we can clean up later though, so feel free to land without changing this, but I wanted to mention it in case there could be some easy changes to be done now. I know also that some existing styles (in src/styles or src/theme) might get in the way of using Grid fully, as it the positioning of the Search Revision dropdown, that's why I'm totlaly fine if you prefer to defer looking at this issue for later.

Carla-Moz:
I'll make a separate ticket to fix positioning of the search input and the results dropdown

src/components/Search/SearchOverTime.tsx
Carla-Moz: Thanks for mentioning this. Sometimes I forget MUI has these things covered. I'll look to stick with MUI in future patches and override unless absolutely necessary.

src/components/Search/SearchOverTime.tsx
Carla-Moz: Yes, I agree. I'll use MUI grid properties instead.

PR #670, opened 2024-06-06
reset view post submit

Carla-Moz:
This PR sets the view back to read-only post click of the Compare button in the Results View edit mode. Closes https://mozilla-hub.atlassian.net/browse/PCF-443

### DIFF
--- a/src/components/Search/CompareWithBase.tsx
+++ b/src/components/Search/CompareWithBase.tsx
@@ -104,9 +104,11 @@ function CompareWithBase({
     (baseRepository === 'try' && newRepository !== 'try') ||
     (baseRepository !== 'try' && newRepository === 'try');
 
-  const possiblyPreventFormSubmission = (e: React.FormEvent) => {
+  const onFormSubmit = (e: React.FormEvent) => {
     const isFormReadyToBeSubmitted = baseInProgressRev !== null;
+    setFormIsDisplayed(false);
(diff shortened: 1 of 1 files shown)


src/components/Search/CompareWithBase.tsx
```diff
@@ -104,9 +104,11 @@ function CompareWithBase({
     (baseRepository === 'try' && newRepository !== 'try') ||
     (baseRepository !== 'try' && newRepository === 'try');
 
-  const possiblyPreventFormSubmission = (e: React.FormEvent) => {
+  const onFormSubmit = (e: React.FormEvent) => {
     const isFormReadyToBeSubmitted = baseInProgressRev !== null;
+    setFormIsDisplayed(false);
     if (!isFormReadyToBeSubmitted) {
+      setFormIsDisplayed(true);
```
Carla-Moz, line 111: Had to set to `true` here because a Search View test breaks - the forms are hidden.

julienw: What about using this line 109:
```
setFormIsDisplayed(!isFormReadyToBeSubmitted)
```
so that you don't need to set it twice?

julienw (approved):
thanks!

src/components/Search/CompareWithBase.tsx
Carla-Moz: Ahh right! lol

PR #673, opened 2024-06-11
Filter columns

julienw:
This implements filtering by column.

The initial commits in the PR are #678 so please don't look at them here.
Then the commits are:
1. update the user-event library: indeed the new version expose a type that I'd like to use in the tests
2. Some moving around of types. This also removes some types that are not used.
3. Some more refactoring around how the short platform name and icon are computed. Indeed I want to reuse this in the filtering.
4. This is the meat of the PR, where the implementation for filtering happens. I tried to implement this so that we'd loop over the results only once for filtering. Ideally we'd do the "processResults" in the same loop too, but decided to leave this for future work, also to keep the code cleaner.
5. This changes the "expand icon" to a "filter icon". Indeed I wanted to provide an indicator when a filter is active on a column. It's only conveyed by the color so it's not perfect, but I still added some title and aria-label to try to provide a good accessibility.
6. and 7. are about tests.

It's not 100% perfect, especially the style for the filtering menu should be polished. Indeed some other styles are conflicting with that and I'd like to work on that in another PR.

[deploy preview](https://deploy-preview-673--mozilla-perfcompare.netlify.app/compare-results?baseRev=ddaa8beed4f649a50d0d6b8b67b373487e46dd56&baseRepo=mozilla-central&newRev=55142668d01174ee804cdee8c877f6d0212fd421&newRepo=mozilla-central&framework=1)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -47,7 +47,7 @@
         "@swc/jest": "^0.2.21",
         "@testing-library/jest-dom": "^5.16.4",
         "@testing-library/react": "^13.3.0",
-        "@testing-library/user-event": "^14.2.1",
+        "@testing-library/user-event": "^14.5.2",
         "@types/jest": "^28.1.4",
         "@types/jest-axe": "^3.5.4",
         "@types/material-ui": "^0.21.12",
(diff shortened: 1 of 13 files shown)


Carla-Moz (approved):
LGTM! We can clean up the styles in a future PR like you said.

PR #678, opened 2024-06-14
Do not add extra options flags if there are no extra options

julienw:
The problem is that `"".split(" ")` produces an array with one empty string. As a result we were adding an empty box.

When I wanted to add a test I realized that the test for ResultsTable was incorrect and was always generating an empty table. I fixed that as well. I ordered the commits so that the changes are well visible.

I also added 2 commits to change the eslint configuration:
1. so that eslint warns us about unused "disable" directives. This is useful for cleanup.
2. so that eslint doesn't warn anymore when somebody uses the "not null assertion" (the `!`) in tests only.

I reconize that this makes a PR with several quite unrelated things in it, I hope that the fact they're separated in several commits makes it OK enough. I did it this way because I used the eslint changes in the following commits.

[Deploy preview](https://deploy-preview-678--mozilla-perfcompare.netlify.app/compare-results?baseRev=ddaa8beed4f649a50d0d6b8b67b373487e46dd56&baseRepo=mozilla-central&newRev=55142668d01174ee804cdee8c877f6d0212fd421&newRepo=mozilla-central&framework=12) / [Beta branch](https://beta--mozilla-perfcompare.netlify.app/compare-results?baseRev=ddaa8beed4f649a50d0d6b8b67b373487e46dd56&baseRepo=mozilla-central&newRev=55142668d01174ee804cdee8c877f6d0212fd421&newRepo=mozilla-central&framework=12)

Before:
image

After:
image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -111,4 +111,10 @@ module.exports = {
       },
     },
   ],
+  // This property is specified both here in addition to the command line in
+  // package.json.
+  // The reason is that the property only warns but the command line option
+  // outputs errors, but the property is useful so that we have the information
+  // directly in editors.
+  reportUnusedDisableDirectives: true,
 };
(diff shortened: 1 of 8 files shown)


Carla-Moz (approved):
Thanks for the fix and relaxing eslint rules for tests

PR #681, opened 2024-06-19
PCF-479 Add an subtests icon where available for each revision rows

beatrice-acasandrei:
https://mozilla-hub.atlassian.net/browse/PCF-479


src/components/CompareResults/RevisionRow.tsx
```diff
@@ -204,6 +221,25 @@ function RevisionRow(props: RevisionRowProps) {
           <strong> {newRuns.length} </strong>
         </div>
         <div className='row-buttons cell'>
+          {header.parent_signature && (
+            <div className='subtests' role='cell'>
+              <div className='subtests-link-button-container'>
+                <Link
+                  href={getSubtestsCompareWithBaseLink(header)}
+                  aria-label='link to home'
```
beatrice-acasandrei, line 229: useless

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -204,6 +221,25 @@ function RevisionRow(props: RevisionRowProps) {
...
+                >
+                  <IconButton
+                    title={Strings.components.revisionRow.title.subtestsLink}
+                    color='primary'
+                    size='small'
+                  >
+                    <SubjectIcon />
```
beatrice-acasandrei, line 236: svg icon

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -194,6 +224,22 @@ function RevisionRow(props: RevisionRowProps) {
           <strong> {newRuns.length} </strong>
         </div>
         <div className='row-buttons cell'>
+          {result.parent_signature && (
```
beatrice-acasandrei, line 227:
```suggestion
          {result.has_subtests && (
```

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -39,8 +42,35 @@ const platformIcons: Record<PlatformShortName, ReactNode> = {
...
+&parentSignature=${result.signature_id}`;
+
+const getSubtestsLink = (view: string, result: CompareResultsItem) => {
+  if ((view == compareView)) return getSubtestsCompareWithBaseLink(result);
+  else
+  {
+    const queryString = window.location.search;
+    const urlParams = new URLSearchParams(queryString);
+    const interval = urlParams.get('selectedTimeRange');
+    return getSubtestsCompareOverTimeLink(result, interval);
```
beatrice-acasandrei, line 68: search params (similar to fetch)

src/components/Shared/Icons/SubtestsIcon.tsx
```diff
@@ -0,0 +1,21 @@
...
+    <SvgIcon viewBox='0 0 30 30'>
+      <g id='SVGRepo_bgCarrier' strokeWidth='0'></g>
+      <g
+        id='SVGRepo_tracerCarrier'
+        strokeLinecap='round'
+        strokeLinejoin='round'
+      ></g>
+      <g id='SVGRepo_iconCarrier'>
+        {' '}
+        <title>windows</title>{' '}
```
beatrice-acasandrei, line 14: Can be removed

src/components/Shared/Icons/SubtestsIcon.tsx
```diff
@@ -0,0 +1,21 @@
...
...44H4.9707V4.28572C4.9707 3.40207 5.68705 2.68572 6.5707 2.68572H12.5707C13.4544 2.68572 14.1707 3.40207 14.1707 4.28572V6.85715C14.1707 7.7408 13.4544 8.45715 12.5707 8.45715H6.5707C5.68705 8.45715 4.9707 7.74081 4.9707 6.85715V6.17144H2.1707V13.5429H4.39927V12.8572C4.39927 11.9735 5.11562 11.2571 5.99927 11.2571H15.4278C16.3115 11.2571 17.0278 11.9735 17.0278 12.8571V15.4286C17.0278 16.3122 16.3115 17.0286 15.4278 17.0286H5.99928C5.11562 17.0286 4.39927 16.3122 4.39927 15.4286V14.7429H1.5707C1.23933 14.7429 0.970703 14.4742 0.970703 14.1429V1.57144C0.970703 1.24006 1.23933 0.971436 1.5707 0.971436ZM5.59927 14.1436V15.4286C5.59927 15.6495 5.77836 15.8286 5.99928 15.8286H15.4278C15.6488 15.8286 15.8278 15.6495 15.8278 15.4286V12.8571C15.8278 12.6362 15.6488 12.4571 15.4278 12.4571H5.99927C5.77836 12.4571 5.59927 12.6362 5.59927 12.8572V14.1421C5.59927 14.1424 5.59928 14.1426 5.59928 14.1429C5.59928 14.1431 5.59927 14.1434 5.59927 14.1436ZM6.5707 3.88572C6.34979 3.88572 6.1707 4.06481 6.1707 4.28572V6.85715C6.1707 7.07806 6.34979 7.25715 6.5707 7.25715H12.5707C12.7916 7.25715 12.9707 7.07806 12.9707 6.85715V4.28572C12.9707 4.06481 12.7916 3.88572 12.5707 3.88572H6.5707Z'></path>{' '}
```
julienw, line 15: try to keep just the path, remove all the other things

src/components/Shared/Icons/SubtestsIcon.tsx
```diff
@@ -0,0 +1,21 @@
+import SvgIcon from '@mui/material/SvgIcon';
+
+function SubtestsIcon() {
+  return (
+    <SvgIcon viewBox='0 0 30 30'>
```
julienw, line 5: try with "0 0 24 24" for the viewbox, that's what the other icons have

julienw: Actually this is "0 0 18 18" in the original SVG, so use that :-)

src/logic/treeherder.ts
```diff
@@ -3,7 +3,7 @@ import { Framework, TimeRange } from '../types/types';
 
 // This file contains functions to request the Treeherder API
 
-export const treeherderBaseURL = 'https://treeherder.mozilla.org';
+export const treeherderBaseURL = 'http://localhost:8000';
```
beatrice-acasandrei, line 6: For testing purposes, will be removed later.

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -6,16 +6,21 @@ import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
...
 import { Strings } from '../../resources/Strings';
 import { Colors, Spacing, ExpandableRowStyles } from '../../styles';
 import type { CompareResultsItem, PlatformShortName } from '../../types/state';
+import { TimeRange } from '../../types/types';
 import { getPlatformShortName } from '../../utils/platform';
 import AndroidIcon from '../Shared/Icons/AndroidIcon';
 import LinuxIcon from '../Shared/Icons/LinuxIcon';
+import SubtestsIcon from '../Shared/Icons/SubtestsIcon';
 import WindowsIcon from '../Shared/Icons/WindowsIcon';
+import type { LoaderReturnValue } from './overTimeLoader';
```
julienw, line 23: nit: I think we need to have both LoaderReturnValue for type correctness:
```ts
import type { LoaderReturnValue as WithBaseLoaderReturnValue } from './loader';
import type { LoaderReturnValue as OverTimeLoaderReturnValue } from './overTimeLoader';
```
Then later:
```ts
const loaderData = useLoaderData() as WithBaseLoaderReturnValue | OverTimeLoaderReturnValue;
```

(see below as well)

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -39,6 +44,36 @@ const platformIcons: Record<PlatformShortName, ReactNode> = {
...
+newParentSignature=${result.new_signature_id}`;
+
+const getSubtestsLink = (result: CompareResultsItem) => {
+  const { view } = useLoaderData() as LoaderReturnValue;
+  if (view == compareView) return getSubtestsCompareWithBaseLink(result);
+  else {
+    const { intervalValue } = useLoaderData() as LoaderReturnValue;
+    return getSubtestsCompareOverTimeLink(result, intervalValue);
+  }
+};
```
julienw, line 75: I think I'd inline this function inside the component code directly:
```js
// Note that the return type is different depending on the view we're in
const loaderData = useLoaderData() as WithBaseLoaderReturnValue | OverTimeLoaderReturnValue;

const subtestsCompareLink = loaderData.view === compareView
  ? getSubtestsCompareWithBaseLink(result)
  : getSubtestsCompareOverTimeLink(result, loaderData.intervalValue);

// then use subtestsCompareLink below
```
What do you think? I think it's quite simple enough that it doesn't need an extra function.

src/components/Shared/Icons/AndroidIcon.tsx
```diff
@@ -3,15 +3,7 @@ import SvgIcon from '@mui/material/SvgIcon';
...
-      <g id='SVGRepo_bgCarrier' strokeWidth='0'></g>
-      <g
-        id='SVGRepo_tracerCarrier'
-        strokeLinecap='round'
-        strokeLinejoin='round'
-      ></g>
-      <g id='SVGRepo_iconCarrier'>
-        {' '}
-        <title>android</title>{' '}
+      <g>
```
julienw, line 6: even this `g` is not necessary. You can keep just the `path`.
(same below)

`g` means `group`, it is meant to group together several other elements to apply some common characteristics or use it as a reference later (if they have an `id`). No need for that here because we have just one path and we don't use it as a reference.

src/logic/treeherder.ts
julienw: Don't forget to change it back :-) unfortunately this prevents me from testing with the deploy preview :/

julienw (commented):
Sorry for the delay, I left a few comments, this looks OK from a quick overview !

The tests are not running (possibly because of the treeherder URL change), also there are linter (just prettier errors I believe) and typescript errors (the mock data need to be changed too)

src/logic/treeherder.ts
beatrice-acasandrei: I haven't merged the backend changes yet, since I want to do some more local tests.

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -65,6 +102,18 @@ function RevisionRow(props: RevisionRowProps) {
...
+  // Note that the return type is different depending on the view we're in
+  const loaderData = useLoaderData() as
+    | WithBaseLoaderReturnValue
+    | OverTimeLoaderReturnValue;
+  const subtestsCompareLink =
+    loaderData.view === compareView
+      ? getSubtestsCompareWithBaseLink(result)
+      : getSubtestsCompareOverTimeLink(
+          result,
+          (loaderData as OverTimeLoaderReturnValue).intervalValue,
```
beatrice-acasandrei, line 114: @julienw I'm not sure about this change - using `loaderData.internalValue` is causing typescript error _Property does not exist on type_

julienw: By changing the condition and inverting the ternary, this works:
```js
  const subtestsCompareLink =
    loaderData.view === compareOverTimeView
      ? getSubtestsCompareOverTimeLink(result, loaderData.intervalValue)
      : getSubtestsCompareWithBaseLink(result);
```

There might be something happening with the WithBaseLoaderReturnValue type, but let's not waste time on that.

src/components/CompareResults/loader.ts
```diff
@@ -228,6 +229,7 @@ export async function loader({ request }: { request: Request }) {
     newRepos,
     frameworkId,
     frameworkName,
+    view: compareView,
```
julienw, line 232: I realized that you also need to add the view in the return value above (with the fake data)

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -6,16 +6,22 @@ import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
...
 import type { CompareResultsItem, PlatformShortName } from '../../types/state';
+import { TimeRange } from '../../types/types';
 import { getPlatformShortName } from '../../utils/platform';
 import AndroidIcon from '../Shared/Icons/AndroidIcon';
 import LinuxIcon from '../Shared/Icons/LinuxIcon';
+import SubtestsIcon from '../Shared/Icons/SubtestsIcon';
 import WindowsIcon from '../Shared/Icons/WindowsIcon';
+import type { LoaderReturnValue as WithBaseLoaderReturnValue } from './loader';
+import type { LoaderReturnValue as OverTimeLoaderReturnValue } from './overTimeLoader';
 import RetriggerButton from './RetriggerButton';
```
julienw, line 25: A recent change at this line makes this file conflict but this should be easy to rebase and fix!

src/components/CompareResults/RevisionRow.tsx
beatrice-acasandrei: I still get the same error even with these changes.l

julienw:
This means the component needs to be rendered with renderWithRouter https://github.com/mozilla/perfcompare/blob/beta/src/__tests__/utils/test-utils.tsx#L43, you can look at examples in other tests.
(sorry I can't help much more at the moment)

esanuandra (approved):

PR #682, opened 2024-06-21
Overtime UI cleanup

Carla-Moz:
This PR cleans up several Compare Overtime UI bugs addressed in PR #674  

- [x] [remove the extra margin on labels](https://github.com/mozilla/perfcompare/pull/682/commits/b023733bf2f586ec74a48b118ab24417324cf77d)
- [x] [fix the the width of the search input and spacing between labels and dropdowns](https://github.com/mozilla/perfcompare/pull/682/commits/50cf7499fa04a3b81a3a0169bf83e3037aa3bd31)
- [x] [fix dark mode colors and icons (use svg for icon instead of github url)](https://github.com/mozilla/perfcompare/pull/682/commits/50cf7499fa04a3b81a3a0169bf83e3037aa3bd31)
- [x] [use visibility hidden or visible to display or hide the edit button ](https://github.com/mozilla/perfcompare/pull/682/commits/50cf7499fa04a3b81a3a0169bf83e3037aa3bd31)


src/components/Search/SearchOverTime.tsx
```diff
@@ -138,7 +138,7 @@ export default function SearchOverTime({
           <Grid
             item
             xs
-            style={{ maxWidth: '360px' }}
+            style={{ maxWidth: '364px' }}
```
julienw: do you know why it's 1px smaller than below?

src/components/Search/EditButton.tsx
```diff
@@ -1,25 +1,27 @@
...
+import pencilDark from '../../theme/img/pencil-dark.svg';
+import pencil from '../../theme/img/pencil.svg';
 
 interface EditButtonProps {
   onEditAction: () => void;
+  mode: string;
 }
 
 const baseComp = Strings.components.searchDefault.base;
-const editImgUrl = baseComp.editIcon;
```
julienw, line 10: shouldn't that be removed from the Strings object now?

src/components/Search/EditButton.tsx
```diff
@@ -1,25 +1,27 @@
...
-);
+export default function EditButton({ onEditAction, mode }: EditButtonProps) {
+  const buttonIcon = (
+    <img
+      id='edit-button-icon'
+      className='icon icon-edit'
+      src={mode === 'light' ? pencil.toString() : pencilDark.toString()}
+      alt='edit-icon'
+    />
+  );
```
julienw, line 24: I was thinking that it could be easier to use SvgIcon (from Mui), that also has this `color="primary"` attribute that uses the palette directly. But then we'd need to configure how we load SVG [per the documentation](https://mui.com/material-ui/icons/#component-prop) and [this documentation](https://react-svgr.com/docs/webpack/)... Currently we load SVG as `asset/inline`, what makes them a base64 data-url, and this is difficult to use inline because of that.

My general thought is that in SVG we can use `fill=currentColor` that makes it reuse the current color simply.. but that doesn't work if it's inside a `img`, this works only when it's inline (as far as I understand). MUI icons use this trick.

I think this would be good to do one of these eventually but out of scope for this last week.

src/styles/Fonts.ts
```diff
@@ -16,12 +15,10 @@ fontFace({
 
 const sharedFontStyles = {
   fontStyle: 'normal',
-  color: Colors.PrimaryText,
 };
 
 const sharedFontStylesDark = {
   fontStyle: 'normal',
-  color: Colors.PrimaryTextDark,
 };
```
julienw, line 22: They all look the same now, shouldn't we have just one version of the fonts values, instead of Light vs Dark?

src/styles/CompareCards.ts
```diff
@@ -62,6 +62,11 @@ export const CompareCardsStyles = (mode: string) => {
         '.edit-btn-wrapper': {
           display: 'flex',
           justifyContent: 'end',
+          $nest: {
+            button: {
+              color: isTrueLight ? Colors.LinkText : Colors.LinkTextDark,
+            },
+          },
```
julienw: I haven't folllowed closely how this button is styled, but could it be easier to use `variant="text" color="primary"` on the button component? Then `color="primary"` should follow the palette automatically. If that's the same colors of course.

julienw (approved):
Thanks, the spacings and dark mode look much better now!
Thanks for the individual commits too!

I left a few comments, but not that some of them are mostly suggestions for future maintenance. Others are suggestions for how to make things more maintenable now, and finally some possible clean up.

I noticed still this small spacing issue:
image
This is in the "non edit" state of the "editable" form.
I'm fine if you want to handle that in a next PR too.

src/styles/Fonts.ts
Carla-Moz: That's right. I overlooked this. Thanks!

src/components/Search/EditButton.tsx
Carla-Moz: Right!

src/components/Search/SearchOverTime.tsx
Carla-Moz: This is a typo. Thanks for catching it.

src/components/Search/EditButton.tsx
Carla-Moz: I'll turn this into a separate issue for a unique PR. But I agree.

src/styles/CompareCards.ts
Carla-Moz: Nice!

PR #683, opened 2024-06-24
Move around some code so that the token retrieval is more self-contained

julienw:
I moved around some code to make the authentication process more self contained and easier to work with.
Now the call to `getTaskclusterAccessToken` will always return an access token: sometimes it will return the one stored in localStorage, sometimes it will trigger the taskcluster authentication.

I added a few TODO to be done in future PRs too.

### DIFF
--- a/src/components/CompareResults/RetriggerButton.tsx
+++ b/src/components/CompareResults/RetriggerButton.tsx
@@ -1,33 +1,15 @@
 import RefreshOutlinedIcon from '@mui/icons-material/RefreshOutlined';
 import { IconButton } from '@mui/material';
 
-import { checkTaskclusterCredentials } from '../../logic/taskcluster';
+import { getTaskclusterAccessToken } from '../../logic/taskcluster';
 import { Strings } from '../../resources/Strings';
-
-function waitForStorageEvent(): Promise<void> {
-  return new Promise((resolve) => {
-    window.addEventListener(
-      'storage',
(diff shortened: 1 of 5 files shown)


src/components/CompareResults/RetriggerButton.tsx
```diff
@@ -1,33 +1,15 @@
 import RefreshOutlinedIcon from '@mui/icons-material/RefreshOutlined';
 import { IconButton } from '@mui/material';
 
-import { checkTaskclusterCredentials } from '../../logic/taskcluster';
+import { getTaskclusterAccessToken } from '../../logic/taskcluster';
 import { Strings } from '../../resources/Strings';
 
-function waitForStorageEvent(): Promise<void> {
```
julienw, line 7: Moved to credentials-storage

src/components/CompareResults/RetriggerButton.tsx
```diff
@@ -1,33 +1,15 @@
...
-}
-
 function RetriggerButton() {
   const onOpenModal = async () => {
-    checkTaskclusterCredentials();
-    await checkLocalStorage();
+    const accessToken = await getTaskclusterAccessToken();
+
+    // TODO do something with accessToken
+    console.log('We have an access token!', accessToken);
```
julienw, line 12: This will be replaced with the call to the retrigger function

julienw:
Reviewed with Andra over zoom!

PR #684, opened 2024-06-25
Add modals for sign in and retriggers

julienw:
5 commits:
* Commit 1 adds the PNG icon for Taskcluster, along with some configuration so that it can be used with Jest and Webpack
* Commit 2 adjusts the backdrop (the thing that's behind the modal to make the rest of page look darker) so that it's not darker in dark mode, but whiter instead
* Commit 3 implements the rough logic of moving around modals with some skeletons for the modals themselves
* Commit 4 implements the Sign in modal
* Commit 5 implements the Retrigger modal

Tell me what you think!

[Deploy preview](https://deploy-preview-684--mozilla-perfcompare.netlify.app/compare-results?baseRev=ddaa8beed4f649a50d0d6b8b67b373487e46dd56&baseRepo=mozilla-central&newRev=55142668d01174ee804cdee8c877f6d0212fd421&newRepo=mozilla-central&framework=1)

[Capture vidéo du 2024-06-28 18-28-51.webm](https://github.com/mozilla/perfcompare/assets/454175/85352c02-5cf1-4bd8-a1c7-5f622d178598)


Carla-Moz:
Screenshot 2024-06-28 at 10 51 31

Maybe we can add more line height here ? The 'g' in 'Sign' is touching the 't' in 'trigger'

src/components/CompareResults/RetriggerButton.tsx
```diff
@@ -1,37 +1,126 @@
...
+        <img src={TaskclusterLogo} role='presentation' width='48' />
       </Box>
-    </Modal>
+      <Typography id='sign-in-modal-title' component='h2' variant='h2'>
+        Sign into Taskcluster to re-trigger the comparison
+      </Typography>
+      <Typography>
+        To be able to retrigger a task within Taskcluster, it’s necessary that
+        you log in to it first. Then PerfCompare will retrigger the task on your
+        behalf.
```
Carla-Moz, line 116: I remember you mentioned something about getting rid of the `Strings` file in the resources folder but for consistency sake, would you mind putting this text there? If not, we should probably have a discussion on how to handle text in the future.

src/components/CompareResults/RetriggerButton.tsx
```diff
@@ -124,26 +129,80 @@ function SignInModal(props: SignInModalProps) {
...
-        </Typography>
-      </Box>
-    </Modal>
+      <Typography id='retrigger-modal-title' component='h2' variant='h2'>
+        Retrigger jobs
+      </Typography>
+      <Typography>
+        Choose how many clones of the base and new tasks will be started.
+      </Typography>
+      <form onSubmit={onFormSubmit}>
```
Carla-Moz, line 192: Just curious. Why did you choose the native form vs the one supplied by react-router?

Carla-Moz (approved):
Looks great! Thank you! Just had a few comments. I do suggest increasing the line height, however.

src/components/CompareResults/RetriggerButton.tsx
julienw: Good question!

The one supplied by react-router is special: it comes with a lot of built-in functionality that interacts with react-router (for navigation especially, but also pushing data to a server).
For this case, we only use a `form` element to encapsulate the selects and the submit button, and use its `submit` event for accessibility (I believe screenreaders have a functionality to submit a form so that they don't have to `<tab>` to the button). We don't need and we don't want all the functionality that comes with react-router.

Hope this makes sense!

src/components/CompareResults/RetriggerButton.tsx
julienw: My personal view is that the Strings file is "just another indirection" that serves no useful purpose at the moment.

I was thinking that the Strings file was for localization, but if we want to do localization in the future, we're better off using https://projectfluent.org/ instead of rolling on our own system. In that case having all strings in the Strings file could be of some help so that we could quickly see what we have to extract there.

However I don't think localizing this project would be useful because it's not user-facing but rather internal developer-facing.

In the mean time though I'll do as you suggest for consistency.

julienw:
> Maybe we can add more line height here ? The 'g' in 'Sign' is touching the 't' in 'trigger'

I agree but looking closer I think we should increase the line-height to 1.5 everywhere in the project, not just here, so I'll file a jira ticket for this.

PR #685, opened 2024-06-26
handling results loading with defer

Carla-Moz:
This PR handles the loading of the results using React Router's `defer ` method. See [here](https://reactrouter.com/en/main/guides/deferred)

Closes: 
[Add a spinner when fetching the compare results](https://mozilla-hub.atlassian.net/browse/PCF-429 )
[Add a loading spinner to Compare Button](https://mozilla-hub.atlassian.net/browse/PCF-430)

- [x] [Add the defer configuration](https://github.com/mozilla/perfcompare/pull/685/commits/719d4d831c80f66c8e7bb726b7c21e46baec4a08)
- [x] [Next, I had to update the types since the return type of the loaders was inaccessible according to typescript](https://github.com/mozilla/perfcompare/pull/685/commits/88d7d2102e412a9eae31b2872855285b9c838278)
- [x] [ handle same page updates on results page.](https://github.com/mozilla/perfcompare/pull/685/commits/b28f34ee669599a3e1061d18ef0afcef45d50f4c)


src/components/CompareResults/ResultsView.tsx
```diff
@@ -34,6 +45,21 @@ function ResultsView(props: ResultsViewProps) {
     document.title = title;
   }, [title]);
 
+  useEffect(() => {
+    const waitForData = async () => {
```
julienw, line 49: add `setLoading(true)` before the `await` and I believe you can get rid of `handleLoaderRefresh`.
It shouldn't do a "double render" at load time because the value is already `true`.

src/components/CompareResults/ResultsView.tsx
```diff
@@ -1,7 +1,15 @@
 import { useEffect } from 'react';
+import React from 'react';
```
julienw, line 2: nit: import `Suspense` and `useState` directly for a better tree shaking operation

src/components/CompareResults/ResultsView.tsx
```diff
@@ -34,6 +45,21 @@ function ResultsView(props: ResultsViewProps) {
...
   }, [title]);
 
+  useEffect(() => {
+    const waitForData = async () => {
+      await results;
+      setLoading(false);
+    };
+
+    void waitForData();
+  }, [results]);
```
julienw, line 55: So this is working, but you know me, I'm not a big fan of effects ;-)

An alternative implementation would use `useLoaderData` and `<Await>` directly in `CancelAndCompareButton` to change the label, and get rid of all the `loading` props.

I wasn't sure this would be much better so I wrote down advantages and drawbacks as I see them:


Advantages:
* less passing props down
* code self-contained in CancelAndCompareButton, so there's no need to duplicate in "with base" and "over time"
* more tightly integrated within React rendering pipeline (with Suspense)

Drawbacks:
* CancelAndCompareButton more tightly coupled to the loader mechanism (initially I thought about passing a promise by prop but I think this is more complex)


In the end I think the advantages outnumber the drawbacks, so I'd rather do this rather than the loading boolean prop.
What do you think?

src/components/CompareResults/ResultsView.tsx
```diff
@@ -34,6 +45,21 @@ function ResultsView(props: ResultsViewProps) {
...
+
+  const handleLoaderRefresh = () => {
+    setLoading(true);
+    const currentUrl = `/compare-results/${location.search}`;
+    fetcher.load(currentUrl);
```
julienw, line 60: Are you sure the explicit load is useful? My understanding is that react-router should handle that directly with its `Form` component already.

(same for the other view)

If we get rid of this + the loading boolean, we can also remove all of the changes related to handleRefresh.

julienw (commented):
Thanks, this works well and the UI looks good to me !

I think the code can be significantly simplified, but I haven't tried, so please tell me what you think!

src/components/CompareResults/loader.ts
```diff
@@ -228,7 +229,21 @@ export async function loader({ request }: { request: Request }) {
...
+  newRevs: string[];
+  newRevsInfo: Changeset[];
+  newRepos: Repository['name'][];
+  frameworkId: Framework['id'];
+  frameworkName: Framework['name'];
+};
+
+//had to be more explicit with the type because the defer
+//function returns a an inaccessible type
+export type LoaderReturnValue = Awaited<DeferredLoaderData>;
```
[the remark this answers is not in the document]
Carla-Moz: Oh gotcha, I initially had it removed but put it back in because of the promise in the results value.

src/components/CompareResults/ResultsView.tsx
Carla-Moz: Yeah that was cleanup I forgot to do.

src/components/CompareResults/ResultsView.tsx
Carla-Moz: We have to pass it down as a prop because the` CancelandCompareButton`component doesn't have access to the loader data on the Search Page so we can't use `useLoaderData` and since we have two loaders; it would be complicated.

src/components/CompareResults/ResultsView.tsx
Carla-Moz: So I looked up the `Form` API and it allows us to reload the document. I have this reload only for the edit feature. I think this solution is the best for the the edit feature. What do you think?

```
   <Form
          action='/compare-over-time-results'
          onSubmit={onFormSubmit}
          className='form-wrapper'
          aria-label='Compare over time form'
          reloadDocument={hasEditButton ?? true}
        >
//// rest of code
```

src/components/CompareResults/ResultsView.tsx
Carla-Moz: This is removed and no longer needed.

src/components/CompareResults/ResultsView.tsx
```diff
@@ -52,11 +77,29 @@ function ResultsView(props: ResultsViewProps) {
...
       <Grid container alignItems='center' justifyContent='center'>
         <Grid item xs={12}>
-          <ResultsMain results={results} />
+          <React.Suspense
+            fallback={
+              <Box
+                display='flex'
+                justifyContent='center'
+                alignItems='center'
+                mb={2}
```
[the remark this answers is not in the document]
Carla-Moz: I've moved the Await and Suspense to Results Main and it works better, cleaning up the two results views.

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -43,10 +47,23 @@ function ResultsMain(props: { results: CompareResultsItem[][] }) {
...
+      {loading ? (
+        <Box
+          display='flex'
+          justifyContent='center'
+          alignItems='center'
+          mb={3}
+          mt={3}
+        >
+          <CircularProgress />
+        </Box>
```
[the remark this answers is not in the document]
Carla-Moz: I have the form reload on edit so this will render as needed.

src/components/CompareResults/OverTimeResultsView.tsx
```diff
@@ -47,21 +73,37 @@ function ResultsView(props: ResultsViewProps) {
...
       <Grid container alignItems='center' justifyContent='center'>
         <Grid item xs={12}>
-          <ResultsMain results={results} />
+          <React.Suspense
+            fallback={
+              <Box
+                display='flex'
+                justifyContent='center'
+                alignItems='center'
+                mb={5}
```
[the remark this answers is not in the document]
Carla-Moz: Oh Box doesn't accept this prop. https://mui.com/system/properties/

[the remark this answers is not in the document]
Carla-Moz: I'll just keep `justifyContent = 'center'`; the others don't matter.

src/components/CompareResults/ResultsView.tsx
Carla-Moz: I've also removed any useEffects as well.

Carla-Moz: Actually I found a work around with the search page and to access the loader only on the Results Page using Await and Suspense. What do you think?

```
  const renderCompareButton = () => {
    if (hasEditButton) {
      const { results } = useLoaderData() as LoaderReturnValue;
      return (
        <React.Suspense
          fallback={
            <Button id='compare-button' color='primary' type='submit' disabled>
              Loading...
            </Button>
          }
        >
          <Await resolve={results}>
            <Button id='compare-button' color='primary' type='submit'>
              {label}
            </Button>
          </Await>
        </React.Suspense>
      );
    } else
      return (
        <Button id='compare-button' color='primary' type='submit'>
          {label}
        </Button>
      );
  };
```

src/components/CompareResults/ResultsView.tsx
julienw: Good find!

However I'm not so sure, because the full page is redrawing, as a user I find this quite distracting. I much prefered the previous UX.

julienw (approved):
Thanks, it's much simpler like that!
I'm not a big fan of using reloadDocument, but after trying an alternative but still simple way during 1h I couldn't find anything good and working as we'd like to.
I left a note in https://github.com/remix-run/react-router/issues/10816#issuecomment-2203752344 trying to advocate for an option for this, maybe we'll have something eventually...

src/components/CompareResults/ResultsView.tsx
julienw: "normal" react hooks can't be used in a if block, but I think the hooks from react router don't mind. So I don't mind either ;-)

julienw:
> Thanks, it's much simpler like that! I'm not a big fan of using reloadDocument, but after trying an alternative but still simple way during 1h I couldn't find anything good and working as we'd like to. I left a note in [remix-run/react-router#10816 (comment)](https://github.com/remix-run/react-router/issues/10816#issuecomment-2203752344) trying to advocate for an option for this, maybe we'll have something eventually... In the mean time let's move forward and we can always spend some time revisiting this later.

They replied to me and I think it's worth trying their suggestion. That is:
* remove the "defer()" call in our loader -- but keep the promise in the result object!
* still keep using Suspense/Await
* of course remove `reloadDocument` to test this

If this works, fine, land it. If this doesn't work, fine, land your current solution!

src/components/Search/CompareOverTime.tsx
```diff
@@ -191,6 +193,8 @@ function CompareOverTime({
           onSubmit={onFormSubmit}
           className='form-wrapper'
           aria-label='Compare over time form'
+          reloadDocument={hasEditButton ?? true}
+          replace
```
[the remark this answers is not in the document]
Carla-Moz: Yes, that's right.

Carla-Moz:
> > Thanks, it's much simpler like that! I'm not a big fan of using reloadDocument, but after trying an alternative but still simple way during 1h I couldn't find anything good and working as we'd like to. I left a note in [remix-run/react-router#10816 (comment)](https://github.com/remix-run/react-router/issues/10816#issuecomment-2203752344) trying to advocate for an option for this, maybe we'll have something eventually... In the mean time let's move forward and we can always spend some time revisiting this later.
> 
> They replied to me and I think it's worth trying their suggestion. That is:
> 
>     * remove the "defer()" call in our loader -- but keep the promise in the result object!
> 
>     * still keep using Suspense/Await
> 
>     * of course remove `reloadDocument` to test this
> 
> 
> If this works, fine, land it. If this doesn't work, fine, land your current solution!

Unfortunately their solution doesn't work. It behaves the same way with or without the `defer` with `Suspense/Await` and no `reloadDocument` - meaning the spinner and button loading states show only on initial load not same page user edits. I'll just keep my solution for now and we can revisit for a better one.

PR #686, opened 2024-06-27
PCF-432 Subtask Add low level functions and implement retrigger function

esanuandra:
This PR is part of [Implement the "login" intermediate popup if credentials are not available](https://mozilla-hub.atlassian.net/browse/PCF-432). We have a function for every request and also a first implementation for retrigger function that orchestrates the low level functions.

- fetchJobInformationFromJobId
- fetchDecisionTaskIdFromPushId
- fetchActionsFromDecisionTask
- fetchTaskInformationFromTaskId
- retrigger - that orchestrates everything


src/types/api.ts
```diff
@@ -15,3 +15,46 @@ export type APIPushResponse = {
...
+  build_architecture: string;
+  build_os: string;
+  build_platform: string;
+  build_platform_id: number;
+  build_system_type: 'taskcluster';
+  end_timestamp: number;
+  failure_classification_id: number; // probably an enum
+  id: number;
+  job_group_description: string;
+  job_group_id: number;
```
julienw, line 29: Mmm I'm curious whether job_group_id could be our decisionTaskId after all? We can look at it later.

src/logic/taskcluster.ts
```diff
@@ -152,3 +158,143 @@ export async function retrieveTaskclusterUserCredentials(
...
+    {},
+    {
+      taskGroupId: decisionTaskId,
+      taskId: null,
+      input: jobInfo.job_type_name,
+    },
+    actionsResponse.variables,
+  );
+
+  if (retriggerAction?.kind === 'hook') {
```
julienw, line 274: nit: rather do a early return:
```js
if (retriggerAction?.kind !== 'hook') {
  throw new Error('Missing hook kind for action');
}

// Carry on with the work
```

julienw: Also put the check right after the `actionsResponse.actions.find(` call, because this is related.

src/logic/taskcluster.ts
```diff
@@ -152,3 +158,143 @@ export async function retrieveTaskclusterUserCredentials(
...
+  // submit retrigger action to Taskcluster
+  const context = Object.assign(
+    {},
+    {
+      taskGroupId: decisionTaskId,
+      taskId: null,
+      input: jobInfo.job_type_name,
+    },
+    actionsResponse.variables,
+  );
```
julienw, line 272: Doesn't that work?
```js
const context = {
  taskGroupId: decisionTaskId,
  taskId: null,
  input: jobInfo.job_type_name,
  ...actionsResponse.variables,
}
```

julienw: Also I believe that `taskId` should be `jobInfo.task_id`, not null.

src/logic/taskcluster.ts
```diff
@@ -152,3 +158,143 @@ export async function retrieveTaskclusterUserCredentials(
...
+
+  if (retriggerAction?.kind === 'hook') {
+    // maybe use JSON.stringify
```
julienw, line 275: why not, but in that case remove it from `triggerHook`.

src/logic/taskcluster.ts
```diff
@@ -152,3 +158,143 @@ export async function retrieveTaskclusterUserCredentials(
...
+  const response = await fetch(url, options);
+
+  void checkTaskclusterResponse(response);
+
+  const json = (await response.json()) as { taskId: string };
+
+  return json.taskId;
+}
+
+export async function retrigger(rootUrl: string, repo: string, jobId: number) {
```
julienw, line 246: nit: please add a comment such as:
```js
// This function's goal is to retrigger an existing job from its jobId. It will
// call all appropriate APIs from taskcluster and treeherder.
```

src/logic/taskcluster.ts
julienw: Actually triggerHook's signature is currently incorrect, hookPayload is a string but it's really an object now. Calling JSON.stringify here (or rather in the call below) would make it correct.

src/logic/taskcluster.ts
```diff
@@ -152,3 +158,143 @@ export async function retrieveTaskclusterUserCredentials(
...
+
+  return json.scopes;
+}
+
+export async function triggerHook(
+  rootUrl: string,
+  accessToken: string,
+  hookGroupId: string,
+  hookId: string,
+  hookPayload: string,
```
julienw, line 222: because they are all strings, it would be better to use a single option object parameter, to avoid dumb mistakes such as swapping 2 parameters.

julienw (approved):
Just a few changes but this should be good to go after that. We'll make it work in the next patch :muscle: !

src/logic/taskcluster.ts
esanuandra: I did that initially but then changed my mind.

src/logic/taskcluster.ts
esanuandra: I get a typescript error and then have an error while calling jsone.

src/__tests__/utils/setupTests.ts
```diff
@@ -36,6 +36,43 @@ global.TextEncoder = TextEncoder;
...
+});
+const mockUse = jest.fn().mockImplementation(() => {
+  return {
+    createTask: jest.fn(() => {
+      return 'ACTION_TASKID';
+    }),
+  };
+});
+
+// after upgrading taskcluster to 39.1.1, fail to `import taskcluster-client-web`
```
esanuandra, line 58: this comment needs to be changed with the filed bug for taskcluster client web

src/logic/taskcluster.ts
```diff
@@ -152,3 +159,89 @@ export async function retrieveTaskclusterUserCredentials(
...
+  rootUrl: string;
+  repo: string;
+  jobId: number;
+  times: number;
+}) {
+  const { rootUrl, repo, jobId, times } = retriggerJobConfig;
+  const jobInfo = await fetchJobInformationFromJobId(repo, jobId);
+
+  const { push_id: pushId } = jobInfo;
+  const decisionTaskId = await fetchDecisionTaskIdFromPushId(repo, pushId);
```
julienw, line 199: In a follow-up, I think it would be more correct to do these calls in the RetriggerButton callback, and pass directly the `decisionTaskId` to this function.

This would be better in terms or architecture, so that the "taskcluster" module doesn't depend on the "treeherder" module; rather the caller (in this case RetriggerButton) orchestrates the calls to each of the modules.

Generally in terms of architecture, we want to avoid that 2 modules of the same layer call each other.

PR #689, opened 2024-07-15
moved typestyle calls outside the component

Carla-Moz:
Ran the profiler and it reduces jank

This PR closes the following issue:[The typestyle calls in the RevisionRow are too costly]( https://mozilla-hub.atlassian.net/browse/PCF-448)

Profiles: [beta](https://share.firefox.dev/46bM2lb) / [deploy preview](https://share.firefox.dev/3y28fFM)

To test:

1.) open PR deploy link: https://deploy-preview-689--mozilla-perfcompare.netlify.app/ 
2.) Select a base revision and a new revision(s) from the `Compare with a base` component
3.) Select the browsertime testing framework - heavy testing framework where you can see the most change in performance
3.) Click the `Compare` button
4.) The results page should perform better re loading vs the most current beta link: https://beta--mozilla-perfcompare.netlify.app/) / [deploy preview](https://profiler.firefox.com/from-browser/calltree/?globalTrackOrder=0wx6&hiddenGlobalTracks=1wx5&hiddenLocalTracksByPid=2792-1wcewx9~22443-0~38509-0~2816-0&thread=yg&v=10)

### DIFF
--- a/src/components/CompareResults/RevisionRow.tsx
+++ b/src/components/CompareResults/RevisionRow.tsx
@@ -25,6 +25,166 @@ import type { LoaderReturnValue as OverTimeLoaderReturnValue } from './overTimeL
 import RetriggerButton from './Retrigger/RetriggerButton';
 import RevisionRowExpandable from './RevisionRowExpandable';
 
+const revisionsRow = {
+  borderRadius: '4px 0px 0px 4px',
+  display: 'grid',
+  margin: `${Spacing.Small}px 0px`,
+  // Should be kept in sync with the gridTemplateColumns from TableHeader
+  gridTemplateColumns: '2fr 1fr 0.2fr 1fr 1fr 1fr 1fr 1fr 2fr 0.2fr',
+};
+
+const typography = {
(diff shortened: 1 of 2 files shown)


canova:
Thanks for the PR! Do you have an example deploy preview url that I can try?

Carla-Moz:
> Thanks for the PR! Do you have an example deploy preview url or STR that I can try?

Hi Nazim, here's the deploy preview url: [PR-689](https://deploy-preview-689--mozilla-perfcompare.netlify.app/)

You can find them above in the netlify comments above. Thanks!

canova (approved):
@Carla-Moz Thanks for the before and after profiles and thank you so much for the detailed explanation!

From the profiles it looks like there is like a 2 seconds improvement (from 14 to 12). Which is a nice start, but I also suspect that there are more areas to improve. It could be nice to capture another profile with non-minified JS (from localhost) so we can see more which functions are taking longer in the flame graph. But it's not the goal of this ticket, so let's land this PR :)

The code changes look great to me as well. Thank you!

PR #691, opened 2024-07-18
moved framework dropdown to results table

Carla-Moz:
This PR closes the following issue: https://mozilla-hub.atlassian.net/browse/PCF-421 

This PR is broken down into 3 commits:

- [x] [ Create a separate framework component for the Results Page table.](https://github.com/mozilla/perfcompare/pull/691/commits/a20f066a8df76f03e8d71abf40170b2019979d3d)
         I didn't want to force the new design and functionality unto the current `framework` component in `Search` so I created a new component in `Results` to reduce clutter, separate concerns, and easily manage the comps.

Screenshot 2024-07-23 at 12 20 37

- [x] [Use `useSearchParams` to change the url, refetch the data and rerender the page ](https://github.com/mozilla/perfcompare/pull/691/commits/25728d011392ed89d81d1d467d8391a3e742c4bf)
       I also added a _hidden input_ to keep the framework data in the url when the users change the revisions via the 'Edit entry' button. 

- [x] [Add test coverage](https://github.com/mozilla/perfcompare/pull/691/commits/1d788e3da8caaec935b31f770218a12dd96cefa1)

To Test:
1.) Open the PR's deployed link: https://deploy-preview-691--mozilla-perfcompare.netlify.app/
2.) Select a base revision and new revision(s) in the `Compare with a base` component
3.) After Results page has finished loading, change the framework in the menu bar above the Results table.
4. ) Spinner should appear as the table loads and disappears when finished loading new results.
5. ) Click `Edit entry` and change one of the revisions.
6. ) Press `Compare`
7. ) The framework value should not change and is not missing in the url. 
8. ) Repeat with `Compare over time`


src/components/Search/CompareWithBase.tsx
```diff
@@ -302,19 +294,24 @@ function CompareWithBase({
...
-              compact={false}
-              frameworkId={frameWorkId}
-              onChange={(event: SelectChangeEvent) => {
-                const id = +event.target.value as Framework['id'];
-                setframeWorkValue(id);
-              }}
-            />
+            {!hasEditButton && (
+              <FrameworkDropdown
+                frameworkId={frameWorkId}
```
esanuandra, line 308: Don't know if it's a type or not. Would it be better to keep the `framework` all lowercase everywhere?

Carla-Moz: Yes, I agree. I actually end up eliminating this state with the new 'framework' name in the next commit since it's not needed.

src/components/Search/FrameworkDropdown.tsx
```diff
@@ -74,8 +67,7 @@ function FrameworkDropdown({
 
   const styles = {
     container: style({
-      minWidth: `${compact ? '280' : '319'}px !important`,
-
+      minWidth: '319px !important',
```
canova, line 70: I'm not a fan of `!important` as it usually indicates a bigger problem with the design of cascading styles, but it's fine here I guess since we already had it :)

canova (commented):
Nice, it looks like a nice place to put this framework dropdown! But with these changes I see that "Download JSON" button gets a bit shorter and therefore it becomes two lines. 
Before:
Screenshot 2024-07-25 at 11 26 08 AM

After:
Screenshot 2024-07-25 at 11 26 02 AM

Making the "search results" form shorter might solve this. Currently it has 50% width, tried it with lower numbers and it looks like 47% fixes the issue. But also it becomes not a _nice_ number like 50% :) I'll leave it to you on the solution.

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -1,25 +1,46 @@
...
 import { style } from 'typestyle';
 
 import { useAppSelector } from '../../hooks/app';
 import { Colors, Spacing } from '../../styles';
 // import type { CompareResultsItem } from '../../types/state';
+import { Repository } from '../../types/state';
+import { Framework } from '../../types/types';
 import DownloadButton from './DownloadButton';
 import type { LoaderReturnValue } from './loader';
+import FrameworkDropdown from './ResultsFrameWorkDropDown';
```
beatrice-acasandrei, line 18:
```suggestion
import ResultsFrameworkDropdown from './ResultsFrameworkDropDown';
``` What do you think about this nit change so that it's more clear this is not the same component as the framework dropdown from the search page?

File: src/components/CompareResults/ResultsFrameWorkDropDown.tsx [hunk 4]
```diff
@@ -0,0 +1,62 @@
+import Box from '@mui/material/Box';
+import MenuItem from '@mui/material/MenuItem';
+import Select, { SelectChangeEvent } from '@mui/material/Select';
+
+import { frameworkMap } from '../../common/constants';
+import { Strings } from '../../resources/Strings';
+import { Spacing } from '../../styles';
+import type { Framework } from '../../types/types';
+
+const strings = Strings.components.searchDefault.sharedCollasped.framkework;
```
beatrice-acasandrei, line 10: nit
```suggestion
const strings = Strings.components.searchDefault.sharedCollasped.framework;
```

julienw:
> But with these changes I see that "Download JSON" button gets a bit shorter and therefore it becomes two lines.

I guess the "Download JSON" button (or rather its container) could be `flex: none`.

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -41,6 +62,33 @@ function ResultsMain() {
...
+      const values = searchParams.getAll(key);
+      // account for revs and repos being arrays
+      if (values.length > 1) {
+        searchParamsObject[key] = values;
+      } else {
+        searchParamsObject[key] = values[0];
+      }
+    });
+
+    setSearchParams({ ...searchParamsObject, framework: id.toString() });
```
julienw, line 88: Can this be replaced with: 
```js
searchParams.set("framework", id.toString());
setSearchParams(searchParams)
```
According to the type in https://reactrouter.com/en/main/hooks/use-search-params, `setSearchParams` accepts a `URLSearchParams` object. I'm passing the same instance here, which doesn't work for react states usually, but I believe this works here because this isn't a state.

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -41,6 +62,33 @@ function ResultsMain() {
...
+      | Repository['name'];
+  };
+
+  const onFrameworkChange = (event: SelectChangeEvent) => {
+    const id = +event.target.value as Framework['id'];
+    setFrameworkIdVal(id);
+    const searchParamsObject = {} as SearchParamsType;
+    searchParams.forEach((_, key) => {
+      const values = searchParams.getAll(key);
+      // account for revs and repos being arrays
```
julienw, line 80: But you're keeping just the first rev/repo with this code, right?
(anyway see below for another implementation)

src/components/CompareResults/ResultsFrameWorkDropDown.tsx
```diff
@@ -0,0 +1,62 @@
...
+  frameworkId,
+  onChange,
+}: FrameworkDropdownProps) {
+  return (
+    <Box display='flex' gap='1px' data-testid={'dropdown-select'}>
+      <Select
+        data-testid='framework-select'
+        label={strings.selectLabel}
+        value={`${frameworkId}`}
+        labelId='select-framework-label-results'
```
julienw, line 42: There's no component with this id, so this doesn't work.

But I believe we need a label (even if it's not in the design) as in https://mui.com/material-ui/react-select/#labels-and-helper-text, so that it doesn't take more space.

src/components/CompareResults/ResultsFrameWorkDropDown.tsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+const sortedFrameworks = sortFrameworks(frameworkMap);
+
+function ResultsFrameworkDropdown({
+  frameworkId,
+  onChange,
+}: FrameworkDropdownProps) {
+  return (
+    <Box display='flex' gap='1px' data-testid={'dropdown-select'}>
```
julienw, line 37: I believe this `Box` isn't useful, you can remove it.
With an extra label, you don't need the test id either, because you'll be able to target the select with its label.

src/components/CompareResults/ResultsFrameWorkDropDown.tsx
julienw, line 1: I don't like this duplication much. I think that the "meat" of the dropdown could be just one component. If we remove the Box and make this component some more configurable, I think it could be used in both cases (home page and results page).

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -1,25 +1,46 @@
...
+  useEffect(() => {
+    const waitForResults = async () => {
+      await results;
+      setTimeout(() => {
+        setIsLoading(false);
+      }, 50);
+    };
+
+    void waitForResults();
+  }, [isLoading]);
```
julienw, line 43: Not a huge fan of this change in this PR, also see below for an accessibility problem with the focus handling.

I think that if we cache the revision info like I mentioned, and keep the Await/Suspense thing _without the `defer`_ we'd properly see the spinner. We can deal with this later, but in the mean time I'd remove the custom spinner mechanism in this PR, because it adds some complexity and I'd rather remove it later when we manage to make Await/Suspense work properly.
What do you think?

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -55,6 +103,17 @@ function ResultsMain() {
...
+              {isLoading ? (
+                <Box display='flex' sx={{ margin: 'auto' }}>
+                  <CircularProgress />
+                </Box>
+              ) : (
+                <ResultsFrameworkDropdown
+                  frameworkId={frameworkIdVal}
+                  onChange={onFrameworkChange}
+                />
+              )}
```
julienw, line 115: This switching makes that a keyboard user loses focus, which isn't great for accessibility... so I'm not super happy about that either. I think we should use the global spinner mechanism rather than come up with a new mechanism just here.

And probably that global spinner should use the Await/Suspense...

src/components/Search/CompareOverTime.tsx
```diff
@@ -229,18 +216,29 @@ function CompareOverTime({
...
-            </div>
+            )}
+
+            {/**** Hidden Input to capture framework when user updates revisions ****/}
+            {hasEditButton && (
+              <Input
+                sx={{ visibility: 'hidden' }}
+                value={frameworkURL}
+                name='framework'
+              ></Input>
```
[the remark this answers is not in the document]
julienw: (same below)

src/components/CompareResults/ResultsFrameWorkDropDown.tsx
julienw: Also I see its width changes depending on what's selected.
I think the width of the  'All revisions' select could be reduced to leave some space for the framework dropdown.
(nice to have) You can also try to use some flex magic so that the 3 inputs (framework / revisions / download) are the same size?

julienw (commented):
Hey Carla, I haven't fully finished the review but here are some initial comments that should keep you busy for today :-)

src/components/CompareResults/ResultsFrameWorkDropDown.tsx
Carla-Moz: I'll make the "meat" of the component the Select component and move it to the `Shared` folder.

src/components/Search/CompareWithBase.tsx
```diff
@@ -302,19 +296,33 @@ function CompareWithBase({
...
             formIsDisplayed={formIsDisplayed}
           />
+
           <Grid
             item
             xs={2}
-            className={`${dropDownStyles.dropDown} ${bottomStyles.container}`}
+            display='flex'
+            justifyContent={hasEditButton ? 'flex-end' : 'space-between'}
+            className={`${dropDownStyles.dropDown}`}
```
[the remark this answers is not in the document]
Carla-Moz: Ah yes, I think I removed a string here and forgot to update. Good catch.

src/components/CompareResults/ResultsFrameWorkDropDown.tsx
Carla-Moz: Yes, true.

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: Oh nice! I mistakenly thought it would change the entire query to only 'framework' rather than preserving the old search params during the update. Good to know!

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: Yes, this is irrelevant with the fix below.

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: Conditional rendering is such a normal workflow in react; I'm sure there's a way to address the accessibility weakness here for losing focus. However, let me go ahead and test whether removing defer and caching the results will show the spinner for same page updates _without having to reload the page._

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: I hear what you're saying. I want to create a new branch to test caching the revisions and removing `defer` because from what I can see from the docs, `_Suspense_` was never meant for same page updates only on initial load. However, I'll be more than happy to revert to `_Suspense_` if the proposed changes above allow for same page updates. :)

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: Like I said above, I'd be happy removing this once I cache the revisions and remove defer, thus allowing the spinner to appear on same page updates.

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: Had to import the `ResultsTable` with lazy to show the spinner:

https://react.dev/reference/react/Suspense#displaying-a-fallback-while-content-is-loading

Carla-Moz:
@julienw, 

[resolved comments commit] (https://github.com/mozilla/perfcompare/pull/691/commits/a48205b03fd4a6d5187d50777f5efe68148e22c5#diff-637e0875b3be0006167a508cfbcfa8d1c3e4bb9a9eafbca7658714ed59d30d6a)

[memoize the revisions commit](https://github.com/mozilla/perfcompare/pull/691/commits/189c418a263ad63423fffc3e7460b8fd9111653b)

Same page updates are quicker, but no spinner. I guess we can leave it as is since the 'Compare/Cancel' buttons disappear when the results are resolved. *shrug* Let me know what you think. I could have done something wrong.

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -55,6 +80,19 @@ function ResultsMain() {
             <div className={styles.title}>Results</div>
             <div className={styles.content}>
               <SearchInput onChange={setSearchTerm} />
+
+              <FormControl sx={{ minWidth: 120 }}>
```
julienw, line 84: this minWidth doesn't seem to work, it's still too small.

I'd suggest to remove it for now and deal with that in a future PR. I really think we need to use flex properly on this line (maybe with the MUI element `Grid` even though that's not the only way)

julienw (commented):
Hey Carla, this is much better!
I'd suggest to remove the changes about defer etc from this PR, as well as a few other things I pointed out. Focus on the dropdown itself and try to not be distracted with the spinner. It's fine if there's no user feedback for now, I think this will be solved with my upcoming work, so don't worry about that.
Thanks again!

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -1,26 +1,34 @@
...
-// import type { CompareResultsItem } from '../../types/state';
+import { Framework } from '../../types/types';
+import FrameworkDropdown from '../Shared/FrameworkDropdown';
 import DownloadButton from './DownloadButton';
 import type { LoaderReturnValue } from './loader';
-import ResultsTable from './ResultsTable';
 import RevisionSelect from './RevisionSelect';
 import SearchInput from './SearchInput';
 
+const ResultsTable = lazy(() => import('./ResultsTable'));
```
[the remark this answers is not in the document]
Carla-Moz: Yes, I hear you. I had my misgivings using it initially, but there's was no way to trigger the `Suspense` spinner with the lazy import after you suggested removing 'defer' from the loaders -- that's also why I removed the `promise `from the `results` --  I'll explain more below. However, I'll revert this import.

src/components/Search/CompareWithBase.tsx
```diff
@@ -95,22 +95,16 @@ function CompareWithBase({
   const [formIsDisplayed, setFormIsDisplayed] = useState(!hasEditButton);
 
   const mode = useAppSelector((state) => state.theme.mode);
+  const [searchParams] = useSearchParams();
   const styles = CompareCardsStyles(mode);
   const dropDownStyles = SearchStyles(mode);
   const hasCancelButton = hasEditButton && formIsDisplayed;
+  const frameworkURL = searchParams.get('framework');
```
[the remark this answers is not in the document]
Carla-Moz: Because it's called `framework` in the query?  I'm a bit confused by this question?


```
...
  const baseRevFromUrl = url.searchParams.get('baseRev');
  const baseRepoFromUrl = url.searchParams.get('baseRepo') as
    | Repository['name']
    | null;
  const newRevsFromUrl = url.searchParams.getAll('newRev');
  const newReposFromUrl = url.searchParams.getAll(
    'newRepo',
  ) as Repository['name'][];
  const frameworkFromUrl = url.searchParams.get('framework');
...
```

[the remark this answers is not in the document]
Carla-Moz: Oh you mean the name of the const. Okay, lol. Yes, I can change the name.

src/components/Search/CompareWithBase.tsx
```diff
@@ -302,19 +296,33 @@ function CompareWithBase({
...
+            )}
+
+            {/**** Hidden Input to capture framework when user updates revisions ****/}
+            {hasEditButton && (
+              <input
+                type='hidden'
+                value={frameworkURL?.toString()}
+                name='framework'
+              ></input>
+            )}
```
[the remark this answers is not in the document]
Carla-Moz: Sometimes you confuse me, Julien, haha. In one review, you say ternary based renders are complex in react and this case, you say it's preferred. Since it's a nit, I'll leave it as is.

src/components/CompareResults/loader.ts
```diff
@@ -216,13 +214,14 @@ export async function loader({ request }: { request: Request }) {
...
+  const [results, baseRevInfo, ...newRevsInfo] = await Promise.all([
+    resultsPromise,
     baseRevInfoPromise,
     ...newRevsInfoPromises,
   ]);
 
-  return defer({
-    results: resultsPromise,
+  return {
+    results,
```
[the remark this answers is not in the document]
Carla-Moz: So I removed the `promise` here because I removed the `defer` mechanism and reverted to the previous return object - where we resolved the results in the `Promise.all`. But I can see this is why I didn't see the `Suspense ` spinner with the `Await `component regarding the results table.. Therefore, I'll put it back. I want you to understand my logic here and undersatnd I'm not this doing this willy-nilly lol.

src/components/Search/SearchFrameworkComp.tsx
```diff
@@ -0,0 +1,87 @@
...
+        {strings.selectLabel}
+        <Tooltip placement='top' title={strings.tooltip}>
+          <InfoIcon fontSize='small' className='dropdown-info-icon' />
+        </Tooltip>
+      </InputLabel>
+      <FrameworkDropdown
+        frameworkId={frameworkId}
+        labelId='select-framework-label'
+        mode={mode}
+        frameworkStyles={styles}
```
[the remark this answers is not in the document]
Carla-Moz: Ah, that was a mistake. Didn't mean to do that. I'll come up with better names next time. Thanks for catching

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -41,6 +49,23 @@ function ResultsMain() {
     }),
   };
 
+  const frameworkStyles = {
+    marginRight: `${Spacing.Medium}px`,
+    marginLeft: `${Spacing.Medium}px`,
```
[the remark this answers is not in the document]
Carla-Moz: Nice!

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -41,6 +49,23 @@ function ResultsMain() {
...
+  };
+
+  const frameworkProps = {
+    'aria-label': 'without label',
```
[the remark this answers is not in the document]
Carla-Moz: Understood.

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: I'll make a separate PR to fix the styles for this row.

src/components/CompareResults/ResultsView.tsx
```diff
@@ -16,9 +16,20 @@ interface ResultsViewProps {
...
 
+  // Memoize baseRevInfo, newRevsInfo, baseRepo
+  const { baseRevInfo, newRevsInfo, baseRepo } = useMemo(
+    () => ({
+      baseRevInfo: loaderData.baseRevInfo,
+      newRevsInfo: loaderData.newRevsInfo,
+      baseRepo: loaderData.baseRepo,
+    }),
+    [loaderData.baseRevInfo, loaderData.newRevsInfo, loaderData.baseRepo],
+  );
```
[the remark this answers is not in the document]
Carla-Moz: Looks like I used `useMemo` incorrectly here. I was under the wrong assumption that would it cache the revisions when say `framework` was updated. Will read the docs more closely next time.

Carla-Moz:
@julienw, resolved your comments [here](https://github.com/mozilla/perfcompare/pull/691/commits/664669817fc3632df010fe61fd8365fd5e948a8c). I think this is in a good shape and we can merge soon. Will make a separate PR for the styles of the row to fit all the dropdown better and for the tests.

src/components/Search/CompareWithBase.tsx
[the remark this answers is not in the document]
julienw: It may be my contraryness :o) 
In this specific case I felt like it was better mostly because it's clear it's one or the other, but also because each part is small.
in the end it's really a matter of preference so both work.

src/components/CompareResults/loader.ts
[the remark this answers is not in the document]
julienw: To make my remark clearer: by "awaiting" the result promise, there's no way the suspense fallback would appear: indeed the first render wouldn't even happen before the loader ends, but we need this first render for the suspense fallback to appear.

In short, here is the sequence of things:

1. the user presses "compare" (or change the value in the "framework" dropdown)
2. react router triggers a rerender with the same data (I believe it's a mistake but this is how it is)
3. react router runs the loader, and waits for its promise to resolve if it's a promise. (reminder that async functions are functions that return a promise)
4. if "defer" isn't used, react router triggers a new react render with the new data returned by the loader. That data can be anything, including a promise. Suspense + Await allows to wait for this promise and show a fallback before that. With the weird behavior I described in https://github.com/mozilla/perfcompare/pull/702#discussion_r1705274740.
5. If "defer" is used, react router waits for the promise(s) passed to defer to be resolved before triggering a new render. Suspense + Await is used only for the first render, not rerenders.

Hope this makes things clearer. I'm still learning how this works too, I find react-router made some weird choices.

src/components/CompareResults/ResultsView.tsx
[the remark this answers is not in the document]
julienw: Tell me if you want to discuss useMemo in a sync meeting :-)

src/components/Search/CompareWithBase.tsx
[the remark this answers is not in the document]
julienw: I didn't mean about the name of the variable, I meant that `CompareWithBase` gets `frameworkIdVal` directly as a prop, so I don't understand why you need to get it from the search params. My understanding is that `frameworkIdVal` already has the information we need? Am I understanding this correctly?

Carla-Moz:
That unfortunate. Closing this PR. Please see https://github.com/mozilla/perfcompare/pull/704  - has all the changes requested, green tests, and clean rebase

PR #693, opened 2024-07-24
Hide compare button on initial results page load

Carla-Moz:
A quick PR to hide the compare button on the initial load of the results page. Make it visible after user hits the `Edit entry `button.

Closes the following issue: https://mozilla-hub.atlassian.net/browse/PCF-411


canova (approved):
LGTM, thanks!

PR #694, opened 2024-07-25
Add some spacing between total runs and explain the letters with a title

canova:
I was looking at the results page and I had no idea what "B" and "N" was first in the results view. I had to go into the code to read the variable to understand.

This patch adds a `title` to these letters so people can hover over them to understand what they are.
Also it didn't have an adequate spacing between those values. Looked at the figma file and it does indeed have spacing. So I added it as well.

Before:
Screenshot 2024-07-25 at 12 58 55 PM

After:
Screenshot 2024-07-25 at 12 59 31 PM

### DIFF
--- a/src/components/CompareResults/RevisionRow.tsx
+++ b/src/components/CompareResults/RevisionRow.tsx
@@ -90,6 +90,7 @@ const stylesLight = {
       },
       '.total-runs': {
         backgroundColor: Colors.Background200,
+        gap: '8px',
       },
       '.row-buttons': {
         backgroundColor: Colors.Background200,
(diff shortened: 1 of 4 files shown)


src/components/CompareResults/RevisionRow.tsx
```diff
@@ -90,6 +90,7 @@ const stylesLight = {
       },
       '.total-runs': {
         backgroundColor: Colors.Background200,
+        gap: '8px',
```
julienw, line 93: you also need to add it to the dark section below

This repetition is a bit of a bummer, we should be able to reduce the duplication with eg a function taking the color as a parameter and returning the values. (not for this PR)

canova: Ah good catch, thanks! Updated it.

Carla-Moz (approved):
LGTM! Thank you for fixing.

PR #695, opened 2024-07-25
Test coverage for retrigger function

esanuandra:


src/components/CompareResults/Retrigger/RetriggerConfigModal.tsx
```diff
@@ -29,6 +29,7 @@ function RetriggerCountSelect({
         defaultValue={0}
         label={label}
         sx={{ height: 32 }}
+        data-testid={`${prefix}-retrigger-select`}
```
julienw, line 32: you don't probably need this, instead I believe you can use the label: https://testing-library.com/docs/queries/bylabeltext

src/__tests__/utils/setupTests.ts
```diff
@@ -75,6 +75,8 @@ jest.mock('taskcluster-client-web', () => {
   };
 });
 
+jest.mock('json-e', () => ({ jsone: jest.fn() }));
```
julienw, line 78: I'm curious, why is it necessary? I would have prefered that we don't mock this library if that's possible.

src/__tests__/CompareResults/Retrigger.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+  render,
+  screen,
+  within,
+} from '../utils/test-utils';
+
+jest.mock('../../utils/location');
+const mockedGetLocationOrigin = getLocationOrigin as jest.Mock;
+const result = getTestData().testCompareData[0];
+
+const setUpUserCredentions = () => {
```
julienw, line 18: nit:
```suggestion
const setUpUserCredentials = () => {
```

src/__tests__/CompareResults/Retrigger.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+      },
+    }),
+  );
+};
+
+describe('Retrigger', () => {
+  it('should display Sign In modal when there are no credentials', async () => {
+    (window.fetch as FetchMockSandbox).get('begin:https://somethingurl', {
+      ceva: 'response',
+    });
```
julienw, line 53: it this useful?

src/__tests__/CompareResults/Retrigger.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+    fireEvent.mouseDown(selectButton);
+
+    listbox = within(await screen.findByRole('listbox'));
+    fireEvent.click(await listbox.findByText('1'));
+
+    const triggerJobsButton = await screen.findByText('Retrigger');
+
+    await user.click(triggerJobsButton);
+
+    expect('something').toBe(null);
```
julienw, line 179: You can expect that the retrigger call has been made.

I think you can inspect all fetch calls with:
`console.log(window.fetch.mock.calls)`
So that you can then do a:
```
expect(window.fetch).toHaveBeenCalledWith(...)
```

src/__tests__/CompareResults/Retrigger.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+    );
+    (window.fetch as FetchMockSandbox).get(
+      'begin:https://firefoxci.taskcluster-artifacts.net/TeSt0FIBQyuzPDktQnTest/0/public/actions.json',
+      {
+        actions: [
+          { hookPayload: {}, kind: 'hook', name: 'retrigger-multiple' },
+        ],
+        variables: {},
+      },
+    );
```
julienw, line 113: we probably don't need that one, as the response is already received by the previous call. In the real life, the previous call is a redirection to that one, but I don't think we need to do it in the test.

src/__tests__/CompareResults/Retrigger.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+    (window.fetch as FetchMockSandbox).get(
+      'begin:https://firefoxci.taskcluster-artifacts.net/TeStSeCoNDuzPDktwoTest/0/public/actions.json',
+      {
+        actions: [
+          { hookPayload: {}, kind: 'hook', name: 'retrigger-multiple' },
+        ],
+        variables: {},
+      },
+    );
```
julienw, line 144: same as above, we probably don't need that one

src/__tests__/CompareResults/Retrigger.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+      {
+        '1234567': { id: 'TeSt0FIBQyuzPDktQnTest', run: '0' },
+      },
+    );
+    (window.fetch as FetchMockSandbox).get(
+      'begin:https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/TeSt0FIBQyuzPDktQnTest/artifacts/public%2Factions.json',
+      {
+        actions: [
+          { hookPayload: {}, kind: 'hook', name: 'retrigger-multiple' },
+        ],
```
julienw, line 101: it would be good to return the real thing, so that the call to json-e could work as expected.
To make it easier to use both here and below, you can put it in a constant above the test, maybe with a `const retriggerMultipleActionDefinition = ...`.

src/__tests__/CompareResults/Retrigger.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+  it('should retrigger one job for base revision and one job for new revision', async () => {
+    setUpUserCredentions();
+    // fetch requests for base revision info
+    (window.fetch as FetchMockSandbox).get(
+      'begin:https://treeherder.mozilla.org/api/project/mozilla-central/jobs/381594973/',
+      {
+        push_id: 1234567,
+      },
+    );
+    (window.fetch as FetchMockSandbox).get(
```
julienw, line 90: nit: you can chain the `.get` calls, so that you don't need to repeat `(window.fetch as FetchMockSandbox)`. It should make this definition a bit lighter.

julienw (commented):
Some more comments!

src/__tests__/utils/setupTests.ts
esanuandra: Screenshot 2024-07-30 at 2 17 06 PM
I have this error that I was trying to fix. I still don't have a solution yet.

src/__tests__/CompareResults/Retrigger.test.tsx
esanuandra: nope, I forgot to remove it from this test. Thanks!

src/__tests__/CompareResults/Retrigger.test.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+    const user = userEvent.setup({ delay: null });
+    await user.click(openModalButton);
+
+    const signInButton = await screen.findByText('Sign in');
+
+    expect(signInButton).toBeInTheDocument();
+  });
+
+  it('should display Retrigger modal when there are credentials', async () => {
```
julienw, line 141: I think this test is not useful, the next test does the same thing. So I'd remove it.

src/__tests__/CompareResults/Retrigger.test.tsx
```diff
@@ -0,0 +1,246 @@
...
+        {
+          actions: [{ ...retriggerMultipleActionDefinition }],
+          variables: {},
+        },
+      );
+
+    setUpUserCredentials();
+    render(<RetriggerButton result={result} />);
+
+    const openModalButton = await screen.findByTitle('retrigger jobs');
```
julienw, line 216: nit: findByRole('button', {name: 'retrigger jobs'})

src/__tests__/CompareResults/Retrigger.test.tsx
```diff
@@ -0,0 +1,246 @@
...
+
+    // select option 1 for new
+    const newSelect = await screen.findByLabelText('New');
+
+    fireEvent.mouseDown(newSelect);
+
+    listbox = within(await screen.findByRole('listbox'));
+    fireEvent.click(await listbox.findByText('1'));
+
+    const triggerJobsButton = await screen.findByText('Retrigger');
```
julienw, line 237: nit: findByRole('button', {name:'Retrigger'})

julienw (approved):
thanks!
A few small comments that should be easy to fix :)

PR #697, opened 2024-07-26
Use the Dialog component instead of the Modal component for the retrigger modals

julienw:
This fixes one issue that @canova reported to me: the various selection dropdowns were showing a grey backdrop behind them when opened:

image

I quickly considered that https://github.com/mozilla/perfcompare/pull/684/commits/15de4c6794737f92c9ddc07d36197719ef009f30 would be the issue. Indeed the selection dropdowns also insert a backdrop, and that change would make them all grey, while before they were transparent.

I tried to change it for just Modals... but turns out the selection dropdowns also use a Modal to position themselves!
Then in https://mui.com/material-ui/api/modal/, I've seen this small sentence:
> If you are creating a modal dialog, you probably want to use the [Dialog](https://mui.com/material-ui/api/dialog/) component rather than directly using Modal.

I missed it when I implemented it back then, before my PTOs.

So here it is: reimplementation of the modals using `Dialog` instead of `Modal`. It looks like it also has some better defaults for the Paper, so I could also remove some CSS. Also because of the better default for the dark mode, I could also remove the background color change for the backdrop, fixing the initial issue.

### Light mode:
Before:
image
image

After:
image
image

### Dark mode:
Before:
image
image

After:
image
image

### DIFF
--- a/src/components/CompareResults/Retrigger/CenteredModal.tsx
+++ b/src/components/CompareResults/Retrigger/CenteredModal.tsx
@@ -1,5 +1,5 @@
 import CloseIcon from '@mui/icons-material/Close';
-import { IconButton, Modal, Paper } from '@mui/material';
+import { IconButton, Dialog, Paper } from '@mui/material';
 
 type CenteredModalProps = {
   open: boolean;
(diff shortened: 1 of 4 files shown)


Carla-Moz (approved):
LGTM!

PR #698, opened 2024-07-30
Remove suspense comps and add lazy import for results table

Carla-Moz:
This [commit ](https://github.com/mozilla/perfcompare/pull/698/commits/c6d2d96e52a3b2f3d3d8364401a5f80cef66927c)removes the `Suspense` component since it doesn't fulfill our task of giving users feedback during the edit work flow. The current implementation requires reloading the page which isn't ideal and honestly, unnecessary. Therefore, I've also removed this from the form. 

My proposed solution adds an `isLoadingResults` state to `ResultView` to manage the rendering of the spinner upon initial loading of the `ResultsTable` and during any revision updates.  In this commit, I implement my fix only on the `CompareWithBase` component. The form in this component updates `isLoadingResults` to `true` until the results are ready. You'll see I added a delay in `ResultsView` since the `results` resolve too quickly for users to see any feedback. 

If the proposed fix is satisfactory upon review, I'll create a new commit to update `CompareOverTime` workflow.
Lastly, I've imported the `ResultTable` in `ResultsMain` using the [lazy API](https://react.dev/reference/react/lazy) to help improve performance. 
[
Profile for beta for initial loading of results table (browsertime framework) ](https://share.firefox.dev/4dk8cnS) / [Profile for my deployed PR '' ''](https://share.firefox.dev/4dm8cDN)

My patch reduces the first major jank by 3 seconds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "test:watch": "npm test -- --watch",
     "test:update": "npm test -- -u",
     "test-all": "run-p test format:check lint tsc",
-    "fix-all": "run-p test:update format lint:fix",
+    "fix-all": "run-p test:update format lint:fix tsc",
     "tsc": "tsc"
   },
   "eslintConfig": {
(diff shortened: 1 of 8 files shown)


Carla-Moz:
No need to review until I add the fix to frameworks dropdown move because this might end up becoming irrelevant.

PR #699, opened 2024-08-01
Add breadcrumbs to subtests

beatrice-acasandrei:


src/components/CompareResults/SubtestsResults/SubtestsBreadcrumbs.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { Spacing } from '../../../styles';
+
+export default function SubtestsBreadcrumbs() {
+  const breadcrumbs = [
+    <div key='1'></div>,
+    <Link key='2' href='/'>
+      Home
+    </Link>,
+    <Link key='3' href='/'>
```
beatrice-acasandrei, line 14: @julienw Here I was thinking of having a function which generates the previous URL. Do you have other suggestions for me?

src/components/CompareResults/SubtestsResults/SubtestsBreadcrumbs.tsx
```diff
@@ -0,0 +1,82 @@
...
+  ];
+
+  return (
+    <Stack spacing={2} direction='row' alignItems='center'>
+      <Breadcrumbs
+        separator={<ChevronLeftIcon fontSize='small' />}
+        aria-label='subtests breadcrumb'
+        sx={{
+          marginLeft: Spacing.xLarge,
+          marginBottom: `${Spacing.layoutXLarge + 4}px`,
```
julienw, line 73: When using `sx`, it's better to use plain numbers, they are then multiplied by "8" automatically.
But if you add the breadcrumbs in the other component like suggestd above, you'll probably just need the margin bottom (maybe with 2 or 3 as the value)

src/components/CompareResults/SubtestsResults/SubtestsBreadcrumbs.tsx
```diff
@@ -0,0 +1,82 @@
...
+      : getPreviousCompareOverTimeURL();
+  const breadcrumbs = [
+    <div key='1'></div>,
+    <Link key='2' href='/'>
+      Home
+    </Link>,
+    <Link key='3' href={allResultsURL}>
+      All results
+    </Link>,
+  ];
```
julienw, line 64: nit: Don't use a const, add them directly below.
(and then you don't need the key)

src/components/CompareResults/SubtestsResults/SubtestsBreadcrumbs.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const allResultsURL =
+    view === subtestsView
+      ? getPreviousCompareWithBaseURL()
+      : getPreviousCompareOverTimeURL();
+  const breadcrumbs = [
+    <div key='1'></div>,
+    <Link key='2' href='/'>
+      Home
+    </Link>,
+    <Link key='3' href={allResultsURL}>
```
julienw, line 61: it's not clear to me if we should use a URL or just a "window.close()", but let's keep it this way for now and we can adjust later.

julienw (commented):
Hey beatrice, sorry for the delay. Here are some comments for this PR, I think that should be good to go after that!

src/components/CompareResults/SubtestsResults/SubtestsBreadcrumbs.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const currentSearchParams = new URLSearchParams(location.search);
+  const interval = currentSearchParams.get('interval');
+  const baseRepo = currentSearchParams.get('baseRepo');
+  const newRev = currentSearchParams.get('newRev');
+  const newRepo = currentSearchParams.get('newRepo');
+  const framework = currentSearchParams.get('framework');
+
+  const params = new URLSearchParams({
+    baseRepo,
+    selectedTimeRange: interval,
```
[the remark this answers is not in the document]
beatrice-acasandrei: https://mozilla-hub.atlassian.net/browse/PCF-490

julienw (approved):
Looks good to me, thanks!
Just some nits but you can land once they're resolved

PR #700, opened 2024-08-01
Avoid a spurious rerender when changing the framework (and other things related to the router)

julienw:
This issue is especially visible when changing the framework... but only after removing the `reloadDocument` property, which I'm not doing in this deploy preview, so you won't be able to check this from the deploy preview. You can checkout the pull request locally and remove `reloadDocument` to check for yourself.

The issue is that the table is rerendered again even though nothing changed yet. From the react profiler, I found that it's rerendered because "context changed", that is the context from react router. This happens when using `useLoaderData` (see the issue I filed in https://github.com/remix-run/react-router/issues/11864).

So I fixed this through 2 commits:
1. Commit 1 is the simplest: it uses the "children function" style of `Await` so that `ResultsTable` can be a memoized component. Indeed `ResultsMain` gets the spurious rerender because of its use of `useLoaderData`.
2. Commit 2 is a bit more complex. It's goal is to remove the use of `useLoaderData` in `RevisionRow`, and more work was needed here. Maybe this could be reverted if https://github.com/remix-run/react-router/issues/11864 is fixed later...

### DIFF
--- a/src/__tests__/CompareResults/RevisionRow.test.tsx
+++ b/src/__tests__/CompareResults/RevisionRow.test.tsx
@@ -1,5 +1,6 @@
 import { ReactElement } from 'react';
 
+import { compareView } from '../../common/constants';
 import { loader } from '../../components/CompareResults/loader';
 import RevisionRow from '../../components/CompareResults/RevisionRow';
 import { Platform } from '../../types/types';
@@ -59,7 +60,7 @@ describe('<RevisionRow>', () => {
       } = getTestData();
 
       rowData.platform = platform as Platform;
-      renderWithRoute(<RevisionRow result={rowData} />);
+      renderWithRoute(<RevisionRow result={rowData} view={compareView} />);
       const shortNameNode = await screen.findByText(shortName);
       expect(shortNameNode).toBeInTheDocument();
       const previousNode = shortNameNode.previousSibling;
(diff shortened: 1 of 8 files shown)


julienw:
Ah the tests are failing, weird because I tested locally... let me fix that.

Carla-Moz:
@julienw 

`"From the react profiler, I found that it's rerendered because "context changed", that is the context from react router." `

Do you mind showing how you found that in the react profiler? I want to get better at reading profiles in general :P

julienw:
> @julienw
> 
> `"From the react profiler, I found that it's rerendered because "context changed", that is the context from react router." `
> 
> Do you mind showing how you found that in the react profiler? I want to get better at reading profiles in general :P

When you look at a recording in the react profiler, you see all the individual renders in the top right rectangle. You can click on one of them.
On the view below, all the colored item are the ones that rendered during that render cycle. By hovering them, the profiler tells you why they were rendered: props changed, state changed, context changed, parent rendered, etc. Sometimes you can get some more information (for props I think we know which props have changed), but for contexts not much. So then this requires some digging in the component itself, to guess what might use a context. 

That's about it!

beatrice-acasandrei (approved):

julienw:
Thanks beatrice for the review!

PR #701, opened 2024-08-01
Memoize the fetches for the revision information

julienw:
The goal for this patch is to memoize the result of fetching the revision information for specific revisions.

This has 2 commits:
1. Commit 1 is a preparation commit: this creates a specific function for this operation, so that it can be memoized in step 2.
2. Commit 2 uses the library https://github.com/planttheidea/moize to implement the memoization itself. I looked at many libraries and decided to use this one because of its simple yet effective API, with all the features we need here. Especially it makes it easy to use an option object for the function, and configure the number of calls to be memoized.
It can also be configured with an expiration, which could be useful for us later if we decide to use it for the results (maybe), and the cache can also be controlled from outside, which can be useful if we'd want to prime it from the revision selections in the home page.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "format": "^0.2.2",
         "json-e": "^4.7.0",
         "material-ui-popup-state": "^5.1.2",
+        "moize": "^6.1.6",
         "notistack": "^2.0.5",
         "process": "^0.11.10",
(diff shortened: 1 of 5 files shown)


Carla-Moz (approved):
I like how the memoize lib is simple and allows for so many configurations. Nothing else to add here. Thanks!

PR #702, opened 2024-08-02
Better use Suspense to get a spinner when data is loading

julienw:
This is ready for review !

Commit 1 is #708.
Commit 2 is the thing that you did in other PRs: remove `defer`
Commit 3 adds a key like mentioned in the comment below. Hopefully the comments in the code should be clear but tell me if we need more.
Commit 4 moves the results table header out of the Suspend area, so that it's visible while the data loads.
Commit 5 is a small fix for the types returned by the loaders.
Commit 6 fixes the DownloadButton after commit 4: because it's now outside of the Suspense/Await, it gets a promise in the normal results page, but the non-promise in the subtests page. This commit makes it work in both cases.

### DIFF
--- a/src/components/CompareResults/ResultsMain.tsx
+++ b/src/components/CompareResults/ResultsMain.tsx
@@ -83,7 +83,7 @@ function ResultsMain() {
             <RevisionSelect />
           </Grid>
           <Grid item xs>
-            <DownloadButton results={results} />
+            <DownloadButton resultsPromise={results} />
           </Grid>
         </Grid>
       </header>
--- a/src/components/CompareResults/ResultsMain.tsx
+++ b/src/components/CompareResults/ResultsMain.tsx
@@ -63,51 +63,49 @@ function ResultsMain() {
 
   return (
     <Container className={styles.container} data-testid='results-main'>
+      <header>
+        <div className={styles.title}>Results</div>
+        <Grid container className={styles.content} spacing={2}>
+          <Grid item md={6} xs={12}>
+            <SearchInput onChange={setSearchTerm} />
+          </Grid>
+          <Grid item xs>
+            <FormControl sx={{ width: '100%' }}>
+              <FrameworkDropdown
+                frameworkId={frameworkIdVal}
+                size='small'
+                variant='outlined'
+                onChange={onFrameworkChange}
+              />
+            </FormControl>
+          </Grid>
+          <Grid item xs>
+            <RevisionSelect />
+          </Grid>
+          <Grid item xs>
+            <DownloadButton results={results} />
+          </Grid>
+        </Grid>
+      </header>
       {/* Using a key in Suspense makes it that it displays the fallback more
           consistently.
           See https://github.com/mozilla/perfcompare/pull/702#discussion_r1705274740
           for more explanation (and questioning) about this issue. */}
       <Suspense
         fallback={
-          <Box display='flex' justifyContent='center'>
+          <Box display='flex' justifyContent='center' sx={{ marginTop: 3 }}>
             <CircularProgress />
           </Box>
         }
         key={generation}
       >
         <Await resolve={results}>
-          {(resolvedResults: CompareResultsItem[][]) => (
-            <>
-              <header>
-                <div className={styles.title}>Results</div>
-                <Grid container className={styles.content} spacing={2}>
-                  <Grid item md={6} xs={12}>
-                    <SearchInput onChange={setSearchTerm} />
-                  </Grid>
-                  <Grid item xs>
-                    <FormControl sx={{ width: '100%' }}>
-                      <FrameworkDropdown
-                        frameworkId={frameworkIdVal}
-                        size='small'
-                        variant='outlined'
-                        onChange={onFrameworkChange}
-                      />
-                    </FormControl>
-                  </Grid>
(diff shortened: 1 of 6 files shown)


src/components/CompareResults/ResultsMain.tsx
```diff
@@ -52,6 +52,7 @@ function ResultsMain() {
             <CircularProgress />
           </Box>
         }
+        key={Date.now()}
```
julienw, line 55: I'm not fully sure this is necessary, we might revisit later when we'll manage to cut the rendering time down.

julienw: The rationale is that the Suspense component will be remounted, and so doesn't keep a state. It seemed like it improved things with my limited testing...

julienw:
The tests are not passing but I won't have time to fix them today. Just wanted to push that out to get feedback from y'all.

Carla-Moz:
```
    Received
           1
            - [Error: Error when requesting treeherder: Treeherder request error]
            + [Error: Uncaught [TypeError: Cannot read properties of undefined (reading 'id')]],
           2: [TypeError: Cannot read properties of undefined (reading 'id')]
           3
            - [Error: Error when requesting treeherder: Treeherder request error]
            + [Error: Uncaught [TypeError: Cannot read properties of undefined (reading 'id')]],
```

I got this error too when I removed the 'defer' in my frameworks dropdown PR. But removing 'defer' allows the spinner to appear on framework change in my PR post rebase. Note, however, this error appeared before I rebased with your merged memoization PRs. The network call are all green so I'm sot sure as the error doesn't give enough details. :/

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: So wait, without this key, the spinner was _not_ appearing when doing updates (changing revisions and frameworks) on the Results page?

Carla-Moz: I ask because after rebasing and resolving your comments on my PR, the spinner appears on updates and framework updates.

Carla-Moz (commented):
Hi Julien! I have a question and I'll approve once answered. Not sure if this key is needed but need your confirmation because I could be missing something.

src/components/CompareResults/ResultsMain.tsx
julienw: @Carla-Moz the fact I'm using "key" is backed by my experiment that you can find in https://codepen.io/julienw/pen/MWMooEB

This experiment is built to replicate a bit how perfcompare behaves currently:
* the loader itself awaits for a timeout of 500ms, this models the revision info fetches
* it returns a promise that resolves 200ms later. This models the compare results fetch.

These timings are important, because the behavior change if you change them (more on that below).

You can see that Suspense has "key" in it. If you press "refresh", we can see "Loading results".
However if you remove the "key", save and wait for the reload, then "refresh" doesn't show "Loading results" anymore.


This behavior is different if you tweak the timings. For example if you totally remove the initial timeout of 500ms, then the fallback always appears, even without the `key`.
Similarly, if you increase the returned promise timeout to more than 500ms, then you see it as well.
My guess (that should be verified by looking at the code) is that Suspense waits for at least the time for the loader to return before showing the fallback. But I'm not sure which library is responsible for that, is that react-router-dom or React itself, or the interaction of them both? I've no idea.

julienw:
> I ask because after rebasing and resolving your comments on my PR, the spinner appears on updates and framework updates.


Thanks to the memoizing of the revision info from PR #701, when only the framework changes, then the revision info fetches are near instant and therefore the loader promoses are also near instants. Then the results fetch takes a significantly bigger amount of time to return, and therefore the spinner appears most of the time indeed.

But it doesn't (always) appear when changing the revisions in edit mode.

Carla-Moz (approved):

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: Thanks for explaining. That makes a lot more sense.

julienw:
dang, by using the key that change at each render, we get back the problem that I solved in #700 grrr. I'll try to find a better key... or maybe more memoization

julienw:
> dang, by using the key that change at each render, we get back the problem that I solved in #700 grrr. I'll try to find a better key... or maybe more memoization

I solved that by using an incremented value returning from the loaders. This seems to work a lot better now.

Carla-Moz (approved):
This looks and works so much better. Thank you!

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -63,46 +63,49 @@ function ResultsMain() {
...
+        </Grid>
+      </header>
+      {/* Using a key in Suspense makes it that it displays the fallback more
+          consistently.
+          See https://github.com/mozilla/perfcompare/pull/702#discussion_r1705274740
+          for more explanation (and questioning) about this issue. */}
       <Suspense
         fallback={
-          <Box display='flex' justifyContent='center'>
+          <Box display='flex' justifyContent='center' sx={{ marginTop: 3 }}>
```
julienw, line 96: @Carla-Moz I bumped this to 3 instead of the previous 2, so that it's the same as the table margin.

PR #705, opened 2024-08-07
Results: Description text for the platforms needs to include the versions

beatrice-acasandrei:


src/components/CompareResults/RevisionRow.tsx
```diff
@@ -292,7 +293,11 @@ function RevisionRow(props: RevisionRowProps) {
       >
         <div className='platform cell' role='cell'>
           <div className='platform-container'>
+          <Tooltip placement='bottom' title={platform} arrow>
+            <div>
```
beatrice-acasandrei, line 297: needed because the child of a Material-UI Tooltip must be able to hold a ref

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -292,7 +293,11 @@ function RevisionRow(props: RevisionRowProps) {
       >
         <div className='platform cell' role='cell'>
           <div className='platform-container'>
+          <Tooltip placement='bottom' title={platform} arrow>
```
beatrice-acasandrei, line 296: @julienw Sparky has asked about this ticket yesterday - wondering if it would it be ok to have it implemented like this

julienw: why not but I'd put the tooltip over both the icon + the shortname
and then you don't the extra div below.

julienw (approved):
looks good to me , thanks!

PR #706, opened 2024-08-07
Used Grid to fix results header

Carla-Moz:
This PR fixes the header styles for the Results Table using `Grid`

Screenshot 2024-08-07 at 09 27 57


src/components/Shared/FrameworkDropdown.tsx
```diff
@@ -7,7 +7,6 @@ import type { Framework } from '../../types/types';
 interface FrameworkDropdownProps {
   frameworkId: Framework['id'];
   labelId?: string;
-  sxStyles?: React.ComponentProps<typeof Select>['sx'];
```
julienw, line 10: thanks for the cleanup!

Carla-Moz: For sure!

PR #708, opened 2024-08-08
Fix the tests in App.test.tsx so that the revision info return at lea…

julienw:
…st one result

@Carla-Moz, I believe this is what triggered the weird errors when removing `defer` and using `Suspense`. I have no idea why this isn't happening without this change...

### DIFF
--- a/src/__tests__/App.test.tsx
+++ b/src/__tests__/App.test.tsx
@@ -8,13 +8,12 @@ import { render, screen, FetchMockSandbox } from './utils/test-utils';
 describe('App', () => {
   beforeEach(() => {
     const { testData } = getTestData();
-    (global.fetch as FetchMockSandbox)
-      .get('glob:https://treeherder.mozilla.org/api/project/*/push/*', {
-        results: [],
-      })
-      .get('begin:https://treeherder.mozilla.org/api/project/', {
+    (global.fetch as FetchMockSandbox).get(
+      'begin:https://treeherder.mozilla.org/api/project/',
+      {
         results: [testData[0]],
-      });
+      },
+    );
   });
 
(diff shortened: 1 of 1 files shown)


src/__tests__/App.test.tsx
```diff
@@ -88,7 +87,6 @@ describe('App', () => {
           'Error when requesting treeherder: (500) Internal Server Error',
         ),
       );
-      expect(console.error).toHaveBeenCalledTimes(14);
```
julienw, line 91: I find these console.error checks to be brittle (because the errors are output in several places, with Await, React, etc) and decided to remove them here and below. I kept the ones that are more consistent.

src/__tests__/App.test.tsx
```diff
@@ -8,13 +8,12 @@ import { render, screen, FetchMockSandbox } from './utils/test-utils';
 describe('App', () => {
   beforeEach(() => {
     const { testData } = getTestData();
-    (global.fetch as FetchMockSandbox)
-      .get('glob:https://treeherder.mozilla.org/api/project/*/push/*', {
-        results: [],
-      })
```
julienw, line 14: When requesting a single hash, we should always get a single result. And our code relies on that (maybe we should change our code that that's another story).

src/__tests__/App.test.tsx
Carla-Moz: They really are. Thanks for removing them.

src/__tests__/App.test.tsx
Carla-Moz: Hmm, that's something to keep note of in the future.

Carla-Moz (approved):
Thanks for the fix!

PR #710, opened 2024-08-08
Use replicate values when available for graphs.

gmierz:
This patch changes the graphs to use the replicates when they are available. There are certain situations where this doesn't occur (such as for summary-level values). Only try runs, and mozilla-central speedometer3 test runs have these replicates available to them.

It depends on the changes in this patch for the Treeherder API: https://github.com/mozilla/treeherder/pull/8165


gmierz:
@julienw not entirely sure how this should be implemented, but it works for me when I have tests with, and without replicates.

gmierz:
To test locally, you'll need to setup a local DB with a couple try runs with performance numbers ingested, and apply the patch from here https://github.com/mozilla/treeherder/pull/8165. Then change the API URL in treeherder.ts to point to the `localhost:PORT` running Treeherder.

src/components/CompareResults/Distribution.tsx
```diff
@@ -33,6 +35,7 @@ function Distribution(props: DistributionProps) {
     name: 'Base',
     median: baseMedian,
     values: baseRuns,
+    replicates: baseRunsReplicates,
```
julienw, line 38: I'd do this instead:
```suggestion
    values: baseRunsReplicates.length ? baseRunsReplicates : baseRuns,
```

And remove all the other changes.

I understand you wanted to keep both values but it doesn't seem useful in this component.

julienw: (same for the "new" of course)

julienw (commented):
I didn't try locally but this looks reasonable! I left a suggestion for a possibly simpler implementation, tell me what you think.

Our mock data in tests need to be changed, tell me if you'd like me to do it because it's a bit cumbersome.

gmierz:
@julienw I think I'm going to need to change the tests to include the new replicates fields in either case, unless you see something different that we could do?

julienw:
> @julienw I think I'm going to need to change the tests to include the new replicates fields in either case, unless you see something different that we could do?

Yes that's it!

That's why I suggested yesterday that we could make the parameter optional so that it would be absent when there's no replicates. But if it's not optional then we need to add it in every test data.

I see that the properties are missing from the type `CompareResultItem` in `src/types/state.ts` currently, I believe it was lost during our git manipulations yesterday.
Then typing `npm run tsc` should make it obvious where it's missing in tests. It will probably be in `src/__tests__/utils/fixtures.ts`

It would be good to add a test covering replicates too. This could happen in the `ResultsView.test.tsx` test by copying part of `Should display Base, New and Common graphs with tooltips`, amending the data in `testCompareDataWithMultipleRuns` to have one with replicates and opening the matching row. But I can help on this one once the other work is done!

src/__tests__/CompareResults/ResultsView.test.tsx
```diff
@@ -159,7 +159,7 @@ describe('Results View', () => {
     expect(graphTitle).toBe('Runs Density Distribution');
   });
 
-  it('should make blobUrl available when "Download JSON" button is clicked', async () => {
+  it('should make blobUrl available when Download JSON" button is clicked', async () => {
```
julienw, line 162: looks like a typo change, you removed the `"` :-)

src/__tests__/CompareResults/ResultsView.test.tsx
```diff
@@ -236,4 +236,142 @@ describe('Results View', () => {
     );
     expect(sessionStorage.taskclusterUrl).toBe(windowOpenUrl.origin);
   });
+
+  it('Should display Base, New and Common graphs with replicates', async () => {
```
julienw, line 240: Thanks for the new tests!

julienw (approved):
This looks good to me, but this depends on the new server, so let's wait until the new version of the server is deployed before merging.
Thanks!

src/__tests__/CompareResults/ResultsView.test.tsx
gmierz: No idea how that happened, but thanks for catching it! I've fixed it in the current change.

gmierz:
@julienw the replicates patch is live on the treeherder backend now! See an example of this patch in action here: https://deploy-preview-710--mozilla-perfcompare.netlify.app/compare-results?baseRev=d0d28f60528be531682bba8e03025f47a6b87647&baseRepo=try&newRev=2d683f4c1f0a93888eab3a651eb319d3818695e6&newRepo=try&framework=13

julienw:
Nice, thanks!
I feel like we need to have a hint on the main row that there are more results inside, but I think that's good enough for now!
Merging

PR #711, opened 2024-08-09
PCF-484 Add the retrigger test functionality for subtests

esanuandra:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/PCF-484)

https://deploy-preview-711--mozilla-perfcompare.netlify.app/subtestsCompareWithBase?baseRev=7742b23a126df264dec9eff7c299afe33c6ca74f&baseRepo=mozilla-central&newRev=2e86cbfb869e1b6abc6ee9cee177f4f7484b7cb2&newRepo=mozilla-central&framework=1&baseParentSignature=4769486&newParentSignature=4769486

### DIFF
--- a/src/components/CompareResults/Retrigger/RetriggerButton.tsx
+++ b/src/components/CompareResults/Retrigger/RetriggerButton.tsx
@@ -3,6 +3,7 @@ import { useState } from 'react';
 import RefreshOutlinedIcon from '@mui/icons-material/RefreshOutlined';
 import { IconButton, Button } from '@mui/material';
 import { useSnackbar } from 'notistack';
+import { style } from 'typestyle';
 
 import {
   getTaskclusterCredentials,
@@ -23,6 +24,20 @@ import { RetriggerSignInModal } from './RetriggerSignInModal';
 
 type Status = 'pending' | 'signin-modal' | 'retrigger-modal';
 
+const styles = {
+  retrigger: style({
+    fontFamily: 'SF Pro',
+    fontSize: '11px',
+    fontWeight: 590,
+    lineHeight: '16.5px',
+    textAlign: 'left',
+    width: '111px',
+    height: '25px',
+    padding: '4px 8px 4px 8px',
+    gap: '4px',
+    borderRadius: '4px 0px 0px 0px',
+  }),
+};
 interface RetriggerButtonProps {
   result: CompareResultsItem;
 }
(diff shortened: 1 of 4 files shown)


src/components/CompareResults/Retrigger/RetriggerButton.tsx
```diff
@@ -163,6 +180,9 @@ function RetriggerButton({ result }: RetriggerButtonProps) {
         onClick={() => void onRetriggerButtonClick()}
       >
         <RefreshOutlinedIcon />
+        {baseParentSignature || newParentSignature ? (
+          <span className={styles.retrigger}>Retrigger test</span>
+        ) : null}
```
julienw, line 185: This doesn't look right (eg the hover style), we really need a normal button (variant "text" I believe) with a label in this case, such as https://mui.com/material-ui/react-button/#buttons-with-icons-and-label

Therefore I'd create 2 simple components for just the buttons, taking "onClick" as a prop. You can create them in this file. `RetriggerIconButton` and `RetriggerFullButton` for example.
Then `RetriggerButton` would take the component to use as a prop `RetriggerButtonComponent`.
Finally each user of the `RetriggerButton` component could specify a different one.

julienw: I thought about this some more and I think that part of my suggestion above is a bit overkill for our case here it's used only twice.
So here is my other simpler suggestion:
* make RetriggerButton take a `variant` prop taking 2 values `"icon" | "text"`
* use a ternary on this prop to decide whether to add the icon button or the text button

src/components/CompareResults/Retrigger/RetriggerButton.tsx
```diff
@@ -156,14 +178,28 @@ function RetriggerButton({ result }: RetriggerButtonProps) {
...
-        color='primary'
-        size='small'
-        onClick={() => void onRetriggerButtonClick()}
-      >
-        <RefreshOutlinedIcon />
-      </IconButton>
+      {variant === 'text' ? (
+        <Button
+          title={Strings.components.revisionRow.title.retriggerJobs}
+          className={styles.retrigger}
```
julienw, line 184: I'd remove all styles, they don't seem useful;

Only the width is problematic, but I have another suggestion (see the main comment)

src/components/CompareResults/Retrigger/RetriggerButton.tsx
```diff
@@ -156,14 +178,28 @@ function RetriggerButton({ result }: RetriggerButtonProps) {
...
+          color='primary'
+          variant={'text'}
+          size='small'
```
julienw, line 187: Once all styles are removed, the small size doesn't look right, so please remove it.

julienw (commented):
Thanks !

I see you clearly tried to workaround how the layout for this row is done.

Currently it uses a flex container, but because the form has `width: 100%` (after #706) it tries to take a lot of space. Instead we should use the MUI's `<Grid>` API like #706 did in the main results page. Would you want to have a look at this in a next PR?

julienw (approved):
thanks!

PR #715, opened 2024-08-09
Results Page: Exit edit mode post updates

Carla-Moz:
This quick PR hides the forms and buttons (exit edit mode) once the user clicks Compare in the Results Page.


julienw (approved):
THanks for the quick fix and the new tests

PR #716, opened 2024-08-09
Compare Over Time: remove reload attribute from the form and update over time tests

Carla-Moz:
Please see title.


julienw (approved):
🤦 I don't test the compare over time enough, thanks for the change

PR #719, opened 2024-08-15
Use position sticky for the headers

julienw:
Problem statement: after the user scrolled down, then the search and filter are out of sight and it can be diffcult to find them back if the page is long. After #712 Full Page Search (Ctrl-F) doesn't work anymore, so it starts to be a bigger problem.

[deploy preview](https://deploy-preview-719--mozilla-perfcompare.netlify.app/compare-results?baseRev=a87b00c4f5cfad5ef2fee563342574340e983619&baseRepo=mozilla-central&newRev=0be9e6fdab7b5eb39e27461cbec046d8a0f9f951&newRepo=mozilla-central&framework=1)

It was much more work than what it looked initially, because the 2 components to make sticky were in very different places and I had to move a lot of code around.

I tried to split it in separate commits. The first commit is #750, I extracted it because it's fairly unrelated to that one. The other commits should be all related to this specific issue but I could split them to separate PRs if you prefer.

Also I didn't do it in the subtests view, mainly because I didn't think about it, but I believe this is still valuable to merge without it.

Tell me what you think!

### DIFF
--- a/src/components/CompareResults/ResultsTable.tsx
+++ b/src/components/CompareResults/ResultsTable.tsx
@@ -145,11 +145,7 @@ export default function ResultsTable() {
     .join(' ');
 
   return (
-    <Box
-      data-testid='results-table'
-      role='table'
-      sx={{ marginTop: 3, paddingBottom: 3 }}
-    >
+    <Box data-testid='results-table' role='table' sx={{ paddingBottom: 3 }}>
       <ResultsControls
         initialSearchTerm={initialSearchTerm}
         frameworkId={frameworkIdVal}
--- a/src/components/CompareResults/ResultsTable.tsx
+++ b/src/components/CompareResults/ResultsTable.tsx
@@ -1,13 +1,18 @@
-import { Suspense, useState, memo } from 'react';
+import { Suspense, useState } from 'react';
 
 import Box from '@mui/material/Box';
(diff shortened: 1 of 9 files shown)


src/components/CompareResults/ResultsTable.tsx
```diff
@@ -124,19 +124,28 @@ export default function ResultsTable() {
...
-        filters={tableFilters}
-        onToggleFilter={onToggleFilter}
-        onClearFilter={onClearFilter}
-      />
+      <Box
+        sx={{
+          position: 'sticky',
+          top: 0,
+          zIndex: 100,
+          bgcolor: 'background.default',
```
julienw, line 132: this adjusts automatically with the dark mode thanks to the palette defined in https://github.com/mozilla/perfcompare/blob/0668d541c997bcedf1a5b43538141d786ac7e32d/src/theme/protocolTheme.ts#L11-L13 and https://github.com/mozilla/perfcompare/blob/0668d541c997bcedf1a5b43538141d786ac7e32d/src/theme/protocolTheme.ts#L29-L31

julienw:
Ah I think I need to do it for the subtests view as well

beatrice-acasandrei (approved):
Lgtm! Thanks!

PR #720, opened 2024-08-15
Make the improvement/regressions eaier to find by adding some colors and icons

canova:
Fixes [Bug 1913290](https://bugzilla.mozilla.org/show_bug.cgi?id=1913290).

This PR adds background colors and svg icons to the "Improvement" and "Regression" labels in the results table.

Before:
Screenshot 2024-08-15 at 3 16 44 PM

After:
Screenshot 2024-08-15 at 3 16 01 PM

Dark mode:

Screenshot 2024-08-15 at 3 30 54 PM

### DIFF
--- a/src/components/CompareResults/RevisionRow.tsx
+++ b/src/components/CompareResults/RevisionRow.tsx
@@ -3,6 +3,8 @@ import { useState, type ReactNode } from 'react';
 import AppleIcon from '@mui/icons-material/Apple';
 import ExpandLessIcon from '@mui/icons-material/ExpandLess';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
+import ThumbDownIcon from '@mui/icons-material/ThumbDown';
+import ThumbUpIcon from '@mui/icons-material/ThumbUp';
 import TimelineIcon from '@mui/icons-material/Timeline';
 import { IconButton } from '@mui/material';
 import Tooltip from '@mui/material/Tooltip';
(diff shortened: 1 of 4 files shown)


src/components/CompareResults/RevisionRow.tsx
```diff
@@ -189,6 +236,12 @@ function determineStatus(improvement: boolean, regression: boolean) {
   return '-';
 }
 
+function determineDeltaHintStyle(improvement: boolean, regression: boolean) {
+  if (improvement) return 'improvement';
+  if (regression) return 'regression';
+  return '';
+}
```
julienw, line 243: nit: Would it be possible to make the classes more specific? I'm thinking of `status-hint-improvement` and `status-hint-regression`

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -189,6 +236,12 @@ function determineStatus(improvement: boolean, regression: boolean) {
   return '-';
 }
 
+function determineDeltaHintStyle(improvement: boolean, regression: boolean) {
```
julienw, line 239: nit:
```suggestion
function determineDeltaHintClass(improvement: boolean, regression: boolean) {
```

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -102,6 +104,28 @@ const stylesLight = {
...
+        backgroundColor: '#D8EEDC',
+      },
+      '.status-hint .MuiSvgIcon-root': {
+        height: '16px',
+      },
+      '.status-hint.improvement .MuiSvgIcon-root': {
+        color: '#017A40',
+      },
+      '.status-hint.regression .MuiSvgIcon-root': {
+        marginTop: '2px',
```
julienw, line 126: can you please add a comment for this marginTop? Eg `// we need to move the icon a bit lower so that it _looks_ center`
(same below)

julienw (approved):
This looks super great!
A few nits and then I think we're good to go!

src/components/CompareResults/RevisionRow.tsx
canova: Ah yeah that looks like a better option. Changed it, thanks!

canova:
Thanks for the reviews!

PR #724, opened 2024-08-18
Persist the search term in the url

canova:
I tried to make it possible to update the URL search parameters without re-rendering the whole page. Unfortunately because of the limitation of react-router, it's not possible on that level. 

I added a custom hook to handle this behavior instead of their hook. I only added it for the search term in this PR, but the other filtering options can be added similarly.

There are a few things to consider though:
- Since the browser navigation still handled by react-router, browser back/forward navigation still re-renders stuff. But that's maybe okay?
- Using `pushState`/`replaceState` while using react-router is not recommended usually by this library. But I'm not worried about this tbh, I think the first item is more important.
- I'm using `pushState` now, but the hook also supports `replaceState`. I'm leaning towards pushState but unsure, would `replaceState` be better for this case?
- There are no tests yet! I can write some if we choose to go with this approach.

Let me know what you think! Happy to get some feedback.

### DIFF
--- a/src/__tests__/CompareResults/SearchInput.test.tsx
+++ b/src/__tests__/CompareResults/SearchInput.test.tsx
@@ -170,4 +170,41 @@ describe('Search by title/test name', () => {
     await user.click(clearButton);
     expect(screen.getByText('spam')).toBeInTheDocument();
   });
+
+  it('should update the url search params properly', async () => {
+    await setupAndRenderComponent();
+
+    const searchInput = await screen.findByRole('textbox', {
+      name: /Search by title/,
+    });
+
+    // set delay to null to prevent test time-out due to useFakeTimers
+    const user = userEvent.setup({ delay: null });
+
+    // With the mock setup in this test, there are 2 results with different
+    // properties. In this test we make each of the input show and disappear in
+    // an alternance, by searching for the values in the different properties.
+    await user.type(searchInput, 'glvideo');
+    await waitFor(() =>
+      expect(screen.queryByText('a11yr')).not.toBeInTheDocument(),
+    );
+    // Make sure that the search params are updated properly.
+    expect(window.location.search).toContain('&search=glvideo');
+
+    await user.clear(searchInput);
+    expect(await screen.findByText('a11yr')).toBeInTheDocument();
+    // Clearing the search should remove the search param.
+    expect(window.location.search).not.toContain('&search');
+
+    await user.type(searchInput, 'a11yr{Enter}');
+    // With enter the update should happen instantly.
+    expect(window.location.search).toContain('&search=a11yr');
+
+    const clearButton = screen.getByRole('button', {
+      name: 'Clear the search input',
+    });
+    await user.click(clearButton);
+    // Clearing the search through button should also remove the search param.
+    expect(window.location.search).not.toContain('&search');
+  });
 });
--- a/src/components/CompareResults/ResultsMain.tsx
+++ b/src/components/CompareResults/ResultsMain.tsx
@@ -11,6 +11,7 @@ import { Await, useLoaderData } from 'react-router-dom';
 import { style } from 'typestyle';
 
 import { useAppSelector } from '../../hooks/app';
+import useRawSearchParams from '../../hooks/useRawSearchParams';
 import { Colors, Spacing } from '../../styles';
 import type { CompareResultsItem } from '../../types/state';
 import { Framework } from '../../types/types';
@@ -29,8 +30,11 @@ function ResultsMain() {
 
   const themeMode = useAppSelector((state) => state.theme.mode);
   const [searchParams, setSearchParams] = useSearchParams();
-  const [searchTerm, setSearchTerm] = useState('');
 
+  // This is our custom hook that updates the search params without a rerender.
+  const [rawSearchParams, updateRawSearchParams] = useRawSearchParams();
+  const initialSearchTerm = rawSearchParams.get('search') ?? '';
+  const [searchTerm, setSearchTerm] = useState(initialSearchTerm);
   const [frameworkIdVal, setFrameworkIdVal] = useState(frameworkId);
 
   const themeColor100 =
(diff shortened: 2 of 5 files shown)


src/hooks/useUpdateSearchParams.ts
```diff
@@ -0,0 +1,59 @@
...
+ * search paramters _without_ creating a rerender on the router level.
+ *
+ * NOTE: It's important to handle the search param you're adding in the react
+ * router loader to make sure that loading a URL from browser or refreshing the
+ * page in browser handles it properly.
+ */
+const useUpdateSearchParams = (): UpdateSearchParams => {
+  // useSearchParams hook provides the current search parameters
+  // and a method to update them, which would trigger a re-render.
+  const [searchParams] = useSearchParams();
```
julienw, line 24: Frankly I'd bail out of everything related to react router in here. So I'd get the searchParams from `location.search` directly, and I'd also expose them like the original react-router hook.

This would have the nice following advantages:
* it could be used anywhere and not just at the top level of a react component
* it wouldn't subscribe to the react router context -- I think that when using `useSearchParams` we'd get a rerender if _something somewhere_ changes the location.
* A developer could still use both hooks (this one and the one from react router) together to be able to get this behavior is they want to.
* by returning the `searchParams` from this hook, I believe we don't need to change the loader because the render function could pick it up when rendering. Maybe this is true only for `replaceState` though? I'm not sure.

Then I'd call the hook `useRawSearchParams` (maybe)?

src/hooks/useUpdateSearchParams.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  // This function updates the search params without causing a re-render
+  // and provides flexibility to use either pushState or replaceState.
+  const updateSearchParams: UpdateSearchParams = (
+    key,
+    value,
```
julienw, line 30: In terms of API I'd take an object so that we can set several params in one call.

src/hooks/useUpdateSearchParams.ts
```diff
@@ -0,0 +1,59 @@
...
+    options = { method: 'push' },
+  ) => {
+    // Create a copy of the current search parameters.
+    const newSearchParams = new URLSearchParams(searchParams.toString());
+
+    if (value) {
```
julienw, line 36: maybe check for null and undefined? So that `0` and `""` work

src/components/CompareResults/SearchInput.tsx
```diff
@@ -8,10 +8,11 @@ import { Strings } from '../../resources/Strings';
 import { simpleDebounce } from '../../utils/simple-debounce';
 
 interface SearchInputProps {
+  initialValue?: string;
```
julienw, line 11: nit: it's usually called `defaultValue`

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -61,13 +67,21 @@ function ResultsMain() {
     setSearchParams(searchParams);
   };
 
+  const onSearchTermChange = (newSearchTerm: string) => {
+    setSearchTerm(newSearchTerm);
+    updateSearchParams('search', newSearchTerm);
```
julienw, line 72: Yeah, I'd use a "replaceState".

My rationale is that when the user presses the back button, IMO they expect to go to the previous page, not the previous search. They don't especially have the mental model that changing the search will push a new history item, because that's not what happens in most sites I know.

The profiler is different because it's more like an "app" and we expect to be able to undo things. Although in the profiler we could possibly use replaceState for that too for the same reason, because searching is not really an action.

julienw (commented):
thanks for trying this!
here are my 2 cents, happy to discuss this more later

canova (commented):
Thanks for the review. I changed the code and indeed changing the API made it a lot better and allowed me to remove the loader changes. It looks cleaner now. Added a test for the search term behavior as well.

canova:
I think it's ready for review now. Removed the WIP tags.

src/hooks/useRawSearchParams.ts
```diff
@@ -0,0 +1,44 @@
...
+
+type RawSearchParams = [URLSearchParams, UpdateSearchParams];
+
+/**
+ * This hook provides an `updateSearchParams` function that can update the
+ * search parameters _without_ creating a re-render on the router level.
+ */
+const useRawSearchParams = (): RawSearchParams => {
+  // Memoize the current search parameters to avoid unnecessary recalculations
+  const searchParams = useMemo(() => {
```
canova, line 20: I used `useMemo` here but probably it's a bit of an overkill... It should be quite fast to construct a new `URLSearchParams`. Let me know if you prefer me to remove it instead.

src/hooks/useRawSearchParams.ts
```diff
@@ -0,0 +1,44 @@
...
+    return new URLSearchParams(window.location.search);
+  }, [window.location.search]);
+
+  // This function updates the search params without causing a re-render
+  // and provides flexibility to use either pushState or replaceState.
+  const updateSearchParams: UpdateSearchParams = (
+    newSearchParams,
+    options = { method: 'replace' },
+  ) => {
```
julienw, line 29: In terms of API, what about having updateSearchParams take a record instead of a SearchParams? The record wouldn't be the full search param but how we'd like to change it. That would move the complexity from the user to the hook.

* if a property has a primary value => use "set"
* if a property has an array value => use "set" for the first one, "append" for the other ones
* if a property has a null or undefined value => use "delete"

This would be a departure from what react router's `useSearchParams` does (because that one always takes the full version, either a Record, a Array, a URLSearchParams or a string, see [type declaration in their doc](https://reactrouter.com/en/main/hooks/use-search-params)), but I think this fits better our use cases where we want to change only one part, at the price of some flexibility.

What do you think?

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -29,8 +30,11 @@ function ResultsMain() {
 
   const themeMode = useAppSelector((state) => state.theme.mode);
   const [searchParams, setSearchParams] = useSearchParams();
-  const [searchTerm, setSearchTerm] = useState('');
 
+  // This is our custom hook that updates the search params without a rerender.
+  const [rawSearchParams, updateSearchParams] = useRawSearchParams();
+  const initialSearchTerm = rawSearchParams.get('search') ?? '';
+  const [searchTerm, setSearchTerm] = useState(initialSearchTerm);
```
julienw, line 37: can you also do the change in SubtestsResultsMain too?

canova: Ah right I forgot about that page. Fixed it now.

src/hooks/useRawSearchParams.ts
canova: We've discussed it in Matrix a bit, and decided that pros and cons are not so clear for one over another. So let's continue with this approach for now and then we can revisit later if we really want.

src/__tests__/CompareResults/SearchInput.test.tsx
```diff
@@ -170,4 +170,41 @@ describe('Search by title/test name', () => {
     await user.click(clearButton);
     expect(screen.getByText('spam')).toBeInTheDocument();
   });
+
+  it('should update the url search params properly', async () => {
```
julienw, line 174: Thanks for the new test!

julienw (approved):
Thanks, this looks perfect!

PR #725, opened 2024-08-19
Fix SelectList item active and selector background colors on dark mode

canova:
This fixes [Bug 1913699](https://bugzilla.mozilla.org/show_bug.cgi?id=1913699).

Please see the STR that I mentioned in the bug. `SelectList` selected item active and selected background colors on dark mode were still using the light mode's color and that's why they were hard to read on a white text. This PR fixes those colors so it's easier to read them, and the UI colors are more consistent on dark mode.

Before:
Screenshot 2024-08-19 at 11 35 54 AM

After:
Screenshot 2024-08-19 at 11 46 20 AM

### DIFF
--- a/src/styles/SelectList.ts
+++ b/src/styles/SelectList.ts
@@ -100,13 +100,13 @@ export const SelectListRaw = {
             borderRadius: '4px',
           },
           '&:active': {
-            backgroundColor: Colors.SecondaryActive,
+            backgroundColor: Colors.SecondaryActiveDark,
             borderRadius: '4px',
           },
         },
       },
       '.item-selected': {
-        backgroundColor: Colors.SecondaryHover,
+        backgroundColor: Colors.SecondaryHoverDark,
         borderRadius: '4px',
       },
       '.revision-hash': {


julienw (approved):
thanks!
I'm not a big fan of how all these styles are done (I think we should try to use more of MUI's builtin themeing capabilities), but this fixes the issue.

Note that I see a similar issue exists also in the framework dropdown, but it's more complex to fix. It's due to these `  cssRule('.MuiPopover-root', ...` rules that are in a lot of places, that I'd like to remove and do in another way. That's for another PR though, clearly out of scope here.

Thanks!

canova:
Thanks for the review!

> Note that I see a similar issue exists also in the framework dropdown, but it's more complex to fix. It's due to these ` cssRule('.MuiPopover-root', ...` rules that are in a lot of places, that I'd like to remove and do in another way. That's for another PR though, clearly out of scope here.

Yes, I actually saw the same thing and tried to fix it, but it quickly became harder than expected because of these rules, so I stopped doing it 😄

PR #727, opened 2024-08-21
Simplify the state changing of the block expansions in the home page

julienw:
Small simplification in how the home page blocks are expanded. The goal is that the state for the "CompareWithBase" block doesn't leak into "CompareOverTime" block, so instead of having this spilled state, a simpler boolean "isExpanded" is passed to both blocks. This also simplifies how they work in the results page.

There should be no behavior change.

Tell me what you think!


julienw:
Note: I accidentally pushed other commits to this PR earlier, but I fixed that now.

beatrice-acasandrei (approved):

PR #728, opened 2024-08-21
Add links to the perfcompare app linking to the old compareview, keeping the same revisions

beatrice-acasandrei:


julienw:
I'd put that in the top banner, but I don't know if that's easy with our current architecture?

src/components/CompareResults/loader.ts
```diff
@@ -249,6 +249,7 @@ type DeferredLoaderData = {
   newRepos: Repository['name'][];
   frameworkId: Framework['id'];
   frameworkName: Framework['name'];
+  intervalValue: TimeRange['value'];
```
julienw, line 252: I really don't like that we're lying here: that we don't return it ever.

I understand why you're doing it but I think it may not be needed if you write the code like this:

```js
const loaderData = useLoaderData() as XXX
if (loaderData.view === overTimeView) {
  const { newRev, intervalValue, ... } = loaderData; // Here typescript should know it's correct 
}
if (loaderData.view === compareView) {
  const { baseRev, ...} = loaderData; // Same
}
```

Can you test this and see if this works?

If that doesn't work you can also add `as XXX` after `loaderData` to force the type, but I _think_ this isn't needed.

I just noticed you did something similar in the subtests file.

_otherwise_, if that's really needed, we can use the same shape in all return values (that is the same properties) but don't lie about their types. Here for example `intervalValue` is always `undefined` (that is `void` in typescript) so we should type it this way.

julienw: This comment applies to the changes to the other loader as well.

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -23,10 +49,40 @@ function ResultsMain() {
...
+  console.log(newRevs);
   return (
     <Container className={styles.container} data-testid='results-main'>
       <header>
-        <div className={styles.title}>Results</div>
+        <div className={styles.title}>Results </div>
+        <Alert severity='info' className={styles.title}>
+          Perfherder links are available for:{' '}
+          {newRevs.map((rev, index) => (
+            <Fragment key={truncateHash(rev)}>
```
julienw, line 78: nit: just use `rev` directly for the key

julienw (changes requested):
Thanks, I really like how this looks and the behavior!
However I think there are a few things to change in how it's implemented, I tried to outline them in the comments below. Hopefully that won't be too much. Tell me what you think!

src/__tests__/CompareResults/RevisionSelect.test.tsx
```diff
@@ -109,18 +123,32 @@ describe('Revision select', () => {
...
+      selector: 'a',
+    }),
+  ).toStrictEqual([]);
+
+  expect(
+    screen.queryAllByText(/a998c42399a8/, {
+      selector: 'a',
+    }),
+  ).toStrictEqual([]);
+  */
```
julienw, line 153: nit: remove these comments

julienw (approved):

PR #730, opened 2024-08-22
Populate the search form in the home page from the URL

julienw:
The goal for this PR is to provide the functionality for the button "Compare against another revision" in treeherder ([example 1](https://treeherder.mozilla.org/jobs?repo=mozilla-central&searchStr=talos&selectedTaskRun=YPtef1jVQ8KcmqPDYw-PFQ.0) / [example 2](https://treeherder.mozilla.org/jobs?repo=mozilla-central&searchStr=browsertime&selectedTaskRun=MC0qnSp7SvyX6e7VI9JhEw.0))

I've been one step further as the original implementation to be able to provide the framework as well. I also decided that it makes sense to prefill the base repository with the same repository too.

I went with using a loader like in other views because we had to fetch information to the server about this revision. Because it's a one-time thing this wasn't strictly necessary and I could have do the fetch from the react component, but then I would have to deal with promise resolution and all sorts of things, and I thought this was simpler this way.

Please look at the 4 commits separately, I believe this will be simpler to review:
* Commit 1: add a loader
* Commit 2: update the component
* Commit 3: update the existing tests so that all tests pass
* Commit 4: add new tests covering the new behavior.

For the tab button, the related code in treeherder is https://github.com/mozilla/treeherder/blob/6df095030feeccfc93536050926f2e5b3627000a/ui/job-view/details/tabs/PerformanceTab.jsx#L238-L249. We can access the framework information from `perfJobDetails[0].perfDocs.framework` (if I remember properly) where we have the framework name, not the id (that's why the loader uses a frameworkName parameter -- actually I feel like we should have used the frameworkName in our URL everywhere too, but that ship has sailed :-) ).

For the job menu, the code is in https://github.com/mozilla/treeherder/blob/6df095030feeccfc93536050926f2e5b3627000a/ui/job-view/pushes/PushActionMenu.jsx#L183.
Here I'm not sure we have access to the framework information, but I guess that's OK we can ignore it here.

[Without any parameters (like before)](https://deploy-preview-730--mozilla-perfcompare.netlify.app/)
[With only the revision/repo parameters](https://deploy-preview-730--mozilla-perfcompare.netlify.app/?newRev=fefd575ace5522580a170ee043e58eab7a93a74e&newRepo=mozilla-central)
[With the revision/repo parameters as well as the framework parameter](https://deploy-preview-730--mozilla-perfcompare.netlify.app/?newRev=fefd575ace5522580a170ee043e58eab7a93a74e&newRepo=mozilla-central&frameworkName=browsertime)
[With bogus value for revision](https://deploy-preview-730--mozilla-perfcompare.netlify.app/?newRev=abcdef&newRepo=mozilla-central)
[With bogus value for repository](https://deploy-preview-730--mozilla-perfcompare.netlify.app/?newRev=fefd575ace5522580a170ee043e58eab7a93a74e&newRepo=foo)
[With bogus value for framework](https://deploy-preview-730--mozilla-perfcompare.netlify.app/?newRev=fefd575ace5522580a170ee043e58eab7a93a74e&newRepo=mozilla-central&frameworkName=foo)

### DIFF
--- a/src/__tests__/CompareResults/__snapshots__/ResultsView.test.tsx.snap
+++ b/src/__tests__/CompareResults/__snapshots__/ResultsView.test.tsx.snap
@@ -20,134 +20,131 @@ exports[`Results View Should display Base, New and Common graphs with 1 value an
          
         <br />
       </div>
-      <div>
+      <div
+        class="fc8wfj0"
+      >
+        <hr
+          class="MuiDivider-root MuiDivider-fullWidth css-9mgopn-MuiDivider-root"
+        />
+      </div>
+      <div
+        class="f9mtzuo"
+      >
         <div
-          class="fc8wfj0"
+          class="fnblmop"
         >
-          <hr
-            class="MuiDivider-root MuiDivider-fullWidth css-9mgopn-MuiDivider-root"
-          />
-           
+          <div>
+            <b>
+              Base
+              :
+            </b>
+             
+            596.87
+             
+             (
+            firefox
+            )
+          </div>
+          <div>
+            chartjs-bubble
+          </div>
+          <div>
+            <div
+              class="f1s9htso"
+            >
+              <div
+                class="f1pfon3k"
+              >
+                587.15
+              </div>
+              <div
+                class="f1pfon3k"
+              >
+                593.04
+              </div>
+              <div
+                class="f1pfon3k"
+              >
+                600.7
+              </div>
+              <div
+                class="f1pfon3k"
+              >
+                602.04
+              </div>
+            </div>
+            <div
+              class="fnujpat"
+            >
+              6.96
+               = 
+              1.17
+              % standard deviation
+            </div>
+          </div>
         </div>
         <div
-          class="f9mtzuo"
+          class="fnblmop"
         >
-          <div
-            class="fnblmop"
-          >
-            <div>
-              <b>
-                Base
-                :
-              </b>
-               
-              596.87
-               
-               (
-              firefox
-              )
-            </div>
-            <div>
-              chartjs-bubble
-            </div>
-            <div>
+          <div>
+            <b>
+              New
+              :
+            </b>
+             
+            605.61
+             
+             (
+            chrome
+            )
+          </div>
+          <div>
+            chartjs-bubble
+          </div>
+          <div>
+            <div
+              class="f1s9htso"
+            >
               <div
-                class="f1s9htso"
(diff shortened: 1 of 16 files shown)


beatrice-acasandrei:
Looks good! Are we planing to improve this feature by allowing URL parameters for multiple new revisions / base revision / time range or is this the final work? In my opinion we might want to allow multiple new revisions since it's one of the new features.

julienw:
> Looks good! Are we planing to improve this feature by allowing URL parameters for multiple new revisions / base revision / time range or is this the final work? In my opinion we might want to allow multiple new revisions since it's one of the new features.

Good question!
Here I focused on making it work for the use case of the button in the performance tab.
I think supporting that we can input several revisions could be possible, and we could also probably share some code with the compare results loader in this process. But should we do that now? What would the end to end scenario look like?

Carla-Moz:
> > Looks good! Are we planing to improve this feature by allowing URL parameters for multiple new revisions / base revision / time range or is this the final work? In my opinion we might want to allow multiple new revisions since it's one of the new features.
> 
> Good question! Here I focused on making it work for the use case of the button in the performance tab. I think supporting that we can input several revisions could be possible, and we could also probably share some code with the compare results loader in this process. But should we do that now? What would the end to end scenario look like?

@julienw @beatrice-acasandrei 

Wouldn't accommodating multiple new revisions mean this feature would also be explicit in the related treeherder code's performance tabs? Because right now, in treeherder, we can only compare with "one" revision with the `Compare against another revision` or `Compare Performance` buttons, right? Just trying to understand better. 

And if that's the case, I think for now we can keep the links with one revision, repo, and framework because users can pick the base, add multiple revisions, or pick the timeRange once they reach PerfCompare, right? If we want to add multiple revisions or timeRange in the home page url etc, we'd have to change this behavior in the treeherder Performance tabs which I think is beyond the scope of this work; users can already do so in PerfCompare. 

Now, another scenario would be if a user knew which base, new revisions, framework,or timeRange they wanted in the home page and could populate the url that way. This is different from the treeherder tab Performance work, right? I think this would merit a new ticket altogether since this is a separate work flow. So I think this PR is done until we want to do another one for a new workflow. 

What do you think? Let me know if this needs clarifying or if I've misunderstood something.

Carla-Moz (approved):

julienw:
@Carla-Moz Fully agree with your comment :-)

thanks for the review!

PR #731, opened 2024-08-22
Update subtests route and modify search param to selectedTimeRange

beatrice-acasandrei:


julienw (approved):
looks good to me, thanks!

(for another PR) I see in codecov that we're missing some test coverage for the subtests for compare over time, it would be good to add a few tests to at least get some coverage there.

PR #732, opened 2024-08-26
Transition subtest header to grid layout

esanuandra:
Follow up for [adding retrigger functionality for subtests](https://github.com/mozilla/perfcompare/pull/711#pullrequestreview-2248064981)
Screenshot 2024-08-26 at 10 21 28 AM

### DIFF
--- a/src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
+++ b/src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
@@ -54,20 +54,6 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
       margin: 0,
       marginBottom: Spacing.Medium,
     }),
-    content: style({
-      display: 'flex',
-      justifyContent: 'space-between',
-      alignItems: 'center',
-    }),
-    buttons: style({
-      display: 'flex',
-      justifyContent: 'flex-end',
-      $nest: {
-        '.MuiButtonBase-root:nth-child(2)': {
-          marginLeft: Spacing.Default,
-        },
-      },
-    }),
   };
 
   return (
@@ -75,12 +61,14 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
       <header>
         <SubtestsBreadcrumbs view={view} />
         <SubtestsRevisionHeader header={subtestsHeader} />
-        <Grid container className={styles.content}>
-          <Grid item md={6} xs={12}>
+        <Grid container spacing={1}>
+          <Grid item xs={12} md={6} sx={{ marginInlineEnd: 'auto' }}>
             <SearchInput onChange={setSearchTerm} />
           </Grid>
-          <Grid item className={styles.buttons} md={4}>
+          <Grid item xs='auto'>
             <DownloadButton resultsPromise={[results]} />
+          </Grid>
+          <Grid item xs='auto'>
             <RetriggerButton result={results[0]} variant='text' />
           </Grid>
         </Grid>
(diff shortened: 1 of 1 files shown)


src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
```diff
@@ -55,21 +56,34 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
...
+      justifyContent: 'space-between',
       alignItems: 'center',
     }),
+    buttons: style({
+      display: 'flex',
+      justifyContent: 'flex-end',
+      $nest: {
+        '.MuiButtonBase-root:nth-child(2)': {
+          marginLeft: Spacing.Default,
+        },
```
julienw, line 68: It's easier to use `gap: Spacing.Default` directly on the flex container for this.


(but you don't need it with the suggestion above)

julienw (commented):
Thanks!
By combining the `Grid` API with some custom CSS,  I believe this will leads to future complications.

Here is a diff that seems to work well:
```diff
diff --git a/src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx b/src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
index 138c8288..ed4efe5e 100644
--- a/src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
+++ b/src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
@@ -49,43 +49,31 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
       backgroundColor: themeColor100,
       margin: '0 auto',
       marginBottom: '80px',
     }),
     title: style({
       margin: 0,
       marginBottom: Spacing.Medium,
     }),
-    content: style({
-      display: 'flex',
-      justifyContent: 'space-between',
-      alignItems: 'center',
-    }),
-    buttons: style({
-      display: 'flex',
-      justifyContent: 'flex-end',
-      $nest: {
-        '.MuiButtonBase-root:nth-child(2)': {
-          marginLeft: Spacing.Default,
-        },
-      },
-    }),
   };
 
   return (
     <Container className={styles.container} data-testid='subtests-main'>
       <header>
         <SubtestsBreadcrumbs view={view} />
         <SubtestsRevisionHeader header={subtestsHeader} />
-        <Grid container className={styles.content}>
-          <Grid item md={6} xs={12}>
+        <Grid container spacing={1}>
+          <Grid item xs={12} md={6} marginInlineEnd='auto'>
             <SearchInput onChange={setSearchTerm} />
           </Grid>
-          <Grid item className={styles.buttons} md={4}>
+          <Grid item xs='auto'>
             <DownloadButton resultsPromise={[results]} />
+          </Grid>
+          <Grid item xs='auto'>
             <RetriggerButton result={results[0]} variant='text' />
           </Grid>
         </Grid>
       </header>
       <SubtestsResultsTable
         filteringSearchTerm={searchTerm}
         results={results}
       />
```

The main difference is that `DownloadButton` and `RetriggerButton` are each wrapped by a `Grid item`. Using `xs=auto` makes them sized according to their natural width. Also I'm using `spacing` directly on the `Grid container` to define the gap between elements (reminder that the value in `spacing` is multiplied by the spacing configutation in the theme, by default 8).
`marginInlineEnd=auto` (equivalent to marginRight but RTL friendly) on the search input pushes all the other elements towards the end, like your `flex-end`.

Tell me what you think!

julienw:
Ah it looks like passing `marginInlineEnd` as a prop doesn't work with our current types, then just use `sx`:
```
 sx={{ marginInlineEnd: 'auto' }}
```

julienw (approved):
thanks!

I'm surprised we don't have any test to change, are we missing some coverage here? Can you have a look at it?

esanuandra:
> thanks!
> 
> I'm surprised we don't have any test to change, are we missing some coverage here? Can you have a look at it?

`SubtestsResultsMain` is fully covered

julienw:
> @julienw `SubtestsResultsMain` is fully covered

I think we don't have a targeted test for this, we only "accidentally" cover it from the retrigger tests.

PR #736, opened 2024-08-27
PCF-497 The warning icon showing up when comparing try vs mozilla-central sho…

beatrice-acasandrei:
…uld give more information about _why_ it's not recommended


Carla-Moz (approved):
LGTM!

PR #737, opened 2024-08-28
PCF-413 Replace the Base search results drop-down check boxes with radio buttons.

beatrice-acasandrei:


src/__tests__/Search/SearchView.test.tsx
```diff
@@ -318,7 +318,7 @@ describe('Base and OverTime Search', () => {
       screen.getAllByText("it's just a flesh wound")[0],
     ).toBeInTheDocument();
     await user.click(screen.getAllByText("you've got no arms left!")[0]);
-    await user.click(screen.getAllByTestId('CheckBoxOutlineBlankIcon')[0]);
+    await user.click(screen.getAllByTestId('CheckBoxOutlineBlankIcon')[0]); //?
```
julienw, line 321: Maybe we can do a better assertion here indeed...

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -82,14 +87,26 @@ function SearchResultsListItem({
...
           <ListItemIcon className='search-revision-item-icon search-revision'>
-            <Checkbox
-              className='search-revision-item-checkbox'
-              edge='start'
-              tabIndex={-1}
-              disableRipple
-              data-testid={`checkbox-${index}`}
-              checked={isChecked}
-            />
+            {searchType === 'base' ? (
```
julienw, line 90: noooo please don't add another searchType check :-)
By doing this, we leak the "outside context" to the inside component.
Ideally we should have this type of check nowhere in our app.

Instead, we can configure the component to make it more explicit and straightforward.

```ts
interface {
  ...
  listItemComponent?: 'checkbox' | 'radio'
}

// later

// Defaults to checkbox
const ListItemComponent = listItemComponent === 'radio' ? Radio : Checkbox;

...

<ListItemComponent
  ...
/>
```

Then the information needs to come from `CompareWithBase` and be propagated all the way down to this component.

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -82,14 +87,26 @@ function SearchResultsListItem({
...
+                data-testid={`radio-${index}`}
+                checked={isChecked}
+              />
+            ) : (
+              <Checkbox
+                className='search-revision-item-checkbox'
+                edge='start'
+                tabIndex={-1}
+                disableRipple
+                data-testid={`checkbox-${index}`}
```
julienw, line 106: ideally we shouldn't use these test ids...
TBH I'm fine if you want to keep "checkbox" for all of them, because that's test only...

julienw (approved):
lgtm, thanks!

PR #738, opened 2024-08-28
PCF-500 Results search in the application should be case insensitive

beatrice-acasandrei:


julienw (approved):
looks good to me, but if you can add a test that checks the case insensitive behavior for at least some of the properties, that would be great

PR #739, opened 2024-08-28
PCF-504 Use units where needed in the expanded row

esanuandra:
Added measurement unit for standard deviation and mean difference.

Screenshot 2024-08-28 at 4 38 50 PM

### DIFF
--- a/src/components/CompareResults/RevisionRowExpandable.tsx
+++ b/src/components/CompareResults/RevisionRowExpandable.tsx
@@ -24,6 +24,9 @@ function RevisionRowExpandable(props: RevisionRowExpandableProps) {
     new_app: newApplication,
   } = result;
 
+  const unit = baseUnit || newUnit;
+  const deltaUnit = unit ? `${unit}` : '';
+
   const themeMode = useAppSelector((state) => state.theme.mode);
   const themeColor200 =
     themeMode == 'light' ? Colors.Background200 : Colors.Background200Dark;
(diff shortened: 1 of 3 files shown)


src/components/CompareResults/RevisionRowExpandable.tsx
```diff
@@ -76,7 +78,7 @@ function RevisionRowExpandable(props: RevisionRowExpandableProps) {
         </div>
         <div className={`${styles.bottomSpace}`}>
           <b>Mean Difference</b>: {deltaPercent}%{' '}
-          {newIsBetter ? 'better' : 'worse'} ({delta})
+          {newIsBetter ? 'better' : 'worse'} ({delta} {baseUnit || newUnit})
```
julienw, line 81: When there's no unit, this gives an extra space, I believe it should'nt be there in that case.
I think you can do something like this:
```
const unit = baseUnit || newUnit;
const deltaUnit = unit ? ` ${unit}` : '';

// Later
{newIsBetter ? 'better' : 'worse'} ({delta}{deltaUnit})

File: src/components/CompareResults/RunValues.tsx [hunk 2]
```diff
@@ -61,7 +61,7 @@ function RunValues(props: RunValuesProps) {
           ))}
         </div>
         <div className={styles.deviation}>
-          {stddev} = {stddevPercent}% standard deviation
+          {stddev} {measurementUnit} = {stddevPercent}% standard deviation
```
julienw, line 64: same thing here

julienw: I find this line being in capital letters very weird
as well as its general structure actually...  It should probably be displayed next to "mean difference" with the same design. I guess it's another jira ticket and we can ask Dasha... can you file one?

julienw (approved):
Looks good to me with these small adjustments

PR #743, opened 2024-08-30
Fixed the fetch for compare over time subtests

beatrice-acasandrei:


julienw (approved):
Thanks for the quick fix!

We're really missing coverage on this code ;-)

PR #752, opened 2024-09-11
PCF-426 Improve the "Results" title to be more visible and contain the context (eg: "Compare with base")

esanuandra:


src/components/CompareResults/ResultsMain.tsx
```diff
@@ -20,13 +27,46 @@ function ResultsMain() {
...
   return (
     <Container className={styles.container} data-testid='results-main'>
       <header>
-        <div className={styles.title}>Results</div>
+        <Grid container className={styles.titleContainer}>
+          <Grid item className={styles.title}>
+            Results
+          </Grid>
+          <Grid item className={styles.subtitle}>
+            {subtitles[view]}
```
Carla-Moz, line 67: How about enclosing the title and subtitle in `h `tags?

Carla-Moz (approved):
LGTM!

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -20,13 +27,46 @@ function ResultsMain() {
     title: style({
       margin: 0,
       marginBottom: Spacing.Medium,
```
julienw, line 29: move this margin to titleContainer

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -20,13 +27,46 @@ function ResultsMain() {
...
+      fontStyle: 'Heading XS',
+      fontFamily: 'Metropolis',
+      fontSize: '24px',
+      fontWeight: 700,
+      lineHeight: '28px',
+      letterSpacing: '-0.01em',
+      textAlign: 'left',
```
julienw, line 36: you don't need the text-align

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -20,13 +27,46 @@ function ResultsMain() {
     title: style({
       margin: 0,
       marginBottom: Spacing.Medium,
+      fontStyle: 'Heading XS',
```
julienw, line 30: you can remove fontStyle, this isn't a valid value

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -20,13 +27,46 @@ function ResultsMain() {
...
+      fontFamily: 'Metropolis',
+      fontSize: '24px',
+      fontWeight: 700,
+      lineHeight: '28px',
```
julienw, line 34: please use FontsRaw.HeadingXS from `src/styles/Fonts.ts` when it makes sense
and of course then add the ones that are different.

(in the future we'll want to use MUI's typography but I think this would be more work atm)

src/components/CompareResults/ResultsMain.tsx
esanuandra: What h tags were you thinking? Can you give me an example?

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -11,22 +15,51 @@ function ResultsMain() {
...
+      ...FontsRaw.BodyDefault,
+      fontSize: '15px',
+      lineHeight: '20px',
+      borderLeft: '1px solid #5B5B66',
+      paddingLeft: '9px',
+    }),
+    titleContainer: style({
+      alignItems: 'center',
+      gap: '9px',
+      margin: `0 0 ${Spacing.Medium} 0`,
```
julienw, line 43: nit: the `margin` doesn't work, you need to add the `px` after spacing.medium.

julienw (approved):
Let's land this :-) Thanks!

PR #758, opened 2024-09-20
PCF-502: increased height of results list to show first five items

Carla-Moz:
Moved specific results list styles to `SearchResultsList` because they were only used once in `styles/SelectList`
Fix for issue[ PCF-502](https://mozilla-hub.atlassian.net/browse/PCF-502). Let me know if the height is satisfactory or if I should increase even more.

### DIFF
--- a/src/__tests__/App.test.tsx
+++ b/src/__tests__/App.test.tsx
@@ -58,7 +58,9 @@ describe('App', () => {
 
     render(<App />);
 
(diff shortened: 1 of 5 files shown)


julienw (approved):
thanks, this looks good for me

I'd still want to get rid of these nested CSS rules that are a bit far from where they're actually used (eg info-caption or revision-hash), but we can do this cleanup later.

PR #759, opened 2024-09-23
PCF-427 Redesign the search form to make it more compact

esanuandra:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/PCF-427)
This PR removes the title header for the search form in Results page.

Screenshot 2024-10-01 at 5 19 23 PM
Screenshot 2024-10-01 at 5 19 41 PM


julienw (commented):
This looks pretty good to me, here are a few nits but otherwise I think it's good to go fixing the tests.

src/components/Search/SearchFormHeader.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const mode = useAppSelector((state) => state.theme.mode);
+  const styles = CompareCardsStyles(mode);
+
+  return (
+    <div
+      className={`compare-card-container compare-card-container--${expandedClass} ${styles.container}`}
+      onClick={onClick}
+      data-testid={`${compareType}-state`}
+    >
+      <div className={`compare-card-text ${styles.cardText}`}>
```
julienw, line 28: Similarly to the Grid above, I think we need a `<h2>` here as well

julienw: as discussed, the h2 is already present below

julienw (approved):
as we discussed it would be good to follow-up with reducing the margins (even though some of the margin is due to some invisible elements) :)
Thanks!

PR #760, opened 2024-09-23
Increase the maximum amount of possible retriggers to 9

julienw:
As requested by @acreskeyMoz!
image

I tried using a native select so that you wouldn't have to open the select before selecting the value with the keyboard but I couldn't style it quickly properly, so I kept MUI's version for now.


acreskeyMoz:
Thank you!

Carla-Moz (approved):

PR #763, opened 2024-10-02
Add date range to time range dropdown

sumairq:
Solves [Bug 1922027](https://bugzilla.mozilla.org/show_bug.cgi?id=1922027)
## Description
  This PR enhances the existing time range dropdown in the CompareOverTime UI by displaying specific dates alongside the predefined time range options. This provides users with better clarity on the exact date ranges they are selecting.

## Changes:
Updated the time range dropdown to display specific start and end dates next to predefined options (e.g., "Last 7 days" now shows the exact dates).

## Screenshot:
image

### DIFF
--- a/src/components/Search/TimeRangeDropdown.tsx
+++ b/src/components/Search/TimeRangeDropdown.tsx
@@ -56,11 +56,9 @@ function TimeRangeDropdown({
       month: '2-digit',
       day: '2-digit',
     };
-    const fromDate = new Date(Date.now() - value * 1000).toLocaleDateString(
-      'en-US',
-      options,
-    );
-    const toDate = new Date(Date.now()).toLocaleDateString('en-US', options);
+    const dateFormatter = new Intl.DateTimeFormat('en-US', options);
+    const fromDate = dateFormatter.format(new Date(Date.now() - value * 1000));
+    const toDate = dateFormatter.format(new Date());
     return ` (${fromDate} - ${toDate})`;
   };
 
--- a/src/components/Search/TimeRangeDropdown.tsx
+++ b/src/components/Search/TimeRangeDropdown.tsx
@@ -1,6 +1,7 @@
 import FormControl from '@mui/material/FormControl';
 import MenuItem from '@mui/material/MenuItem';
 import Select, { SelectChangeEvent } from '@mui/material/Select';
+import { Box } from '@mui/system';
 import { style } from 'typestyle';
 
 import { timeRangeMap } from '../../common/constants';
@@ -10,7 +11,6 @@ import {
   ButtonsLightRaw,
   ButtonsDarkRaw,
   DropDownItems,
-  DateRange,
   FontSize,
 } from '../../styles';
 import type { TimeRange } from '../../types/types';
@@ -40,6 +40,13 @@ function TimeRangeDropdown({
         },
       },
     }),
+    menuItem: style({
+      display: 'block',
+    }),
+    dateRange: style({
+      display: 'flex',
+      justifyContent: 'space-between',
+    }),
   };
 
   const handleTimeRangeSelect = async (event: SelectChangeEvent) => {
@@ -48,7 +55,6 @@ function TimeRangeDropdown({
   };
   const menuItemStyles =
     mode === 'light' ? DropDownItems.Light : DropDownItems.Dark;
-  const dateRangeStyles = mode === 'light' ? DateRange.Light : DateRange.Dark;
 
   const displayDateRange = (value: number) => {
     const options: Intl.DateTimeFormatOptions = {
@@ -82,12 +88,17 @@ function TimeRangeDropdown({
             <MenuItem
               value={value}
               key={value}
-              className={`timerange-dropdown-item ${menuItemStyles}`}
+              className={`timerange-dropdown-item ${menuItemStyles} ${styles.menuItem}`}
             >
-              {text}
-              <span className={`${FontSize.Small} ${dateRangeStyles}`}>
-                {displayDateRange(Number(value))}
-              </span>
+              <div className={styles.dateRange}>
+                {text}
+                <Box
+                  sx={{ color: 'text.secondary' }}
+                  className={`${FontSize.Small}`}
+                >
+                  {displayDateRange(Number(value))}
+                </Box>
+              </div>
             </MenuItem>
           ))}
         </Select>
--- a/src/components/Search/TimeRangeDropdown.tsx
+++ b/src/components/Search/TimeRangeDropdown.tsx
@@ -14,6 +14,7 @@ import {
   FontSize,
 } from '../../styles';
 import type { TimeRange } from '../../types/types';
+import { formatDateRange } from '../../utils/format';
 
 const strings = Strings.components.searchDefault.overTime.collapsed.timeRange;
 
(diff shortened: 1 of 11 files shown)


src/styles/DropDownMenu.ts
```diff
@@ -19,6 +19,29 @@ const sharedListStyles = {
...
+  Dark: {
+    ...sharedDateRangeStyles,
+    color: Colors.SecondaryTextDark,
+  },
+};
+
+export const DateRange = {
+  Light: style(DateRangeRaw.Light),
+  Dark: style(DateRangeRaw.Dark),
+};
```
julienw, line 43: I have a lot of comments about the styles.

### 1. About the styles themselves.

I'm not a big fan of the absolute positioning. I tried to remove them locally, and this actually gives a good result: without the absolute positioning the ranges will show inside the dropdown button too! This is good!

We have a good tool to achieve the same behavior without using absolute positioning: use flex with `justify-content: space-between`!

Therefore my suggestion is to add a flex wrapper around the text and the display of the date ranges.

### 2. How and where to specify it?

I see you tried to follow the style we're using in a lot of places. Unfortunately this style isn't super great and is very verbose. Instead we can use MUI's own theming mechanism, that _already_ supports dark mode automatically. As you can see in https://github.com/mozilla/perfcompare/blob/main/src/theme/protocolTheme.ts, `text.secondary` will point to `SecondaryText` or `SecondaryTextDark` automatically depending on the theme.

How to use it? One way is with `sx` prop in all MUI components. Another way is with the `styled` function provided with MUI (we don't use that one yet but it's a valid option -- very similar to the use of `typestyle` in its result).

You need a MUI component to use it, this doesn't work with `<span>`. But you can use `<Box>` that inserts a `<div>` (the `component` props can change the actual HTML element but a `div` works fine if we use `flex` like suggested in point 1):

```js
<Box sx={{ color: "text.secondary",  ... other properties }}> ... </Box>`
```

It is much simpler because you don't need to do the `if` condition for the dark mode, etc, and all the code related to dark mode is in one place.

Of course as you've probably noticed, the project doesn't use this trick everywhere, and sometimes it would need some signifcant change to be able to do it because it's not as straightforward as using `text.secondary`, so the way you did isn't bad either, it's just more complicated than what it could be with our current state.

julienw (changes requested):
Thanks Sumair!

This looks great and works fine !
However here are a few changes that are IMO needed before we can land this :)
Tell me what you think!

sumairq:
Thank you @julienw  for the very high-quality and thorough review!
I am learning a lot from your feedback, and it has improved not only the code but also my thinking process. :raised_hands:

src/components/Search/TimeRangeDropdown.tsx
```diff
@@ -43,6 +56,16 @@ function TimeRangeDropdown({
   const menuItemStyles =
     mode === 'light' ? DropDownItems.Light : DropDownItems.Dark;
 
+  const displayDateRange = (value: number) => {
+    const options: Intl.DateTimeFormatOptions = {
+      dateStyle: 'medium',
+    };
+    const dateFormatter = new Intl.DateTimeFormat('en-US', options);
```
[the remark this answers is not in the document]
julienw: Because you'll have to use the same date formatter in SearchOverTime (see above), it would be a good idea to move the format part to another file. I'd suggest a new file `src/utils/format.js`.

In it you could have:
```js
const DATE_FORMATTER = new Intl.DateTimeFormat('en-US', { dateStyle: 'medium' });

export function formatDate(date: Date) {
  return dateFormatter.format(date);
}

export function formatDateRange(date1: Date, date2: Date) {
  return formatDate(date1) + " - " + formatDate(date2);
}
```

And then from `TimeRangeDropdown` you could use `formatDateRange` and just add the parentheses and the extra space.

In the future `formatDateRange` could be improved to not display useless things (maybe), for example when the same year we wouldn't display it. But don't bother with that now.

src/components/Search/TimeRangeDropdown.tsx
```diff
@@ -65,9 +88,17 @@ function TimeRangeDropdown({
             <MenuItem
               value={value}
               key={value}
-              className={`timerange-dropdown-item ${menuItemStyles}`}
+              className={`timerange-dropdown-item ${menuItemStyles} ${styles.menuItem}`}
             >
-              {text}
+              <div className={styles.dateRange}>
```
julienw, line 93: you don't need the extra div, just add the `space-between` to the MenuItem directly.

julienw (changes requested):
Thanks, this is much better!

It would be good to change also the text displayed in the results view, when the edit mode isn't active. This happens in `SearchOverTime` line 157 https://github.com/mozilla/perfcompare/blob/3a555e336889620c957bd58294994f51ef9e981a/src/components/Search/SearchOverTime.tsx#L157

Can you look at it?

I also left some other nits to the existing code, tell me what you think

src/components/Search/TimeRangeDropdown.tsx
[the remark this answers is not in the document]
sumairq: Great idea!

 I went ahead and moved the formatting logic into `src/utils/format.ts` as suggested. This will definitely help keep things clean and reusable. 
 
 Thanks for the feedback! :+1:

sumairq:
Another question : 
Do you think a space between paranthesis and the Date Ranges is better than no space eg

Space             |  No Space
image |  image

julienw:
> @julienw Another question : Do you think a space between paranthesis and the Date Ranges looks better than no space ?
> Space 	No Space
> image 	image

I'd go with the no space version :-) thanks for asking

src/utils/format.ts
```diff
@@ -0,0 +1,11 @@
...
+  dateStyle: 'medium',
+});
+
+export function formatDate(date: Date) {
+  return dateFormatter.format(date);
+}
+
+export function formatDateRange(date1: Date, date2: Date) {
+  return formatDate(date1) + ' - ' + formatDate(date2);
+}
```
julienw, line 11: you know what, I just realized that Intl.DateTimeFormat knows how to format ranges (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatRange).
But I'm not sure we want that in this context.

From my test, the best formatter would be:
```js
const dateRangeFormatter = new Intl.DateTimeFormat('en-US', {
  month: "short",
  day: "numeric"
});
// then
export function formatDateRange(date1: Date, date2: Date) {
  return dateRangFormatter.formatRange(date1, date2);
}
```

By not specifying the year, it would show up only when needed.

Can you please try that and see how this looks?

sumairq: Here is how it looks:

month: "short", day: "numeric"            |  dateStyle: 'medium'
image  |  image



I like how concise it is while still conveying the complete information.

julienw: Yeah, I'd like the concise format better if this was showing up separate from the others.
But in this case, I find the format at the right easier to parse visually, because the same operation is present at the same location for every rows.
So let's keep `dateStyle: medium`, but thanks for trying it!

src/components/Search/TimeRangeDropdown.tsx
julienw: thanks for the explanation, let's keep it like you did it, but please add a comment before line 47 with this explanation:
```
// This extra flexible div is needed, so that the information looks the same in 
// both the MenuItem (in the dropdown options) and the SelectedItem (in
// the dropdown, when the dropdown is closed).
```

julienw (approved):
This looks good to me with the extra comment, thanks!

You'll need to rebase and regenerate the snapshots as well, because they conflict with a recent commit.

PR #764, opened 2024-10-08
PCF-425 Add colors and icons for the confidence information

esanuandra:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/PCF-425)
This PR adds colorful icons to confidence column (Low, Medium, High).

[deploy link with the example](https://deploy-preview-764--mozilla-perfcompare.netlify.app/compare-over-time-results?baseRepo=mozilla-central&selectedTimeRange=1209600&newRev=9b3deb938fdd67b35a981cf1613a817dfb7eb65f&newRepo=mozilla-central&framework=1)

Screenshot 2024-10-08 at 4 52 33 PM

### DIFF
--- a/src/components/CompareResults/RevisionRow.tsx
+++ b/src/components/CompareResults/RevisionRow.tsx
@@ -1,8 +1,11 @@
 import { useState, type ReactNode } from 'react';
 
 import AppleIcon from '@mui/icons-material/Apple';
+import DragHandleIcon from '@mui/icons-material/DragHandle';
 import ExpandLessIcon from '@mui/icons-material/ExpandLess';
 import ExpandMoreIcon from '@mui/icons-material/ExpandMore';
+import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
+import KeyboardArrowUpIcon from '@mui/icons-material/KeyboardArrowUp';
 import ThumbDownIcon from '@mui/icons-material/ThumbDown';
 import ThumbUpIcon from '@mui/icons-material/ThumbUp';
 import TimelineIcon from '@mui/icons-material/Timeline';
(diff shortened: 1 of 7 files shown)


beatrice-acasandrei (approved):

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -265,10 +265,10 @@ const platformIcons: Record<PlatformShortName, ReactNode> = {
   Unspecified: '',
 };
 
-const confidenceIcons = {
-  Low: <KeyboardArrowDownIcon sx={{ color: '#D7264C' }} />,
-  Medium: <DragHandleIcon sx={{ color: '#5B5B66' }} />,
-  High: <KeyboardArrowUpIcon sx={{ color: '#017A40' }} />,
+const confidenceIcon = {
```
julienw, line 268: just wondering why you rename it here, but not in the Subtests page :) I don't mind which one you use (with s or without s) but let's try to use the same one.

src/theme/protocolTheme.ts
```diff
@@ -23,6 +23,10 @@ const lightMode = {
     secondary: Colors.SecondaryText,
     disabled: Colors.TextDisabled,
   },
+  icons: {
+    success: Colors.IconLightSuccess,
+    error: Colors.IconLightError,
+  },
```
julienw, line 29: I would have added all of them to `text` above but I don't mind either, we can always revisit if we see it's useful for text as well.

julienw (approved):
thanks!

src/theme/protocolTheme.ts
esanuandra: I would go with revisit later for this one. I don't see it used on texts yet.

src/components/CompareResults/RevisionRow.tsx
esanuandra: Ah, missed that, there was a `getConfidenceIcon` function at some point, but changed my mind.

PR #768, opened 2024-10-08
Display Local Time Zones in Search Results and Revisions

### LINKED ISSUE #767: Display Local Time Zones for Revisions in Search Results and Selected Revisions
**Description:**  
Display the local time zones of revisions in both the search results and the selected revision

**Steps to Reproduce:**

1. Read the [PerfCompare Documentation](https://docs.google.com/document/d/1cpQEZXw0M5QjmNL2F1S9NKjWmz6A9Ks7zELVNTXLeB4/edit#heading=h.u62ha0d6mpxl) to get familiar with the app.
2. Navigate to the home page: https://perf.compare/ 
3. Click on any search input
4. Observe that the time zones are not included in the revisions
image
5. Observe the time zones are not included in the selected revision
image

**Acceptance Criteria:**
1. Display the local time zone alongside each revision timestamp in the search results.
2. Ensure the local time zone is also displayed for any selected revision.
3. The time zone should be correctly formatted and easy to read.

image

osujipius:
This pull request adds the user's local timezone in the search results and selected revision section.
closes #767
Note - The dayjs library used previously for the date and time did not have a clean way to access the timezone, you would need to install extra plugins which could be expensive.

Screenshot 2024-10-08 at 22 29 46
Screenshot 2024-10-08 at 22 29 58
Screenshot 2024-10-09 at 18 07 40


osujipius:
@Carla-Moz Please take a look when you can

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -68,6 +68,21 @@ function SearchResultsListItem({
...
+      hour: '2-digit',
+      minute: '2-digit',
+      hour12: false,
+      timeZoneName: 'short',
+    })
+    .replace(',', '') // Removes the comma between the date and time
+    .replace(/(\d{2})\/(\d{2})\/(\d{2})/, '$2/$1/$3'); // Reformat date as MM/DD/YY
+
+  const timeZone = localDateTime.split(' ').pop();
+
```
syedbarimanjan, line 85: You need to show the utc time in the tooltip not the localtime.

osujipius: Ahhh Ok 
That makes more sense... will correct that now

osujipius:
@syedbarimanjan I have corrected the issue in the tooltip

syedbarimanjan:
> @syedbarimanjan I have corrected the issue in the tooltip

You should update the images too.

osujipius:
> > @syedbarimanjan I have corrected the issue in the tooltip
> 
> You should update the images in the pr description too.

Done

src/components/Search/SelectedRevisionItem.tsx
```diff
@@ -51,6 +50,18 @@ function SelectedRevisionItem({
...
+      year: '2-digit',
+      month: '2-digit',
+      day: '2-digit',
+      hour: '2-digit',
+      minute: '2-digit',
+      hour12: false,
+      timeZoneName: 'short',
+    })
+    .replace(',', '') // Removes the comma between the date and time
+    .replace(/(\d{2})\/(\d{2})\/(\d{2})/, '$2/$1/$3'); // Reformat date as MM/DD/YY
```
julienw, line 64: This is very fragile, but I understand why you did it.

I need to explain a few things:

The screenshot Carla shared in #767 shows the date and time using the USA notation: MM/DD/YY. But this is confusing to all other users. Indeed in most other parts of the world, we use DD/MM/YY. Therefore for some dates it's confusing. For example 01/10/24 could be either January 10 or October 1st.

My recommendation would be to use `month: "short"` for this reason.

For day, hour, minute, I'd use "numeric" so that they can be 1 digit if possible.

You need to understand that `toLocaleString` with the first argument of `undefined` will return a different format for every language (and even possibly for different browsers and operating systems as well). So using `replace` afterwards won't work always. Instead we need to embrace the fact that the result won't always be exactly the same as specified.

So please remove these `replace` :-)

Finally I think we want to force the `en-US` locale because the interface is in english only currently, so please replace `undefined` with `en-US`.

julienw: The same comment applies to the other code above.

julienw: And we also want the tooltip here by the way :-) It could be good to extract this little piece of functionality (showing a date with both the local time + the utc time as a tooltip) to a separate component that we could reuse in these 2 places.

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -110,24 +137,26 @@ function SearchResultsListItem({
...
-                <div className='info-caption'>
-                  <div className='info-caption-item item-author'>
-                    {' '}
-                    <MailOutlineOutlinedIcon
-                      className='mail-icon'
-                      fontSize='small'
-                    />{' '}
-                    {item.author}
-                  </div>
+                <Tooltip title={utcDateTime} placement='top'>
```
julienw, line 140: I think that the tooltip should be on the date time only, if that's possible.

julienw (changes requested):
Thanks! I pointed out how things can be improved. Hope this makes sense!

osujipius:
@julienw All improvements made! Please take another look when you can

src/components/Search/SearchResultsListItem.tsx
osujipius: Oh I missed the part you asked why I wasn't using dayjs. Well cause you would need to install additional plugins to work with timezones, so I thought the native javascript way would be better

src/components/Search/SearchResultsListItem.tsx
osujipius: I have tweaked this

src/components/Search/DateTimeDisplay.tsx
```diff
@@ -0,0 +1,34 @@
...
+  // Format for local time
+  const localDateTime = itemDate?.toLocaleString('en-US', {
+    year: '2-digit',
+    month: 'short',
+    day: 'numeric',
+    hour: 'numeric',
+    minute: 'numeric',
+    hour12: false,
+    timeZoneName: 'short',
+  });
```
julienw, line 13: instead of calling `toLocalString` repeatedly, which can be a performance issue, you can create a Intl.DateTimeFormat outside of the component.

Outside of the component, you'd do something like this:
```js
const localDateFormat = new Intl.DateTimeFormat('en-US', { ... });
const utcDateFormat = new Intl.DateTimeFormat('en-US', { ... });
``

Then inside the component, you can use it this way:
```js
const localDateTime = localDateFormat.format(itemDate);
const utcDateTime = utcDateFormat.format(itemDate);
```

This way, the expensive initialization of the date-time locale formatting is done just once at startup.

julienw (commented):
it's pretty good, I think we'll be able to merge this after the last change :-)
thanks!

src/components/Search/DateTimeDisplay.tsx
osujipius: That makes perfect sense!
And I have tweaked it

osujipius:
> @osujipius Hi I'd like to let you know you've been selected to go on the next round! Congrats! Please see the matrix channel for details https://chat.mozilla.org/#/room/#perfcompare-outreachy:mozilla.org

Thanks Carla! 🥳

PR #770, opened 2024-10-08
Issue # 767 Display Local Time Zones for Revisions in Search Results and Selected Revisions

### LINKED ISSUE #767: Display Local Time Zones for Revisions in Search Results and Selected Revisions

hibaa03:
This pull request implements the following changes close #767 :

**Display Local Time Zones:**
Added functionality to display the local time zone alongside each revision timestamp. It shows me time in my local timezone (GMT+5) as in the screenshot below.
Image 17-10-2024 at 11 27 PM

**UTC Timestamp Tooltip:**
Introduced a tooltip that shows the UTC timestamp when hovering over the local time. 
Image 17-10-2024 at 11 27 PM (1)

**Time Zone in Selected Revision:**
Ensured the local time zone is also displayed for any selected revision, along with the UTC on hover through a tooltip.
Image 17-10-2024 at 11 27 PM (2)

**Compactness:**
Displayed time and time zone in a more digestible format. Created a separate DateTime component to handle both the GMT display and UTC tooltip functionality, simplifying the time display logic.


src/__tests__/Search/SearchResultsList.test.tsx
syedbarimanjan, line 1: You have accidentally removed the comments from this file.

syedbarimanjan:
Add a ```close #bug-number``` to your pr description so that if the pr gets merged the issue will close automatically.

src/__tests__/Search/SearchResultsList.test.tsx
hibaa03: thanks for pointing it out, fixed it.

syedbarimanjan: You have put some comments on wrong lines.

hibaa03: could you please point me to it?

hibaa03: sorry nevermind. working on it

Carla-Moz:
Great effort, Hiba, however, the design requires a more compact version of the timezones. See below. 

image

Your version also breaks the style for my time zone:

image

hibaa03:
Hi @Carla-Moz, I appreciate your feedback. I have made the changes accordingly, could you please review it when possible?

package.json
```diff
@@ -24,13 +24,14 @@
     "json-e": "^4.7.0",
     "material-ui-popup-state": "^5.1.2",
     "moize": "^6.1.6",
+    "moment-timezone": "^0.5.46",
```
julienw, line 27: We don't really want to add `moment` as we already have `dayjs` :-) `moment` is a huge library.

Also nowadays the web platform already has this functionality, you can read about it on MDN => https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString
It's possible we can replace `dayjs` in the process too.

src/__tests__/CompareResults/ResultsView.test.tsx
```diff
@@ -16,6 +16,8 @@ import {
   FetchMockSandbox,
 } from '../utils/test-utils';
 
+jest.setTimeout(60000);
```
julienw, line 19: you probably don't need this change :-) right? Can you explain why you added these?

julienw (changes requested):
Adding `moment` isn't great for us because it's a huge library. Instead you can use either `dayjs` that we already have, or directly the web platform API `toLocaleString` that I mention in my comment below.

Both options would work for me, feel free to pick one.
dayjs documentation about timezones is in https://day.js.org/docs/en/plugin/timezone.

src/__tests__/CompareResults/ResultsView.test.tsx
hibaa03: I added the jest.setTimeout(60000) to prevent timeout issues during tests as they were failing because of it. The only way I could solve it and have all tests passed

package.json
hibaa03: Initially I used day.js with timezone plugin to handle time zones, but the output format is limited to either IANA timezones (America/New_York) or the GMT offset (GMT+/-X), as day.js doesn’t provide abbreviations like PST or EST, so I added moment (as to make it look like the screenshot Carla has sent above, you can also see how it was before when I used dayjs (in the second screenshot she shared)). Let me know how you’d prefer me to proceed!

hibaa03: Image 11-10-2024 at 7 49 PM
This is also what I can get with day.js. Would this be better and compact (compared to IANA timezones) so I can make the changes accordingly?

hibaa03: I have made the changes! Please review them when possible

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -66,7 +72,19 @@ function SearchResultsListItem({
   const ListItemComponent = listItemComponent === 'radio' ? Radio : Checkbox;
   const revisionHash = truncateHash(item.revision);
   const commitMessage = getLatestCommitMessage(item);
-  const itemDate = new Date(item.push_timestamp * 1000);
+
+  const itemDate = dayjs.unix(item.push_timestamp);
+
+  const date = itemDate.format('MM/DD/YY');
```
julienw, line 78: The screenshot Carla shared in https://github.com/mozilla/perfcompare/issues/767 shows the date and time using the USA notation: MM/DD/YY. But this is confusing to all other users. Indeed in most other parts of the world, we use DD/MM/YY. Therefore for some dates it's confusing. For example 01/10/24 could be either January 10 or October 1st.

Therefore my recommendation would be to use `MMM` instead of `MM`, so that the first letters of the month is used. Also it would be better to use a single digit for the day when possible, so use `D` instead of `DD`.
Then I believe the format would look like `MMM D, YY`, but please check if this looks right.

This wasn't mentioned in the issue, sorry about that.

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -66,7 +72,19 @@ function SearchResultsListItem({
...
+  const localTime = itemDate.format('HH:mm');
```
julienw, line 79: similarly here, it would be better to have just 1 digit when possible, therefore you can use `H:m`.

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -66,7 +72,19 @@ function SearchResultsListItem({
...
+
+  // convert timezone to GMT
+  const gmtOffset = `GMT${itemDate
+    .format('Z')
+    .replace(':00', '')
+    .replace(/^([+-])0?/, '$1')}`;
```
julienw, line 85: I don't like this a lot, but I don't see a better solution using only dayjs :/

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -8,14 +8,21 @@ import ListItemButton from '@mui/material/ListItemButton';
...
+import utc from 'dayjs/plugin/utc';
 import { style } from 'typestyle';
 
 import { Spacing } from '../../styles';
 import type { Changeset } from '../../types/state';
 import { truncateHash, getLatestCommitMessage } from '../../utils/helpers';
 
+// day.js with timezone support
+dayjs.extend(utc);
+dayjs.extend(timezone);
```
julienw, line 24: I think you don't use the timezone plugin anymore, can you please remove this line?

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -66,7 +72,19 @@ function SearchResultsListItem({
...
+  const date = itemDate.format('MM/DD/YY');
+  const localTime = itemDate.format('HH:mm');
+
+  // convert timezone to GMT
+  const gmtOffset = `GMT${itemDate
+    .format('Z')
+    .replace(':00', '')
+    .replace(/^([+-])0?/, '$1')}`;
+
+  const utcTime = itemDate.utc().format('MM/DD/YY HH:mm [UTC]');
```
julienw, line 87: same comments as above :)

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -125,7 +142,19 @@ function SearchResultsListItem({
                       className='time-icon'
                       fontSize='small'
                     />
-                    {String(dayjs(itemDate).format('MM/DD/YY HH:mm'))}
+
+                    {/* Tooltip with UTC time */}
+                    <Tooltip title={`${utcTime}`}>
+                      <span>
+                        {date}{' '}
+                        <span style={{ fontWeight: 'bold' }}>
```
julienw, line 150: not sure why you changed the style, I think it's not needed. Then you can remove the spans as well

src/components/Search/SearchResultsListItem.tsx
```diff
@@ -125,7 +142,19 @@ function SearchResultsListItem({
                       className='time-icon'
                       fontSize='small'
                     />
-                    {String(dayjs(itemDate).format('MM/DD/YY HH:mm'))}
+
+                    {/* Tooltip with UTC time */}
+                    <Tooltip title={`${utcTime}`}>
```
julienw, line 147: it would be good to extract this functionality to a separate react component, so that you can use it both here and in SelectedRevisionItem. Can you please do that?

package.json
julienw: Can you remove this change from package.json and run npm i again?

package.json
```diff
@@ -24,13 +24,14 @@
...
     "notistack": "^2.0.5",
     "process": "^0.11.10",
     "react": "^18.2.0",
     "react-chartjs-2": "^5.2.0",
     "react-dom": "^18.2.0",
     "react-redux": "^8.0.2",
-    "react-router-dom": "^6.18.0",
+    "react-router-dom": "^6.26.2",
```
julienw, line 34: can you revert this change pleaase ? :-)

julienw (changes requested):
Thanks Hiba, this is better!
here are some more comments :-) thanks again

hibaa03:
@julienw thank you for the thorough feedback, I have made the changes accordingly. I reverted the changes made before, fixed the date and time format as you corrected, and created a separate component DateTime. 
Please give it a look when you are able to and let me know if there is anything else to change! :)

julienw:
Hey @hibaa03 , I merged the other PR that does the same thing (https://github.com/mozilla/perfcompare/pull/768). The main difference is that that PR used an API from the web platform instead of the library, which should be more performant and also takes less space in the bundle.
I'm closing your PR now, but this was really good work, congrats again!

PR #773, opened 2024-10-09
Reduce margins for search form

esanuandra:
Follow up PR for https://github.com/mozilla/perfcompare/pull/759

### DIFF
--- a/src/components/Search/CompareOverTime.tsx
+++ b/src/components/Search/CompareOverTime.tsx
@@ -154,9 +154,7 @@ function CompareOverTime({
         {hasEditButton && (
           <div
             className={`edit-btn-wrapper ${
-              hasEditButton && !formIsDisplayed
-                ? 'show-edit-btn'
-                : 'hide-edit-btn'
+              formIsDisplayed ? 'hide-edit-btn' : 'show-edit-btn'
             }`}
           >
             <EditButton onEditAction={handleEdit} mode={mode} />
--- a/src/components/Search/CompareOverTime.tsx
+++ b/src/components/Search/CompareOverTime.tsx
@@ -145,7 +145,13 @@ function CompareOverTime({
       className={`compare-card-container content-base content-base--${expandedClass} ${styles.container} wrapper--overtime ${wrapperStyles.wrapper}`}
     >
       <Divider className='divider' />
-      <Box sx={{ paddingBlock: hasEditButton ? 4 : 6, paddingInline: 6 }}>
+      <Box
+        sx={{
+          paddingInline: 6,
+          paddingBlockStart: hasEditButton ? 4 : 6,
+          paddingBlockEnd: hasEditButton && formIsDisplayed ? 4 : 6,
+        }}
+      >
         <Form
           action='/compare-over-time-results'
           onSubmit={onFormSubmit}
--- a/src/components/Search/CompareOverTime.tsx
+++ b/src/components/Search/CompareOverTime.tsx
@@ -151,15 +151,17 @@ function CompareOverTime({
         aria-label='Compare over time form'
       >
         {/**** Edit Button ****/}
-        <div
-          className={`edit-btn-wrapper ${
-            hasEditButton && !formIsDisplayed
-              ? 'show-edit-btn'
-              : 'hide-edit-btn'
-          }`}
-        >
-          <EditButton onEditAction={handleEdit} mode={mode} />
-        </div>
+        {hasEditButton && (
+          <div
+            className={`edit-btn-wrapper ${
+              hasEditButton && !formIsDisplayed
+                ? 'show-edit-btn'
+                : 'hide-edit-btn'
+            }`}
+          >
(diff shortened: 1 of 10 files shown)


src/styles/CompareCards.ts
```diff
@@ -109,7 +109,7 @@ export const CompareCardsStyles = (mode: string) => {
               height: 'auto',
             },
             '.form-wrapper': {
-              padding: `${Spacing.xxLarge}px ${Spacing.xxLarge + 2}px`,
+              padding: `${Spacing.xLarge}px ${Spacing.xxLarge + 2}px`,
             },
```
julienw, line 113: I think we want to keep the `xxLarge` on the SearchView.
One way to do it would be to have 2 different classes: `form-wrapper` and `form-wrapper-compact` for example. Then in `CompareWithBase` and `CompareOverTime` you could pick one or the other depending on the value of `hasEditButton`.

Alternatively you can remove the CSS from here and add it directly to the component, for example adding a new `<Box>` around the  `<Form>` (the Form is from react-router and can't have the `sx` prop):

```
<Box sx={{paddingBlock: hasEditButton ? 4 : 6, paddingInline: 6}}>
  <Form>
    ...
  </Form>
</Box>
```

(remember that `sx` multiplies the number with the spacing value, that is 8px)

This proposal also changes for `padding-inline` `Spacing.xxLarge + 2` (50px) to `Spacing.xxLarge` (6 * 8 = 48px), which makes more sense to me.

julienw: Thinking about this more, I believe we should go with the second solution.

Why? Because I believe the bottom padding should be even more tweaked, with many cases that are known from within the component:

1. hasEditMode = false => bigger padding block, to give more space on the home page.
2. hasEditMode = true, formDisplayed = false => the _bottom_ padding (that is `padding-block-end`) should be larger, because the _visual_ top padding is larger, due to the edit button taking some space in addition to the real top padding. Therefore orobably 48px would be visually looking good.
3. hasEditMode = true, formDisplayed = true => the bottom padding can move back to 32px because the cancel and compare button take some space too.

Can you try this and we'll see how it looks like?

julienw (changes requested):
Thanks, this looks pretty good to me.

I see there's one behavior change: previously, when loading the results, the "Compare" button was showing up, but disabled with a "loading..." text. Now with this patch, it doesn't appear anymore. 

I think it's OK because the spinner makes it obvious that the results are being fetched, but I'd like to know what @Carla-Moz thinks.

Note that it's showing up only if the user click on "edit mode" again.... but in that case I believe we should probably make it interactive so that the user can change their mind... But that's for another PR I think.

I'm marking "request changes" because I believe the paddings need to be tweaked some more, see the comments for more information.

Carla-Moz:
> Thanks, this looks pretty good to me.
> 
> I see there's one behavior change: previously, when loading the results, the "Compare" button was showing up, but disabled with a "loading..." text. Now with this patch, it doesn't appear anymore.
> 
> I think it's OK because the spinner makes it obvious that the results are being fetched, but I'd like to know what @Carla-Moz thinks.
> 
> Note that it's showing up only if the user click on "edit mode" again.... but in that case I believe we should probably make it interactive so that the user can change their mind... But that's for another PR I think.
> 
> I'm marking "request changes" because I believe the paddings need to be tweaked some more, see the comments for more information.

I think it's fine as is. The spinner is enough feedback.

As for the edit mode when the button is in disabled/loading mode for the framework change for example, it happens fairly quickly that I honestly don't think it's worth changing to make it interactive.

julienw:
> 
> I think it's fine as is. The spinner is enough feedback.

Thanks!

> 
> As for the edit mode when the button is in disabled/loading mode for the framework change for example, it happens fairly quickly that I honestly don't think it's worth changing to make it interactive.

There are a lot of cases where it's fast :-)
But there are also some cases where it's fairly slow:
* with the browsertime framework there are many requests and this makes the serverside call slower.
* with compare over time over a longer time, this can be especially slow too.

Not urgent though!

julienw (changes requested):
I think we can make the padding even more appealing, tell me what you think!

src/components/Search/CompareOverTime.tsx
```diff
@@ -144,53 +145,65 @@ function CompareOverTime({
...
-        aria-label='Compare over time form'
-      >
-        {/**** Edit Button ****/}
-        <div
-          className={`edit-btn-wrapper ${
-            hasEditButton && !formIsDisplayed
-              ? 'show-edit-btn'
-              : 'hide-edit-btn'
-          }`}
+      <Box sx={{ paddingBlock: hasEditButton ? 4 : 6, paddingInline: 6 }}>
```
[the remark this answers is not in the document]
esanuandra: Sorry, it didn't click to me that formDisplayed is a flag. I thought the whole thing was an explanation for hasEditMode changes.
Thank you, I will try this.

julienw (approved):
lgtm!

PR #774, opened 2024-10-09
enhancement: Make results tiltle editable and URL persistent

### LINKED ISSUE #769: Results Page: Allow users to change the page title of PerfCompare and have the title persist in the url
**Description:** As a user, I want to be able to change the title of PerfCompare and have the new title persist in the url, so that I can share the updated title with others.

**Steps to Reproduce:**

1. Read the PerfCompare Documentation to get familiar with the app.
2. Navigate to the title section of the PerfCompare results: [ResultsPage](https://main--mozilla-perfcompare.netlify.app/compare-results?baseRev=901eda21e48867cdcc016a0dba81e51c0d4dd16b&baseRepo=mozilla-central&newRev=b48e31d47d1f562424c7693ad93f74ed39251edb&newRepo=mozilla-central&newRev=e5eef47a4db1acee5e17f5b4ea6219301d995c63&newRepo=mozilla-central&framework=13.)
3. Observe that there is currently no option to modify the title and have it persist in the URL.

image

**Acceptance Criteria:**
1. The title should be editable by the user.
2. The new title should persist in the URL so that it is retained after a page refresh.
3. When a user shares the URL with others, the custom title should be displayed.
4. The url should have the updated page title.

FatumaA:
This PR makes the Results page title editable and persists in the URL state. 

Before, the Results view had "Results" as the title. This title could not be customized and was not part of the URL parameters.
 
Now: 

A screenshot of the results page with an arrow pointing to the title and a thick red line below the URL 
params at the top of the image

Before the changes:

A screenshot of the results page with an arrow pointing to the title and a thick red line below the URL 
params at the top of the image

Closes #769


syedbarimanjan:
You should look into the CI tests which are failing.

hibaa03:
Hi, could you add the screenshots to the description for more clarification? Also run the tests and validations locally to see which ones are failing and work on the issues as you will see on your terminal. Maybe its due to jest snapshot which you will have to update

julienw (changes requested):
Thanks, this looks pretty good as a first draft!

I left a comment about the mechanism to change the search params.

Otherwise I think that it should be make editable after a click on a button.

So here is how it could look like:

1. at load time, we still have the "Results" by default (or the value in the URL if present) => the good thing is that most of the tests would still work.
2. add a little edit icon (with MUI's Edit icon) next to this text. => i'd put it on the right of the text, this would look like : Results [Edit icon] | Compare with a base
3. when the user clicks the edit icon, then the text changes into an input like you did.

It would be good to extract this functionality into a separate react component.

Finally it would be good to add a test to test this functionality.

Thanks again!

FatumaA:
> Thanks, this looks pretty good as a first draft!
> 
> I left a comment about the mechanism to change the search params.
> 
> Otherwise I think that it should be make editable after a click on a button.
> 
> So here is how it could look like:
> 
> 1. at load time, we still have the "Results" by default (or the value in the URL if present) => the good thing is that most of the tests would still work.
> 2. add a little edit icon (with MUI's Edit icon) next to this text. => i'd put it on the right of the text, this would look like : Results [Edit icon] | Compare with a base
> 3. when the user clicks the edit icon, then the text changes into an input like you did.
> 
> It would be good to extract this functionality into a separate react component.
> 
> Finally it would be good to add a test to test this functionality.
> 
> Thanks again!

Thanks for the review. Working on it 🙂.

FatumaA:
> Thanks, this looks pretty good as a first draft!
> 
> I left a comment about the mechanism to change the search params.
> 
> Otherwise I think that it should be make editable after a click on a button.
> 
> So here is how it could look like:
> 
> 1. at load time, we still have the "Results" by default (or the value in the URL if present) => the good thing is that most of the tests would still work.
> 2. add a little edit icon (with MUI's Edit icon) next to this text. => i'd put it on the right of the text, this would look like : Results [Edit icon] | Compare with a base
> 3. when the user clicks the edit icon, then the text changes into an input like you did.
> 
> It would be good to extract this functionality into a separate react component.
> 
> Finally it would be good to add a test to test this functionality.
> 
> Thanks again!

Hi, 
I think this is ok now. Except for the tests. I've commented out the last 2 tests because they keep timing out and I'm not sure how to go about that.

I'd appreciate a look there, thank you 🙏🏽

hibaa03:
Make sure your internet connection isn't slow and you release some memory of your system if needed. And run the tests using npm test --  --testTimeout 60000 if they're still timing out.

FatumaA:
> Make sure your internet connection isn't slow and you release some memory of your system if needed. And run the tests using npm test -- --testTimeout 60000 if they're still timing out.

Could be slow internet, I did increase timeout and same result

src/components/CompareResults/ResultsTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+      });
+
+      setIsEditing(false);
+    }
+  };
+
+  return (
+    <>
+      <div className={isEditing ? styles.hide : styles.title}>
+        <label htmlFor='results' data-testid='results-title-label'>
```
julienw, line 64: it doesn't make a lot of sense to use a `label`. Indeed the label and the input are never present together.

Just use a normal span :-)

src/components/CompareResults/ResultsTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+          {resultsTitle}
+        </label>
+        <EditIcon
+          className={styles.editIcon}
+          data-testid='edit-icon'
+          onClick={handleEdit}
+        />
+      </div>
+      {isEditing && (
+        <Input
```
julienw, line 74: I remember you had a label just for screen readers here, in the first version of your patch; that was a neat idea! so you can add it back here.

src/components/CompareResults/ResultsTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+    <>
+      <div className={isEditing ? styles.hide : styles.title}>
+        <label htmlFor='results' data-testid='results-title-label'>
+          {resultsTitle}
+        </label>
+        <EditIcon
+          className={styles.editIcon}
+          data-testid='edit-icon'
+          onClick={handleEdit}
+        />
```
julienw, line 71: I realize we were using a different icon in https://github.com/mozilla/perfcompare/blob/3a555e336889620c957bd58294994f51ef9e981a/src/components/Search/EditButton.tsx#L16-L23
You can copy this code (you don't need the id and the classes). You can pass the theme mode from ResultsTitle as a prop.

Also you need to embed this icon in an IconButton as explained in https://mui.com/material-ui/react-button/#icon-button, the onClick will be on that.

src/components/CompareResults/ResultsTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+        method: 'push',
+      });
+
+      setIsEditing(false);
+    }
+  };
+
+  return (
+    <>
+      <div className={isEditing ? styles.hide : styles.title}>
```
julienw, line 63: instead of using a `styles.hide`, it's easier and more readable in this case to do a ternary:
```js
{isEditing ? <Input ... > : <div>...</div>}
```

This way it's super clear that we have one or the other.

src/components/CompareResults/ResultsTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+    title: style({
+      ...FontsRaw.HeadingXS,
+      fontWeight: 700,
+      letterSpacing: '-0.01em',
+      margin: 0,
+    }),
+    editIcon: style({
+      marginLeft: '10px',
+      cursor: 'pointer',
+      color: Colors.LinkText,
```
julienw, line 27: you won't probably need this if you use the IconButton component.

src/components/CompareResults/ResultsTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+export const ResultsTitle = () => {
+  const [searchParams, updateSearchParams] = useRawSearchParams();
+  const [resultsTitle, setResultsTitle] = useState(
+    searchParams.get('title') || 'Results',
+  );
+  const [isEditing, setIsEditing] = useState(false);
+
+  const styles = {
+    title: style({
```
julienw, line 18: Also the text and the icon will need to be vertically aligned. If you use a `display :flex` or `display: inline-flex` + `align-items: center` this should make them aligned.

src/components/CompareResults/ResultsTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+      ...FontsRaw.HeadingXS,
+      fontWeight: 700,
+      letterSpacing: '-0.01em',
+      margin: 0,
```
julienw, line 22: I think they should be applied on the title span only, but not the full title div containign the icon.

src/components/CompareResults/ResultsTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+        </label>
+        <EditIcon
+          className={styles.editIcon}
+          data-testid='edit-icon'
+          onClick={handleEdit}
+        />
+      </div>
+      {isEditing && (
+        <Input
+          id='results'
```
julienw, line 75: id is not needed :-)

src/components/CompareResults/ResultsTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+    setResultsTitle(newTitle);
+  };
+
+  const handleEdit = () => setIsEditing(true);
+
+  const handleEditComplete = (
+    event: React.FocusEvent | React.KeyboardEvent,
+  ) => {
+    if (
+      (event as React.KeyboardEvent).key === 'Enter' ||
```
julienw, line 45: It would be good to handle 'Escape' as a 'cancel' operation as well. Then it would just setIsEditing but do not update anything otherwise.

src/components/CompareResults/ResultsTitle.tsx
```diff
@@ -0,0 +1,87 @@
...
+      display: 'none',
+    }),
+  };
+
+  const handleResultsChange = (event: React.ChangeEvent<HTMLInputElement>) => {
+    const newTitle = event.target.value.trim();
+    setResultsTitle(newTitle);
+  };
+
+  const handleEdit = () => setIsEditing(true);
```
julienw, line 39: it would be good to set the focus on the input. You can do it this way:
1. add a react reference on the input (look at `React.createRef`)
2. in an effect:
```js
useEffect(() => {
  if (isEditing && inputRef.current) {
    inputRef.current.focus();
    inputRef.current.select();
  }
}, [isEditing]);
```

julienw (changes requested):
Thanks!
I left a lot of comments but I'd like to highlight that this is very good work still. This is difficult to get right! Hopefully we can get this patch to a mergeable state :-)

src/components/CompareResults/ResultsTitle.tsx
FatumaA: I think the ID is needed to tie it to the label for accessibility purposes. Unless we nest the input inside the label?

FatumaA:
> Thanks! I left a lot of comments but I'd like to highlight that this is very good work still. This is difficult to get right! Hopefully we can get this patch to a mergeable state :-)

Thank you Julien 😊. I think you can take another look now.

Carla-Moz:
@FatumaA Hi I'd like to let you know you've been selected to go on the next round! Congrats! 

FatumaA:
> @FatumaA Hi I'd like to let you know you've been selected to go on the next round! Congrats! Please see the matrix channel for details https://chat.mozilla.org/#/room/#perfcompare-outreachy:mozilla.org

Thank you Carla! 🙏🏽

src/components/CompareResults/ResultsTitle.tsx
julienw: ah yes you're right, thanks for correcting me

Issue #766, opened 2024-10-08
Results Page: Sort Results tests by Confidence - High Confidence first by default 


Carla-Moz:
**Description:**  
As a user, I want the results view to be sorted by confidence level, with high-confidence results displayed first by default, so that I can quickly access the most reliable information without manually sorting.

Steps to Reproduce:

1. Read the [PerfCompare Documentation](https://docs.google.com/document/d/1cpQEZXw0M5QjmNL2F1S9NKjWmz6A9Ks7zELVNTXLeB4/edit#heading=h.u62ha0d6mpxl) to get familiar with the app.
2. Navigate to the results table: [Results](https://perf.compare/compare-results?baseRev=901eda21e48867cdcc016a0dba81e51c0d4dd16b&baseRepo=mozilla-central&newRev=b48e31d47d1f562424c7693ad93f74ed39251edb&newRepo=mozilla-central&newRev=e5eef47a4db1acee5e17f5b4ea6219301d995c63&newRepo=mozilla-central&framework=13.)
3. Observe that the column filters are not sorted by confidence level, with high-confidence results displayed first by default

![image](https://github.com/user-attachments/assets/582126f8-722c-4073-87c6-e2ef95e9902f)

**Acceptance Criteria:**
1. Results should be sorted by confidence in descending order (high confidence first) by default.
![image](https://github.com/user-attachments/assets/30d2910a-6939-4be1-b5b9-3329b47d02f5)

mercybassey:
@Carla-Moz can i work on this?

Carla-Moz:
Hi @mercybassey feel to work on it!

PR #775, opened 2024-10-09
sorted confidence level in decending order

mercybassey:
This PR displays results in descending order based on confidence levels, from high to medium and then to low.

Issue: #766


hibaa03:
Hi, check the README doc to check and run the validators and tests if you're facing issues with it

mercybassey:
> Hi, check the README doc to check and run the validators and tests if you're facing issues with it

Thank you @hibaa03. I am currently working on it.

mercybassey:
@Carla-Moz my tests are still failing, I'd really appreciate your help.

julienw:
I believe that the test issues come from the fact that now that the "high" confidence value is selected by default, this filters out the results. If you look at the test data we're using in our tests (src/__tests__/utils/fixtures.ts), you'll see that some of the results are using `confidence: 'High'`, however in some tests we use only the first result (because it makes the test faster) that has `confidence: 'Low'`. Without looking closer, I think this is the issue.

I think the easiest fix would be to change the first result in `testCompareData` to have `confidence: 'High'` so that it would be displayed in these tests.
Another fix would be to manually select the other options in the filter, in each of these tests, but this seems more work.

julienw (changes requested):
The initial issue incorrectly mentioned to _sort_ the values, but we really just want to _filter_ them. Sorting is too complex with our current structure. Thanks for your understanding.
Can you please make these changes?

mercybassey:
> The initial issue incorrectly mentioned to _sort_ the values, but we really just want to _filter_ them. Sorting is too complex with our current structure. Thanks for your understanding. Can you please make these changes?

Sure

mercybassey:
@julienw it seems just changing the first result in `testCompareData` to have confidence: 'High' didn't resolve the issue.

mercybassey:
> @julienw it seems just changing the first result in `testCompareData` to have confidence: 'High' didn't resolve the issue.

This is my result locally. 
perf

julienw:
Hey @mercybassey, I'm gonna close this PR now, because some recent changes in our codebase made this change less useful and we'd like to gather feedback before doing it. But thanks again for your work!

PR #777, opened 2024-10-09
#766 Modified confidence filter to display results by high confidence by default

bhekidaweti:
Updated filters to sort by confidence level to high by default

### DIFF
--- a/src/components/CompareResults/ResultsTable.tsx
+++ b/src/components/CompareResults/ResultsTable.tsx
@@ -104,7 +104,7 @@ export default function ResultsTable() {
   const [searchTerm, setSearchTerm] = useState(initialSearchTerm);
   const [frameworkIdVal, setFrameworkIdVal] = useState(frameworkId);
   const [tableFilters, setTableFilters] = useState(
-    new Map([['confidence', new Set(['High'])]]) // ColumnID -> Set<Values to remove>// Default confidence filter
+    new Map([['confidence', new Set(['High'])]]), // ColumnID -> Set<Values to remove>// Default confidence filter
   );
 
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


hibaa03:
HI, for the CI tests failing read [this](https://github.com/mozilla/perfcompare?tab=readme-ov-file#running-all-validators-and-tests) part of the README.

bhekidaweti:
Thanks, I am working on it.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -104,9 +104,17 @@ export default function ResultsTable() {
...
+    new Map([['confidence', new Set(['High'])]]), // ColumnID -> Set<Values to remove>// Default confidence filter
   );
 
+  useEffect(() => {
+    setTableFilters((prevFilters) => {
+      const newFilters = new Map(prevFilters);
+      newFilters.set('confidence', new Set(['High']));
+      return newFilters;
+    });
+  }, []);
```
julienw, line 116: can you please explain why you used this effect in addition to the default state above?
Thanks

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -104,9 +104,17 @@ export default function ResultsTable() {
   const [searchTerm, setSearchTerm] = useState(initialSearchTerm);
   const [frameworkIdVal, setFrameworkIdVal] = useState(frameworkId);
   const [tableFilters, setTableFilters] = useState(
-    new Map() as Map<string, Set<string>>, // ColumnID -> Set<Values to remove>
+    new Map([['confidence', new Set(['High'])]]), // ColumnID -> Set<Values to remove>// Default confidence filter
```
julienw, line 107: Have you tried the code locally? This does the opposite of what you think: this _removes_ all items with the "high" confidence, we want to do the opposite.

julienw (changes requested):
thanks for your contribution,
please note that the code does the opposite of what's needed: it removes all results with the "high" confidence instead of keeping them only.

Thanks!

Issue #765, opened 2024-10-08
Results Page: Persist the column filters in the results table in the url 


Carla-Moz:
Description: As a user, when I try to filter the improvements or regressions in the results table by using the column filters, I expect the filtered view to be retained when I refresh the page. However, currently, the filters are not persisted, and the default view is restored upon refreshing.

Acceptance Criteria:

    Column filters should persist in the URL so that the exact same filtered view is retained after a page refresh.
    When a user shares the filtered view URL with others, they should see the same filtered results.
    The URL should dynamically update as filters are applied or removed.

Steps to Reproduce:
    Read the [PerfCompare Documentation](https://docs.google.com/document/d/1cpQEZXw0M5QjmNL2F1S9NKjWmz6A9Ks7zELVNTXLeB4/edit#heading=h.u62ha0d6mpxl) to get familiar with the app. 
    Navigate to the results table: [Results](https://perf.compare/compare-results?baseRev=901eda21e48867cdcc016a0dba81e51c0d4dd16b&baseRepo=mozilla-central&newRev=b48e31d47d1f562424c7693ad93f74ed39251edb&newRepo=mozilla-central&newRev=e5eef47a4db1acee5e17f5b4ea6219301d995c63&newRepo=mozilla-central&framework=13.)
    Apply filters to any columns (e.g., show only improvements or regressions).
    Refresh the page.
    Observe that the column filters are not persisted, and the default view is shown.

Expected Result: Filters applied to the columns should persist in the URL, and refreshing the page or sharing the URL should show the same filtered results.

![image](https://github.com/user-attachments/assets/5c958213-660d-4bd8-a5d9-828eff04e3eb)

![image](https://github.com/user-attachments/assets/36e3611f-d509-4f6a-8bbf-eded6541c8b7)

Hint!
Check out this closed PR to help you out. [Persist the search term in the url #724 ](https://github.com/mozilla/perfcompare/pull/724)

syedbarimanjan:
Hi I would like to work on this.

Carla-Moz:
@syedbarimanjan yes, go ahead and try it out! You don't need to wait to be assigned.

Netacci:
@Carla-Moz Should this be implemented for only status and not platform and confidence? or is that a separate issue?

Carla-Moz:
Good question! Yes, this should be implemented for status, platform, and confidence. I've updated the description.

PR #778, opened 2024-10-09
#765 Results Page: Persist the column filters in the results table in the url 

Netacci:
## Description

This PR addresses the issue where the results table resets during a page reload after a user applies column filters (e.g., platform, status, confidence). The filtered values do not persist, causing the results table to revert to its default state.

Before

https://github.com/user-attachments/assets/6bef395e-659b-4073-8e77-14d2e4426d99

The problem occurred because the filters were not stored in the URL, leading to their loss upon page reload or when the URL is shared. As a result, the filtered state is not retained, and users have to reapply their filters each time a page reloads.

## Test cases

1. By default, all filter values are checked, and as a result, all filter values are appended to the URL parameters.
 
   To achieve this, I used `useEffect` to set an initial filter state. This `useEffect` sets all the checked filter values to the URL.

https://github.com/user-attachments/assets/c16c8d64-cada-49b6-b45f-9c5cd2660168

2. When specific filters are applied or modified, the relevant filter values are added or updated in the URL.
  
   I adjusted the `onToggle` function to ensure that filter values are added to the URL when they are checked and removed when 
   unchecked. This helps in maintaining persistence on page reload.

https://github.com/user-attachments/assets/122e7585-ae5a-485a-98bc-2230cac24f9e

3. On page reload or URL sharing, filtered values appended to the URL remain the same thereby persisting the results table filtered state.

https://github.com/user-attachments/assets/614321c0-efa9-44b6-9077-de9647bf6f3f

4. If all filters are unchecked, the results table retains the unfiltered state, and the URL maintains the filter parameter keys with empty values. 

This was challenging to implement at first, because when all filter values are unchecked and the page reloads, the default behavior (from test case 1) would normally re-check all filters and set them back to checked. To resolve this, instead of removing the parameter keys entirely from the URL, I set their values to empty. I also updated the `useEffect` to run once on page load and check if there is existing checked values to be applied before setting the default. That way there's a significant difference between when a user has unchecked all filters and when the default filter settings should apply, ensuring the correct behavior on page load.

https://github.com/user-attachments/assets/27b825e7-e737-435c-a510-1e7513d0da70

5. When the `clear filter` button is clicked, filtered values are added back to the URL.
    I updated the `clearFilter` function to set all filter values to the URL.

https://github.com/user-attachments/assets/dc403997-b4f3-40a1-ae7a-a30b812f4562

## Steps to test

- Navigate to results table page
- Apply Column filters by platform, status and/or confidence
- Refresh the page and observe the filters persist both in the URL and the results table
- Copy the URL to another tab and observe the filtered state is still maintained.

### DIFF
--- a/src/__tests__/CompareResults/ResultsView.test.tsx
+++ b/src/__tests__/CompareResults/ResultsView.test.tsx
@@ -29,7 +29,8 @@ function renderWithRoute(component: ReactElement) {
 
   return renderWithRouter(component, {
     route: '/compare-results/',
-    search: '?baseRev=spam&baseRepo=mozilla-central&framework=2',
+    search:
+      '?baseRev=spam&baseRepo=mozilla-central&framework=2&platform=Windows%2COSX%2CLinux%2CAndroid&status=No+changes%2CImprovement%2CRegression&confidence=Low%2CMedium%2CHigh',
     loader,
   });
 }
@@ -122,7 +123,8 @@ describe('Results View', () => {
       <ResultsView title={Strings.metaData.pageTitle.results} />,
       {
         route: '/compare-results/',
-        search: '?baseRev=spam&baseRepo=mozilla-central&framework=2',
+        search:
+          '?baseRev=spam&baseRepo=mozilla-central&framework=2&platform=Windows%2COSX%2CLinux%2CAndroid&status=No+changes%2CImprovement%2CRegression&confidence=Low%2CMedium%2CHigh',
         loader,
       },
     );
@@ -198,7 +200,8 @@ describe('Results View', () => {
       <ResultsView title={Strings.metaData.pageTitle.results} />,
       {
         route: '/compare-results/',
-        search: '?baseRev=spam&baseRepo=mozilla-central&framework=2',
+        search:
+          '?baseRev=spam&baseRepo=mozilla-central&framework=2&platform=Windows%2COSX%2CLinux%2CAndroid&status=No+changes%2CImprovement%2CRegression&confidence=Low%2CMedium%2CHigh',
         loader,
       },
     );
@@ -256,7 +259,8 @@ describe('Results View', () => {
       <ResultsView title={Strings.metaData.pageTitle.results} />,
       {
         route: '/compare-results/',
-        search: '?baseRev=spam&baseRepo=mozilla-central&framework=2',
+        search:
+          '?baseRev=spam&baseRepo=mozilla-central&framework=2&platform=Windows%2COSX%2CLinux%2CAndroid&status=No+changes%2CImprovement%2CRegression&confidence=Low%2CMedium%2CHigh',
         loader,
       },
     );
@@ -325,7 +329,8 @@ describe('Results View', () => {
       <ResultsView title={Strings.metaData.pageTitle.results} />,
       {
         route: '/compare-results/',
-        search: '?baseRev=spam&baseRepo=mozilla-central&framework=2',
+        search:
+          '?baseRev=spam&baseRepo=mozilla-central&framework=2&platform=Windows%2COSX%2CLinux%2CAndroid&status=No+changes%2CImprovement%2CRegression&confidence=Low%2CMedium%2CHigh',
         loader,
       },
     );
--- a/src/__tests__/CompareResults/__snapshots__/ResultsView.test.tsx.snap
+++ b/src/__tests__/CompareResults/__snapshots__/ResultsView.test.tsx.snap
@@ -922,14 +922,14 @@ exports[`Results View The table should match snapshot and other elements should 
       >
         <button
           aria-haspopup="true"
-          aria-label="Platform (Click to filter values. Some filters are active.)"
+          aria-label="Platform (Click to filter values)"
           class="MuiButtonBase-root MuiButton-root MuiButton-contained MuiButton-containedSecondary MuiButton-sizeMedium MuiButton-containedSizeMedium MuiButton-disableElevation MuiButton-root MuiButton-contained MuiButton-containedSecondary MuiButton-sizeMedium MuiButton-containedSizeMedium MuiButton-disableElevation css-gsklu7-MuiButtonBase-root-MuiButton-root"
           tabindex="0"
           type="button"
         >
           Platform
           <svg
-            class="MuiSvgIcon-root MuiSvgIcon-colorPrimary MuiSvgIcon-fontSizeSmall css-1yguepm-MuiSvgIcon-root"
+            class="MuiSvgIcon-root MuiSvgIcon-fontSizeSmall css-reqpbo-MuiSvgIcon-root"
             data-testid="FilterListIcon"
             focusable="false"
             role="img"
@@ -939,7 +939,7 @@ exports[`Results View The table should match snapshot and other elements should 
               d="M10 18h4v-2h-4v2zM3 6v2h18V6H3zm3 7h12v-2H6v2z"
             />
             <title>
-              Some filters are active
+              No active filters
             </title>
           </svg>
           <span
@@ -969,14 +969,14 @@ exports[`Results View The table should match snapshot and other elements should 
       >
         <button
           aria-haspopup="true"
-          aria-label="Status (Click to filter values. Some filters are active.)"
+          aria-label="Status (Click to filter values)"
           class="MuiButtonBase-root MuiButton-root MuiButton-contained MuiButton-containedSecondary MuiButton-sizeMedium MuiButton-containedSizeMedium MuiButton-disableElevation MuiButton-root MuiButton-contained MuiButton-containedSecondary MuiButton-sizeMedium MuiButton-containedSizeMedium MuiButton-disableElevation css-gsklu7-MuiButtonBase-root-MuiButton-root"
           tabindex="0"
           type="button"
         >
           Status
           <svg
-            class="MuiSvgIcon-root MuiSvgIcon-colorPrimary MuiSvgIcon-fontSizeSmall css-1yguepm-MuiSvgIcon-root"
+            class="MuiSvgIcon-root MuiSvgIcon-fontSizeSmall css-reqpbo-MuiSvgIcon-root"
             data-testid="FilterListIcon"
             focusable="false"
             role="img"
@@ -986,7 +986,7 @@ exports[`Results View The table should match snapshot and other elements should 
               d="M10 18h4v-2h-4v2zM3 6v2h18V6H3zm3 7h12v-2H6v2z"
             />
             <title>
-              Some filters are active
+              No active filters
             </title>
           </svg>
           <span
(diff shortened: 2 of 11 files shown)


src/components/CompareResults/ResultsTable.tsx
```diff
@@ -104,21 +104,85 @@ export default function ResultsTable() {
...
   const onClearFilter = (columnId: string) => {
+    const possibleValues = getPossibleValues(columnId);
+
     setTableFilters((oldFilters) => {
       const newFilters = new Map(oldFilters);
       newFilters.delete(columnId);
+      const values = possibleValues?.join(',') || '';
+
+      searchParams.set(columnId, values);
+      setSearchParams(searchParams);
```
syedbarimanjan, line 126: You need to use the ```useRawSearchParams()``` hook as the ```useSearchParams()``` hook rerenders the page everytime a filter is changed. Look at this pr [here](https://github.com/mozilla/perfcompare/pull/724) to find out more.

syedbarimanjan (commented):
Left some comments.

src/__tests__/CompareResults/__snapshots__/ResultsView.test.tsx.snap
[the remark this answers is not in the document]
Netacci: Hi @syedbarimanjan Thanks, but this isn't ready for review yet. I want to fix the tests first

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -104,21 +104,78 @@ export default function ResultsTable() {
...
+    return config.possibleValues;
+  };
   const onClearFilter = (columnId: string) => {
+    const possibleValues = getPossibleValues(columnId);
     setTableFilters((oldFilters) => {
       const newFilters = new Map(oldFilters);
       newFilters.delete(columnId);
+      const values = possibleValues?.join(',') || '';
+      searchParams.set(columnId, values);
+      updateRawSearchParams(searchParams);
```
julienw, line 124: probably we should just remove this column.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -104,21 +104,78 @@ export default function ResultsTable() {
...
     setTableFilters((oldFilters) => {
       const newFilters = new Map(oldFilters);
       newFilters.set(columnId, filters);
+      const uncheckedValues = [
+        ...((newFilters.get(columnId) ?? []) as Set<string>),
+      ];
+      const allUnchecked = uncheckedValues.length === possibleValues?.length;
+      // remove value from params if all filter is unchecked
+      if (allUnchecked) {
+        searchParams.set(columnId, '');
```
julienw, line 168: I'd use `delete` to remove it instead.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -104,21 +104,78 @@ export default function ResultsTable() {
...
+    updateRawSearchParams(searchParams);
+    setTableFilters(initialFilters);
+  }, []);
+
   const onToggleFilter = (columnId: string, filters: Set<string>) => {
+    const possibleValues = getPossibleValues(columnId);
     setTableFilters((oldFilters) => {
       const newFilters = new Map(oldFilters);
       newFilters.set(columnId, filters);
+      const uncheckedValues = [
```
julienw, line 162: I'd rather do all of this in an effect with a dependency on the tableFilters state.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -104,21 +104,78 @@ export default function ResultsTable() {
   const [searchTerm, setSearchTerm] = useState(initialSearchTerm);
   const [frameworkIdVal, setFrameworkIdVal] = useState(frameworkId);
   const [tableFilters, setTableFilters] = useState(
-    new Map() as Map<string, Set<string>>, // ColumnID -> Set<Values to remove>
+    new Map(), // ColumnID -> Set<Values to remove>
   );
+  const filterableKeys = ['status', 'platform', 'confidence'];
```
julienw, line 109: it would be good to compute this from `cellsConfiguration`

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -104,21 +104,78 @@ export default function ResultsTable() {
...
+        initialFilters.set(key, new Set(values));
+      } else {
+        searchParams.set(key, possibleValues.join(','));
+        initialFilters.set(key, new Set(possibleValues));
+      }
+    });
+
+    updateRawSearchParams(searchParams);
+    setTableFilters(initialFilters);
+  }, []);
```
julienw, line 155: some suggestions to make this less complex:
1. use an effect with the "tableFilters" state to update the search params
2. do not set the state inside the effect, instead provide a default value when useState is called. It could be good to have functions to convert from the URL params to the filters and the other way around too.
3. try to not add anything to the URL if the default value is used
4. prefix the keys in the search param with some common prefix (maybe filter-XXX, eg filter-confidence)
5. I'm not sure yet about the value we use. Maybe it would be better to use some shorter keys instead of the full values, so that the displayed values can change and keep the URL working.

Happy to discuss this more if necessary!

src/__tests__/CompareResults/ResultsView.test.tsx
```diff
@@ -29,7 +29,8 @@ function renderWithRoute(component: ReactElement) {
 
   return renderWithRouter(component, {
     route: '/compare-results/',
-    search: '?baseRev=spam&baseRepo=mozilla-central&framework=2',
+    search:
+      '?baseRev=spam&baseRepo=mozilla-central&framework=2&platform=Windows%2COSX%2CLinux%2CAndroid&status=No+changes%2CImprovement%2CRegression&confidence=Low%2CMedium%2CHigh',
```
julienw, line 33: it would be good that we don't have to change all these urls... that is when we use the URL without the filters, it would just use the default filters.

julienw (changes requested):
Thanks, this looks pretty good! The implementation is a bit complex though, so I left some comments about it, tell me what you think!

src/types/types.ts
```diff
@@ -7,6 +7,7 @@ export type CompareResultsTableConfig =
       key: string;
       disable?: boolean;
       gridWidth?: string;
+      possibleValues?: string[];
```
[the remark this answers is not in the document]
Netacci: Okay noted!

src/components/CompareResults/ResultsTable.tsx
Netacci: Thanks for these corrections. I was going with the idea that a user will want to know the filtered result they have based on the URL(mostly when shared) but I see how that complicates implementation. I'll work on these corrections.

julienw: The idea was that in cellsConfiguratipn we'd have another property `urlValue`, as an array with the values we'd use for the URL for each of the values. The rationale is that it would be good to decorrelate the 2 uses:
1. one use is to display a value to the user => `possibleValues` (maybe it could be renamed to `displayedValues`? Not sure
2. one use is to add a value to the URL
3. and actually I can think of a 3rd use => the values we get from the server. Currently we use the same `possibleValues` array for this. But don't think about changing that in this PR, we can do that later.

So to decorrelate uses 1 and 2, we need 2 arrays for each of them.

Is that clearer?

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -107,10 +107,43 @@ export default function ResultsTable() {
...
+      if (!possibleValues) return;
+      const paramValue = rawSearchParams.get(`filter-${key}`);
+      initialFilters.set(key, new Set(paramValue?.split(',')));
+    });
+    setTableFilters(initialFilters);
+  };
+
+  useEffect(() => {
+    accessURlParams();
+  }, []);
```
julienw, line 134: Instead of using an effect, it would be better to use the 3rd version of useState (as described in the react documentation: https://react.dev/reference/react/useState#usestate).

This would set the initial value of `tableFilters` by using the URL Params directly.

By using an effect, instead, this impairs the performance. Indeed this works like this:
1. React renders the component (and the rest of the tree)
2. React runs the effect.
3. If the effect changes the state, React renders the component again... and the rest of the tree as well!

We want to avoid 2 and 3. By setting the right state right away at the first render time, only the first step is run, so this is much faster.

I'd like to stress that you need to use the 3rd version of `useState`, not the first or second. Why? Because we want the computation to happen just once. This is explained in this part of their documentation: https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state

src/components/CompareResults/ResultsTable.tsx
Netacci: Thanks for the clarification, this makes sense now.

src/components/CompareResults/ResultsTable.tsx
Netacci: @julienw Thank you for the detailed explanation! 

I’ve updated the code to use lazy initialization for tableFilters based on the URL Params as suggested. This way, the computation only happens once, and React renders the component only once during the initial render.

Thanks again for pointing this out, and for the useful documentation links!

Netacci: Hi @julienw I have implemented these changes and waiting for a re-review

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -103,14 +103,42 @@ export default function ResultsTable() {
...
+        rawSearchParams.set(`filter-${key}`, searchValue.join(','));
+      } else {
+        rawSearchParams.delete(`filter-${key}`);
+      }
+    });
+    updateRawSearchParams(rawSearchParams);
+  };
+
+  useEffect(() => {
+    if (tableFilters.size > 0) {
```
julienw, line 131: This `if` condition shouldn't be necessary: indeed you also want to update the URL when the value changes and is empty :-)

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -103,14 +103,42 @@ export default function ResultsTable() {
...
+      accessFilterValues();
+    }
+  }, [tableFilters]);
 
   const onClearFilter = (columnId: string) => {
     setTableFilters((oldFilters) => {
       const newFilters = new Map(oldFilters);
       newFilters.delete(columnId);
+      rawSearchParams.delete(`filter-${columnId}`);
+      updateRawSearchParams(rawSearchParams);
```
julienw, line 141: This should be unnecessary with the effect above especially if you remove the if.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -103,14 +103,42 @@ export default function ResultsTable() {
...
   const [frameworkIdVal, setFrameworkIdVal] = useState(frameworkId);
-  const [tableFilters, setTableFilters] = useState(
-    new Map() as Map<string, Set<string>>, // ColumnID -> Set<Values to remove>
-  );
+  const [tableFilters, setTableFilters] = useState(() => {
+    const initialFilters = new Map() as Map<string, Set<string>>;
+    cellsConfiguration.forEach(({ key, possibleValues }) => {
+      if (!possibleValues) return;
+      const paramValue = rawSearchParams.get(`filter-${key}`);
+      initialFilters.set(key, new Set(paramValue?.split(',')));
```
julienw, line 111: it would be good to not set anything if paramValue is not present

julienw (commented):
Thanks, this looks pretty good!

I just noticed a few small things that could be improved, and also a test for the functionality is missing.

That's OK to not fix this PR, please focus on your new task, but I still wanted to give you this last feedback.

Carla-Moz:
@Netacci Hi I'd like to let you know you've been selected to go on the next round! Congrats!

src/components/CompareResults/ResultsTable.tsx
Netacci: Yes, I see what you mean. I'll update this

Netacci:
> Thanks, this looks pretty good!
> 
> I just noticed a few small things that could be improved, and also a test for the functionality is missing.
> 
> That's OK to not fix this PR, please focus on your new task, but I still wanted to give you this last feedback.

Okay thanks for this.

julienw:
Hey! Thanks again for your work :)

In case you're interested, this feature has been implemented in #808, #809 and #810.
I'm closing this PR now!

Netacci:
> Hey! Thanks again for your work :)
> 
> In case you're interested, this feature has been implemented in #808, #809 and #810.
> I'm closing this PR now!

Thanks, I'll check them out.

PR #786, opened 2024-10-12
Persist Filters in URL to Maintain View After Page Refresh

Dorcas-BD:
This pull request fixes [issue (#765)](https://github.com/mozilla/perfcompare/issues/765) where the filters applied to the results table were not persisted in the URL, resulting in the default view being restored upon page refresh. With this update, users can now expect the following behaviour:

- **Filters Persisted:** When filters for the platform, status, or confidence are applied, they are retained even after the page is refreshed.
- **URL Updates Dynamically:** The URL updates as filters are applied or removed, allowing users to share specific filtered views with others.
- **Consistent User Experience:** Users will see the same filtered results when navigating back to the results table or when sharing the filtered URL.

## Implementation Details

The following changes were made to implement the persistence of filters:

1. **Effect Hook for Initial Filters:**
   - The `useEffect` hook initialises the filters based on the current URL search parameters. If no parameters are found, it defaults to using the possible values for each filter.
   - The `initialFilters` state is set accordingly to reflect the current filter configuration.

2. **Filter Clearing Functionality:**
   - The `onClearFilter` function was updated to ensure that when a filter is cleared, the URL is also updated to reflect all possible values.

3. **Filter Toggling Functionality:**
   - The `onToggleFilter` function ensures that when filters are toggled, the URL parameters are updated accordingly to maintain consistency with the displayed results.

### Before Changes
beforeee

#### Platform
platformmm

#### Status
statussss

#### Confidence
confidenceee

## Steps to Reproduce

1. Run the application
2. Compare with a base
3. Apply filters to the columns (e.g., show only improvements, regressions, by platform, or filter by confidence or platform).
4. Refresh the page.
5. Observe that the column filters are now persisted, maintaining the expected filtered view.
6.  Note that the URL should update as filters are applied or removed.

### DIFF
--- a/src/components/CompareResults/ResultsTable.tsx
+++ b/src/components/CompareResults/ResultsTable.tsx
@@ -116,7 +116,11 @@ export default function ResultsTable() {
       const paramsValue = rawSearchParams.get(key) ?? possibleValues.join(',');
       initialFilters.set(
         key,
-        new Set(possibleValues.filter((v) => !paramsValue.includes(v))),
+        new Set(
+          possibleValues.filter(
+            (possibleValue) => !paramsValue.includes(possibleValue),
+          ),
+        ),
       );
 
       if (!rawSearchParams.has(key)) rawSearchParams.set(key, paramsValue);
@@ -128,7 +132,7 @@ export default function ResultsTable() {
 
   const onClearFilter = (columnId: string) => {
     const possibleValues = cellsConfiguration.find(
-      (c) => c.key === columnId,
+      (cell) => cell.key === columnId,
     )?.possibleValues;
 
     if (!possibleValues) return;
@@ -140,7 +144,7 @@ export default function ResultsTable() {
 
   const onToggleFilter = (columnId: string, filters: Set<string>) => {
     const possibleValues = cellsConfiguration.find(
-      (c) => c.key === columnId,
+      (cell) => cell.key === columnId,
     )?.possibleValues;
 
     if (!possibleValues) return;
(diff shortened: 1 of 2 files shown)


src/components/CompareResults/ResultsTable.tsx
```diff
@@ -107,20 +107,52 @@ export default function ResultsTable() {
...
+    if (!possibleValues) return;
+
+    const paramsValue =
+      filters.size === possibleValues.length
+        ? ''
+        : possibleValues.filter((v) => !filters.has(v)).join(',');
+
+    setTableFilters((prev) => new Map(prev).set(columnId, filters));
+    rawSearchParams.set(columnId, paramsValue);
+    updateRawSearchParams(rawSearchParams);
```
Carla-Moz, line 155: This looks like a great solution. Good job on your first attempt and thanks for the detailed PR description! However, would you mind using shorthand or whole words instead of initials ? For example : `cell vs `c` etc? Otherwise, I think this would be a good PR to be merged.

Carla-Moz (changes requested):

src/components/CompareResults/ResultsTable.tsx
Dorcas-BD: Okay, I will do that. Thank you!

Dorcas-BD:
Hello @Carla-Moz, I’ve updated the code to use whole words instead of initials as suggested. Please take a look at the changes!

Dorcas-BD:
Hi @Carla-Moz,

I noticed that the 'Changes Requested' label was added to my PR. Could you please clarify what specific changes are needed?

Thank you!

Carla-Moz:
@Dorcas-BD Did you see my comment above about the changes?

Also, I'd like to let you know you've been selected to go on the next round! Congrats!

Dorcas-BD:
No, @Carla-Moz 

I only saw the initial comment, made the changes, and got the "Resolved comment/Re-review" label. But I didn’t see any further comments when the second "Changes Requested" label was added.

The initial comment I received was:
"This looks like a great solution. Good job on your first attempt and thanks for the detailed PR description..."

Thanks so much for the great news about moving to the next round! I'll check out the details on the channel.

Dorcas-BD:
Thank you for the update!😊 @julienw 

I'm glad to hear the feature has been implemented. I appreciate the chance to contribute and look forward to future tasks!

PR #792, opened 2024-10-23
PCF-296 Remove interval param when comparing with a base

beatrice-acasandrei:


src/types/state.ts
```diff
@@ -98,7 +98,6 @@ export type CompareResultsItem = {
   delta_percentage: number;
   magnitude: number;
   is_improvement: boolean;
-  is_empty: boolean;
```
beatrice-acasandrei, line 101: I removed this field from the response in https://github.com/mozilla/treeherder/pull/8216

julienw (approved):
looks good to me, thanks!

PR #793, opened 2024-10-23
Add more options to the filtering menus, and add a "no value" filter for the confidence column

julienw:
This PR does a few things:
Commit 1 adds a "no value" filter for the confidence column. It was driving me crazy.
Commit 2 adds a few options to all filtering menus: it adds an entry "show only XXX" for every possible values. I find it might be easier to use for our users.

[=> Deploy preview <=](https://deploy-preview-793--mozilla-perfcompare.netlify.app/compare-results?baseRev=495128bbc3780b459e1f76fe004c96df81765383&baseRepo=mozilla-central&newRev=75d81c21a6ae5a09b6516aa57f4cd3da0c2d982e&newRepo=mozilla-central&newRev=fa1fe4195c4c98ed4541164cae171c52f30c58fe&newRepo=mozilla-central&framework=1)

image


esanuandra:
I'm not sure if only changing the icon color of the column is enough to suggest these results are filtered.

julienw:
> I'm not sure if only changing the icon color of the column is enough to suggest these results are filtered.

Yeah, Dasha suggested another design (basically showing the number of checked filters) but I never got to it. If that's not done yet, I'll file another JIRA ticket about that with the screenshots so that we don't lose them (they're on slack currently).
But the color was there before already, this was in my first impelmentation, so I don't really want to change that in this PR...

PR #798, opened 2024-11-06
PCF-524 Display the mean instead of the median in the results table rows + update the column title + display median in the expanded row as well

esanuandra:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/PCF-524)

[localhost link](http://localhost:3000/compare-results?baseRev=a635fc061556de8501069d83b02b083892ab2802&baseRepo=mozilla-central&newRev=08a85f1b7545a8bb7ed6e720fd78705f4f17f570&newRepo=mozilla-central&newRev=c9170b504cdce7fd51d1934bad32709bea49b112&newRepo=mozilla-central&framework=1)
[deploy link](https://deploy-preview-798--mozilla-perfcompare.netlify.app/compare-results?baseRev=a635fc061556de8501069d83b02b083892ab2802&baseRepo=mozilla-central&newRev=08a85f1b7545a8bb7ed6e720fd78705f4f17f570&newRepo=mozilla-central&newRev=c9170b504cdce7fd51d1934bad32709bea49b112&newRepo=mozilla-central&framework=1)


julienw:
Hey @esanuandra, while you're here, I think it would be good to handle https://mozilla-hub.atlassian.net/browse/PCF-459 as well. What do you think?

src/components/CompareResults/RevisionRowExpandable.tsx
```diff
@@ -79,14 +83,18 @@ function RevisionRowExpandable(props: RevisionRowExpandableProps) {
...
-        <div className={`${styles.bottomSpace}`}>
-          <b>Mean Difference</b>: {deltaPercent}%{' '}
+        <div>
+          <b>Difference of means</b>: {deltaPercent}%{' '}
           {newIsBetter ? 'better' : 'worse'} ({delta} {deltaUnit})
         </div>
+        <div>
+          <b>Difference of medians</b>: {(newMedian - baseMedian).toFixed(2)} (
+          {(((newMedian - baseMedian) / baseMedian) * 100).toFixed(2)}%)
+        </div>
```
julienw, line 93: It's a bit weird when one of the values is zero, so I'd do something like this:
```js
{newMedian && baseMedian ? (
  <div>
     ...
  </div>
) : null}
```

julienw: It would be good to have the percentage value first and the absolute value next, like for the means, otherwise it's difficult to read. And also to have the unit for the absolute value like for the means.

It would be easier to do the computation before the JSX part so that you just have to use the values here.

It would also be good to compute if it's worse/better using the `lower_is_better` property, again to do exactly like the difference of the means, for an easier read.

src/components/CompareResults/RunValues.tsx
```diff
@@ -63,6 +64,9 @@ function RunValues(props: RunValuesProps) {
             </div>
           ))}
         </div>
+        <div>
+          <b>Mean</b>: {avg} <b>Median</b>: {median}
```
julienw, line 68: nits: add units, also add a comma after the mean value to better visually separate the 2 values.

julienw (changes requested):
Thanks, here are a few more comments to make it easier to read for the user!

julienw (commented):
I find that the use of `\u00a0` not easy to read here. Maybe you want [`white-space: nowrap;`](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) on the divs instead?

I like that you tried to add the "lower is better" information but I think we can do it in a more explicit way. I left a comment about it, tell me what you think!

We also need to handle better the cases where either new and base is 0 because it didn't run, but I think it's out of scope for this PR.

julienw (approved):
Thanks for your patience, I think this PR brings a lot of value to our users!
Let's land it now !

PR #799, opened 2024-11-12
Add Tests for SubtestsResultsView Component

sumairq:
This PR introduces a set of tests for the `SubtestsResultsView` component to validate its rendering and key functionalities. The main areas tested include:

- **Rendering and Snapshot**: Verifies that the component renders correctly and matches the expected snapshot.
- **Main Content Display**: Confirms that the main content in the view is present.
- **Breadcrumbs Navigation**: Ensures breadcrumbs are displayed for easier navigation within the component.
- **Subtests Results Table**: Checks that the subtests results table is rendered in the view.
- **Retrigger Button Functionality**: Tests the retrigger functionality, confirming the correct alerts and authorization code requests are triggered.
- **Download JSON Button**: Validates that clicking the "Download JSON" button correctly triggers the creation of a blob URL.

Since the `SubtestsOverTimeResultsView` is identical to `SubtestsResultsView` only snapshot testing has been done for `SubtestsOverTimeResultsView`


src/components/CompareResults/SubtestsResults/SubtestsOverTimeResultsView.tsx
```diff
@@ -26,7 +26,10 @@ function SubtestsResultsView(props: SubtestsResultsViewProps) {
   }, [title]);
 
   return (
-    <div className={styles.container}>
+    <div
+      data-testid='beta-version-subtests-compare-over-time-results'
```
julienw, line 30: see above, this is probably not necessary, you can revert this change.

julienw (changes requested):
Hey Sumair,
First thanks a ton for these new tests, this will be extremely useful!
I left a few comments to make them more solid and faster to run as well.
I noticed you were using the test ids a lot, but in reality they should be used as a last ressort. See the "Priority" paragraph at https://testing-library.com/docs/queries/about/#priority for more information about that.
Otherwise after these small changes it should be ready to go!

src/__tests__/CompareResults/SubtestsResultsView.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+        '?baseRev=d775409d7c6abb76362a3430e9880ec032ad4679&baseRepo=mozilla-central&newRev=22f4cf67e8ad76b5ab2a00b97837d1d920b8c2b7&newRepo=mozilla-central&framework=1&baseParentSignature=4769486&newParentSignature=4769486',
+      loader,
+    });
+
+    const table = await screen.findByTestId('subtests-results-table');
+
+    expect(table).toBeInTheDocument();
+  });
+
+  it('should request authorization code when "Retrigger" button is clicked', async () => {
```
julienw, line 115: Isn't this already tested in Retrigger.test.jsx?

sumairq: No, it's not tested in `Retrigger.test.jsx`. However, it is covered in `ResultsView.test.tsx`, where the retrigger button for the results view is tested.
I have tested the retrigger button just the same but for subtests view in `SubtestsResultsView.test.tsx`

julienw (commented):
thanks, but I think you forgot to handle the review comments fully :) can you have another look please?

julienw (approved):
looks good to me now, we can merge once the snapshots are updated and the tests are green!

PR #803, opened 2024-11-19
PCF-549 Add the name of the device used instead of just "Android"

esanuandra:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/PCF-549)

[localhost test link](http://localhost:3000/compare-results?baseRev=b6fe6098a8bbf74edca23103c6756be13d238378&baseRepo=mozilla-central&newRev=b136e0dc43e39a2da4f7d375ef545d230caa86c0&newRepo=mozilla-central&newRev=9358b6a02a04f62e0c9c7fcba4877700076a28f5&newRepo=mozilla-central&framework=13&search=speedometer)

[deploy preview test link](https://deploy-preview-803--mozilla-perfcompare.netlify.app/compare-results?baseRev=b6fe6098a8bbf74edca23103c6756be13d238378&baseRepo=mozilla-central&newRev=b136e0dc43e39a2da4f7d375ef545d230caa86c0&newRepo=mozilla-central&newRev=9358b6a02a04f62e0c9c7fcba4877700076a28f5&newRepo=mozilla-central&framework=13&search=speedometer)

Adding android device name next to platform name.

Screenshot 2024-11-19 at 18 01 40

### DIFF
--- a/src/components/CompareResults/RevisionRow.tsx
+++ b/src/components/CompareResults/RevisionRow.tsx
@@ -18,7 +18,10 @@ import { useAppSelector } from '../../hooks/app';
 import { Strings } from '../../resources/Strings';
 import { Colors, Spacing, ExpandableRowStyles } from '../../styles';
 import type { CompareResultsItem, PlatformShortName } from '../../types/state';
-import { getPlatformShortName, getPlatformVersion } from '../../utils/platform';
+import {
+  getPlatformShortName,
+  getPlatformAndVersion,
+} from '../../utils/platform';
 import AndroidIcon from '../Shared/Icons/AndroidIcon';
 import LinuxIcon from '../Shared/Icons/LinuxIcon';
 import SubtestsIcon from '../Shared/Icons/SubtestsIcon';
@@ -335,7 +338,6 @@ function RevisionRow(props: RevisionRowProps) {
 
   const platformShortName = getPlatformShortName(platform);
   const platformIcon = platformIcons[platformShortName];
-  const androidPlatform: PlatformShortName = 'Android';
 
   const [expanded, setExpanded] = useState(false);
 
@@ -364,10 +366,7 @@ function RevisionRow(props: RevisionRowProps) {
           <Tooltip placement='bottom' title={platform} arrow>
             <div className='platform-container'>
               {platformIcon}
-              <span>{platformShortName}</span>
-              {platformShortName === androidPlatform && (
-                <span>&nbsp;{getPlatformVersion(platform)}</span>
-              )}
+              <span>{getPlatformAndVersion(platform)}</span>
             </div>
           </Tooltip>
         </div>
--- a/src/components/CompareResults/RevisionRow.tsx
+++ b/src/components/CompareResults/RevisionRow.tsx
@@ -18,7 +18,7 @@ import { useAppSelector } from '../../hooks/app';
 import { Strings } from '../../resources/Strings';
 import { Colors, Spacing, ExpandableRowStyles } from '../../styles';
 import type { CompareResultsItem, PlatformShortName } from '../../types/state';
-import { getPlatformShortName } from '../../utils/platform';
+import { getPlatformShortName, getPlatformVersion } from '../../utils/platform';
 import AndroidIcon from '../Shared/Icons/AndroidIcon';
 import LinuxIcon from '../Shared/Icons/LinuxIcon';
 import SubtestsIcon from '../Shared/Icons/SubtestsIcon';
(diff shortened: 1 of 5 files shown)


src/components/CompareResults/RevisionRow.tsx
```diff
@@ -364,6 +365,12 @@ function RevisionRow(props: RevisionRowProps) {
             <div className='platform-container'>
               {platformIcon}
               <span>{platformShortName}</span>
+              {platformShortName === androidPlatform && (
+                <span>
+                  &nbsp;
+                  {platform.split('-')[2]}
+                </span>
+              )}
```
julienw, line 373: It would be better to move that to `getPlatformShortName`.
The rationale is that we also want to add the platform version (especially MacOSX and Windows) and this will be even more checks, so it's better to have this well self contained in one location. Moreover there's already a block for Android there. And you'll be able to add unit tests for that ;-)

julienw: Also when you move the code to `getPlatformShortName`, please provide one example of how the platform looks like close to this `split`, in a comment.

julienw (changes requested):
This looks good to me, but it would be better to have it in `getPlatformShortName`, with some additional unit tests for it because I feel like this is a can of worms.

Also is the name of the platform consistent enough that we can do a split like that? I'm afraid we might have issues, but maybe we can hardcode some of the non consistent one in the future...

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -364,6 +365,9 @@ function RevisionRow(props: RevisionRowProps) {
             <div className='platform-container'>
               {platformIcon}
               <span>{platformShortName}</span>
+              {platformShortName === androidPlatform && (
+                <span>&nbsp;{getPlatformVersion(platform)}</span>
+              )}
```
julienw, line 370: make a `getPlatformAndVersion` that calls `getPlatformShortName` and return the full string

use `\u00a0` inside the string as the non-breaking space.

src/utils/platform.ts
```diff
@@ -13,3 +13,11 @@ export const getPlatformShortName = (
   if (platformName.toLowerCase().includes('android')) return 'Android';
   return 'Unspecified';
 };
+
+export const getPlatformVersion = (platformName: string): string | null => {
+  // Example of adroid platform
```
julienw, line 18: typo

src/components/CompareResults/RevisionRow.tsx
```diff
@@ -364,6 +365,9 @@ function RevisionRow(props: RevisionRowProps) {
             <div className='platform-container'>
               {platformIcon}
               <span>{platformShortName}</span>
+              {platformShortName === androidPlatform && (
```
julienw, line 368: the condition isn't needed :-)

julienw (commented):
Just putting up the notes we discussed during the meeting :-)
looks good otherwise!

src/utils/platform.ts
```diff
@@ -13,3 +13,13 @@ export const getPlatformShortName = (
   if (platformName.toLowerCase().includes('android')) return 'Android';
   return 'Unspecified';
 };
+
+export const getPlatformAndVersion = (platform: string): string => {
+  const platformShortName = getPlatformShortName(platform);
+  // Example of android platform, the split is specific for this platform
+  // Other platforms are built differently
+  // android-hw-p6-13-0-android-aarch64-shippable-qr
+  if (platformShortName === 'Android')
```
julienw, line 22: I feel like we need to check if platform with `android-hw-` to know if that's the right format. Otherwise we break when the format isn't what we expect (for example select "build metrics" with your deploy preview link).

This will avoid the issues with our tests too.

```suggestion
  if (platform.startsWith('android-hw-'))
```

src/utils/platform.ts
```diff
@@ -13,3 +13,13 @@ export const getPlatformShortName = (
   if (platformName.toLowerCase().includes('android')) return 'Android';
   return 'Unspecified';
 };
+
+export const getPlatformAndVersion = (platform: string): string => {
```
julienw, line 17: can you please add a small unit test testing just that function? It would be in `src/__tests__/utils/platform.ts`.
No need to do anything comprehensive, but it would be good to test one android platform string that works (`android-hw-p6-13-0-android-aarch64-shippable-qr`), one that doesn't (`android-5-0-aarch64-release`), and probably one from another OS such as `windows11-64-shippable-qr`.

src/utils/platform.ts
esanuandra: In the tests there is an example `'android-em-7-0-x86_64-lite-qr'`.
The check for `'android-hw-'` breaks for this.

julienw: -em- is for emulator so I believe it's irrelevant to your change, so on the contrary it does what we want :-)

julienw (approved):
thanks!

PR #805, opened 2024-11-20
PCF-501 Search by email should  be using substrings

beatrice-acasandrei:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/PCF-501)

### DIFF
--- a/src/__tests__/Search/SearchView.test.tsx
+++ b/src/__tests__/Search/SearchView.test.tsx
@@ -241,7 +241,7 @@ describe('Base and OverTime Search', () => {
 
     expect(
       await screen.findByText(
-        'The search input must be a minimum of three characters in length.',
+        'The search input must be at least three characters.',
       ),
     ).toBeInTheDocument();
 
@@ -271,15 +271,13 @@ describe('Base and OverTime Search', () => {
     await user.type(searchInput, 'jo');
     // No error appears while the user type.
     expect(
-      screen.queryByText(
-        'The search input must be a minimum of three characters in length.',
-      ),
+      screen.queryByText('The search input must be at least three characters.'),
     ).not.toBeInTheDocument();
 
     // But this appears after a while.
     expect(
       await screen.findByText(
-        'The search input must be a minimum of three characters in length.',
+        'The search input must be at least three characters.',
       ),
     ).toBeInTheDocument();
     await user.type(searchInput, 'hncleese');
--- a/src/__tests__/Search/SearchView.test.tsx
+++ b/src/__tests__/Search/SearchView.test.tsx
@@ -22,9 +22,9 @@ function setupTestData() {
   const { testData } = getTestData();
   (global.fetch as FetchMockSandbox)
     .get(
-      'begin:https://treeherder.mozilla.org/api/project/try/push/?author=',
+      'begin:https://treeherder.mozilla.org/api/project/try/push/?author_contains=',
       (url) => {
-        const author = new URL(url).searchParams.get('author');
+        const author = new URL(url).searchParams.get('author_contains');
         return { results: testData.filter((item) => item.author === author) };
       },
     )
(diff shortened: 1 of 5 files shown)


src/components/Search/SearchInputAndResults.tsx
```diff
@@ -68,9 +67,7 @@ export default function SearchInputAndResults({
 
   const searchRecentRevisions = useCallback(
     async (searchTerm: string) => {
-      const emailMatch = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
-      const longHashMatch = /\b[a-f0-9]{40}\b/;
-      const shortHashMatch = /\b[a-f0-9]{12}\b/;
+      const authorMatch = /[^0-9a-fA-F]/;
```
Carla-Moz, line 70: I wonder if `authorInfoMatch`would be a better name?

Carla-Moz (commented):
Thanks, Beatrice! Also, please add a brief description about the purpose of your PR next time so reviewers can get better context ( ex: jira or the bug link).

src/__tests__/Search/SearchView.test.tsx
```diff
@@ -219,108 +218,6 @@ describe('Base and OverTime Search', () => {
     expect(comment[0]).not.toBeInTheDocument();
   });
 
-  it('Should not call fetch if search value is not a hash or email', async () => {
```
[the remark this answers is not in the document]
julienw: agreed!

src/__tests__/Search/SearchView.test.tsx
```diff
@@ -219,108 +218,6 @@ describe('Base and OverTime Search', () => {
...
-    // fetch is called 6 times:
-    // - 3 times on initial load: one for each input, that is 2 in "compare with
-    //   base", 1 in "compare over time"
-    // - 3 times from the user interaction: 1 time for each "clear", because the
-    //   other user interactons are invalid and therefore don't trigger any
-    //   fetches (this is the goal for this test).
-    expect(global.fetch).toHaveBeenCalledTimes(6);
-  });
-
-  it('Should debounce user interaction', async () => {
```
julienw, line 267: Mmm I think this one should stay

src/components/Search/SearchInputAndResults.tsx
```diff
@@ -68,9 +67,7 @@ export default function SearchInputAndResults({
 
   const searchRecentRevisions = useCallback(
     async (searchTerm: string) => {
-      const emailMatch = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
-      const longHashMatch = /\b[a-f0-9]{40}\b/;
-      const shortHashMatch = /\b[a-f0-9]{12}\b/;
+      const authorInfoMatch = /[^0-9a-fA-F]/;
```
julienw, line 70: nit: please add a comment explaining the reasoning, something such as:
```
// If the searchTerm doesn't look like a hash, then we assume it might be an author.
// If it looks like a hash, then we assume it's a hash.
// NOTE: In the future we might want to be more clever and request both
// endpoints even when it looks like a hash. For example a request such as "ade" could
// return results for an author named "adenot" _and_  results for a hash "ade821ac".
// In that case it would be better to show the results for the author.
```

julienw (approved):
love it, thanks!

src/resources/Strings.tsx
```diff
@@ -147,4 +147,8 @@ export const Strings = {
       },
     },
   },
+  errors: {
+    warningText:
+      'The search input must be a minimum of three characters in length.',
```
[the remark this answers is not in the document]
julienw: Hey Beatrice, ChatGPT suggested me these 2 possible sentences:

    "The search input must be at least three characters."
    "The search input must contain at least three characters."

Especially "in length" seems extraneous :-)

After this change I'm good about the other change too.

src/components/Search/SearchInputAndResults.tsx
```diff
@@ -82,12 +88,9 @@ export default function SearchInputAndResults({
...
       if (!searchTerm) {
         searchParameters = { repository };
-      } else if (emailMatch.test(searchTerm)) {
+      } else if (authorInfoMatch.test(searchTerm)) {
         searchParameters = { repository, author: searchTerm };
-      } else if (
-        longHashMatch.test(searchTerm) ||
-        shortHashMatch.test(searchTerm)
-      ) {
+      } else if (minimumCharacterLength.test(searchTerm)) {
```
julienw, line 93: as we discussed, use `searchTerm.length < 3` as a first if condition to set the error, as it's more readable, and then use the regexp as it was before.

PR #806, opened 2024-11-21
Added the checked item count to the filter header name

Carla-Moz:
This PR updates the styles of the filter header to match the figma designs.

Screenshot 2024-11-20 at 19 11 14

This is what it looks like now:

Screenshot 2024-11-20 at 19 11 54

Need to add tests for this fix. Will create a separate ticket.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10051,9 +10051,9 @@
       }
     },
     "node_modules/http-proxy-middleware": {
-      "version": "2.0.6",
-      "resolved": "https://registry.npmjs.org/http-proxy-middleware/-/http-proxy-middleware-2.0.6.tgz",
-      "integrity": "sha512-ya/UeJ6HVBYxrgYotAZo1KvPWlgB48kUJLDePFeneHsVujFaW5WNj2NgWCAE//B1Dl02BIfYlpNgBy8Kf8Rjmw==",
+      "version": "2.0.7",
+      "resolved": "https://registry.npmjs.org/http-proxy-middleware/-/http-proxy-middleware-2.0.7.tgz",
+      "integrity": "sha512-fgVY8AV7qU7z/MmXJ/rxwbrtQH4jBQ9m7kp3llF0liB7glmFeVZFBepQb32T3y8n8k2+AEYuMPCpinYW+/CuRA==",
       "dev": true,
       "dependencies": {
         "@types/http-proxy": "^1.17.8",
@@ -13758,12 +13758,12 @@
       "integrity": "sha512-+HzcV2H+rbSJzApgkj0NdTakkC+bnyeiUxgT6/m7mjcz1CmM22KYFKp+EVj1sWe4UYcnriJr5uqHQD/gMHLD+g=="
     },
     "node_modules/micromatch": {
-      "version": "4.0.5",
-      "resolved": "https://registry.npmjs.org/micromatch/-/micromatch-4.0.5.tgz",
-      "integrity": "sha512-DMy+ERcEW2q8Z2Po+WNXuw3c5YaUSFjAO5GsJqfEl7UjvtIuFKO6ZrKvcItdy98dwFI2N1tg3zNIdKaQT+aNdA==",
+      "version": "4.0.8",
+      "resolved": "https://registry.npmjs.org/micromatch/-/micromatch-4.0.8.tgz",
+      "integrity": "sha512-PXwfBhYu0hBCPw8Dn0E+WDYb7af3dSLVWKi3HGv84IdF4TyFoC0ysxFd0Goxw7nSv4T/PzEJQxsYsEiFCKo2BA==",
       "dev": true,
       "dependencies": {
-        "braces": "^3.0.2",
+        "braces": "^3.0.3",
         "picomatch": "^2.3.1"
       },
       "engines": {
@@ -25343,9 +25343,9 @@
       }
     },
     "http-proxy-middleware": {
-      "version": "2.0.6",
-      "resolved": "https://registry.npmjs.org/http-proxy-middleware/-/http-proxy-middleware-2.0.6.tgz",
-      "integrity": "sha512-ya/UeJ6HVBYxrgYotAZo1KvPWlgB48kUJLDePFeneHsVujFaW5WNj2NgWCAE//B1Dl02BIfYlpNgBy8Kf8Rjmw==",
+      "version": "2.0.7",
+      "resolved": "https://registry.npmjs.org/http-proxy-middleware/-/http-proxy-middleware-2.0.7.tgz",
+      "integrity": "sha512-fgVY8AV7qU7z/MmXJ/rxwbrtQH4jBQ9m7kp3llF0liB7glmFeVZFBepQb32T3y8n8k2+AEYuMPCpinYW+/CuRA==",
       "dev": true,
       "requires": {
         "@types/http-proxy": "^1.17.8",
@@ -28037,12 +28037,12 @@
       "integrity": "sha512-+HzcV2H+rbSJzApgkj0NdTakkC+bnyeiUxgT6/m7mjcz1CmM22KYFKp+EVj1sWe4UYcnriJr5uqHQD/gMHLD+g=="
     },
     "micromatch": {
-      "version": "4.0.5",
-      "resolved": "https://registry.npmjs.org/micromatch/-/micromatch-4.0.5.tgz",
-      "integrity": "sha512-DMy+ERcEW2q8Z2Po+WNXuw3c5YaUSFjAO5GsJqfEl7UjvtIuFKO6ZrKvcItdy98dwFI2N1tg3zNIdKaQT+aNdA==",
+      "version": "4.0.8",
+      "resolved": "https://registry.npmjs.org/micromatch/-/micromatch-4.0.8.tgz",
+      "integrity": "sha512-PXwfBhYu0hBCPw8Dn0E+WDYb7af3dSLVWKi3HGv84IdF4TyFoC0ysxFd0Goxw7nSv4T/PzEJQxsYsEiFCKo2BA==",
       "dev": true,
       "requires": {
-        "braces": "^3.0.2",
+        "braces": "^3.0.3",
         "picomatch": "^2.3.1"
       }
     },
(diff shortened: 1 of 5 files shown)


julienw:
I wonder if we should make it look like more the mockup? Especially removing the icon, and add the blue color to the `(n)` text. What do you think @Carla-Moz ?

Carla-Moz:
> I wonder if we should make it look like more the mockup? Especially removing the icon, and add the blue color to the `(n)` text. What do you think @Carla-Moz ?

I thought about that but the table headers titles clash with the dropdown styles so I made the decision to combine them. Below is a figma image of the table header columns with the filter icon, name, and carrot down.

Screenshot 2024-11-21 at 08 18 46

Screenshot 2024-11-21 at 08 20 45

We'd have to repeat the "status" name if we followed the designs strictly. Dasha didn't show us a dropdown design for the table header filters.

Carla-Moz:
However, I can make the (n) blue or yellow re light or dark themes.

julienw:
Yes I was talking only about making the "(2)" status blue, and removing the icon, not to do what's inside the dropdown (I totally understand these ones are hard to change at the moment)!

julienw:
In short do your best, but don't overdo it; I just thought it could be good to remove the icon and make the number colored instead, but don't need to make it exactly the same (especially regarding to alignments etc).

package-lock.json
julienw, line 1: I believe this change was mistakenly added?

src/components/CompareResults/TableHeader.tsx
```diff
@@ -32,7 +33,20 @@ function FilterableColumn({
...
+  //this returns number of possible values that are checked
+  const possibleCheckedValues = possibleValues.reduce(
+    (count: number, value) => {
+      if (!uncheckedValues?.has(value)) {
+        count++;
+      }
+      return count;
+    },
+    0,
+  );
```
julienw, line 49: dosn't this work? `possibleValues.length - uncheckedValues.size`

src/components/CompareResults/TableHeader.tsx
```diff
@@ -67,19 +81,22 @@ function FilterableColumn({
...
           fontSize='small'
           color={hasFilteredValues ? 'primary' : 'inherit'}
-          sx={{ marginInlineStart: 1 }}
+          sx={{ marginInlineStart: 1, marginRight: 0.5, marginLeft: 0 }}
           titleAccess={
             hasFilteredValues ? 'Some filters are active' : 'No active filters'
           }
         />
+        {name}
+        <div className={filterCountStyle}>({possibleCheckedValues})</div>
```
julienw, line 98: If you use Box instead of div, you could use `sx` here

package-lock.json
Carla-Moz: Yeah, I can amend and remove that.

src/components/CompareResults/TableHeader.tsx
Carla-Moz: I don't mind doing that. I suppose we should transition out of `typestyle` soon for future styles development? At least have a consensus surrounding the topic. Maybe lets bring it up in next sync when we're all here.

src/components/CompareResults/TableHeader.tsx
Carla-Moz: Yes, that makes sense but `uncheckedValues.size` is possible undefined so I went the longest route. My brain likes the longest route too much lol. 

However this works:
```
  const possibleCheckedValues =
    possibleValues.length - (uncheckedValues?.size || 0);
```

src/components/CompareResults/TableHeader.tsx
julienw: here I was more concerned about the consistency in this same component :)

but otherwise yeah we should make a decision and write it down

src/components/CompareResults/TableHeader.tsx
julienw: Also I kind of would like to reverse the meaning of what's in the set now... I think it was a bad idea of mine to have in the set what's been deselected as opposed to what's been selected (I think it was easier to implement this way at the time).

We can do it afterwards.

src/components/CompareResults/TableHeader.tsx
Carla-Moz: I understand. Sounds good!

src/components/CompareResults/TableHeader.tsx
Carla-Moz: Yes, I was initially confused by the unchecked values vs checked but how it is now makes sense because by default all the values are checked vs an opposite user flow where by default the values are unchecked.

Carla-Moz: @julienw But now I'm questioning why they're all checked by default. It's like all the filters _are on_ which is an uncommon user flow, no? Usually, by default, the filter is unchecked, then you check it, and it filters the table. 

Also, why do we have two options in the dropdown that do the same thing? This needs revisiting. 

Example: "Select only Windows"  and the filter option where I've only selected Windows. 

Screenshot 2024-11-22 at 11 24 44

Carla-Moz: Also, we might need to add checkboxes here too like in the original figma design. That might make it less confusing. I'll file a ticket.

image

julienw: > Also, why do we have two options in the dropdown that do the same thing? This needs revisiting.
> Example: "Select only Windows" and the filter option where I've only selected Windows.

Not the same thing :-)

"Select only Windows" will deselect everything else where needed.
But clicking just "Windows" will keep the other options.

src/components/CompareResults/TableHeader.tsx
```diff
@@ -72,19 +77,29 @@ function FilterableColumn({
...
           fontSize='small'
           color={hasFilteredValues ? 'primary' : 'inherit'}
-          sx={{ marginInlineStart: 1 }}
+          sx={{ marginInlineStart: 1, marginRight: 0.5, marginLeft: 0 }}
           titleAccess={
             hasFilteredValues ? 'Some filters are active' : 'No active filters'
           }
         />
+        {name}
+        <Box
```
julienw, line 93: Note: In the design we don't show the counter when no value are selected, but I think we should still show it in this case because this would filter out every values and give an empty list. So I'm good with this choice.

src/components/CompareResults/TableHeader.tsx
```diff
@@ -72,19 +77,29 @@ function FilterableColumn({
...
+          sx={{
+            paddingLeft: 0.5,
+            color: themeMode == 'light' ? Colors.LinkText : Colors.LinkTextDark,
+            fontWeight: 800,
+          }}
+        >
+          ({possibleCheckedValues})
+        </Box>
+        <KeyboardArrowDownIcon sx={{ marginLeft: 1 }} />
```
julienw, line 102: nit: use `marginInlineStart`

src/components/CompareResults/TableHeader.tsx
```diff
@@ -72,19 +77,29 @@ function FilterableColumn({
...
           titleAccess={
             hasFilteredValues ? 'Some filters are active' : 'No active filters'
           }
         />
+        {name}
+        <Box
+          sx={{
+            paddingLeft: 0.5,
+            color: themeMode == 'light' ? Colors.LinkText : Colors.LinkTextDark,
+            fontWeight: 800,
```
julienw, line 97: I'd remove the boldness for now. In Dasha's design the text is bold as well, but then all headers should be bold, and I think this is out of scope for this PR. When we'll add the boldness to all headers we can also add it here.

src/components/CompareResults/TableHeader.tsx
```diff
@@ -72,19 +77,29 @@ function FilterableColumn({
...
+          borderRadius: 0.5,
+          padding: '6px 12px',
         })}
       >
-        {name}
         <FilterListIcon
           fontSize='small'
           color={hasFilteredValues ? 'primary' : 'inherit'}
-          sx={{ marginInlineStart: 1 }}
+          sx={{ marginInlineStart: 1, marginRight: 0.5, marginLeft: 0 }}
```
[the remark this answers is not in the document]
Carla-Moz: I have not idea where this marginInlineStart came from. I need to do a better job of cleaning up my code before commiting. :/

src/components/CompareResults/TableHeader.tsx
```diff
@@ -72,19 +77,29 @@ function FilterableColumn({
...
           titleAccess={
             hasFilteredValues ? 'Some filters are active' : 'No active filters'
           }
         />
```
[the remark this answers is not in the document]
Carla-Moz: @julienw There seems to be some misunderstanding here. 

Dasha's design does include _the arrow down in the table header._ 

figma-design

My deployed url link design:
image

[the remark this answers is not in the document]
julienw: > Dasha's design does include the filter icon in the table header.

Not her latest design for this :-) see slack message from June 17 and the linked design: https://www.figma.com/design/LRV76asSVOULlsWbRNZYp0/PerfCompare?node-id=3431-43375&node-type=section&m=dev

Yes I agree it's confusing because all the other designs are including the icon...

src/components/CompareResults/TableHeader.tsx
```diff
@@ -72,19 +75,21 @@ function FilterableColumn({
...
-          color={hasFilteredValues ? 'primary' : 'inherit'}
-          sx={{ marginInlineStart: 1 }}
-          titleAccess={
-            hasFilteredValues ? 'Some filters are active' : 'No active filters'
-          }
-        />
+        <Box
+          sx={{
+            paddingInlineStart: 0.5,
+            color: hasFilteredValues ? 'primary' : 'inherit',
```
julienw, line 86: Turns out that just `primary` doesn't work, we need `primary.main`. In the `FilterListIcon` (that is `SvgIcon`) like before, there's some logic to do that, but with  `Box` we need to be explicit.

```suggestion
            color: hasFilteredValues ? 'primary.main' : 'inherit',
```

julienw: Also I wonder if we should just remove the condition to have it always blue:
```suggestion
            color: 'primary.main',
```

what do you think?

julienw (approved):
Thanks, this looks good to me! Only the blue color doesn't work but I've put the solution in a comment, also with an additional question.

With this fixed this is good to go!

src/components/CompareResults/TableHeader.tsx
Carla-Moz: Yes, that works. The condition isn't needed. Thanks!

PR #809, opened 2024-11-21
Use keys for column filters

julienw:
This is another preparatory work for the "persist filters into the URL" feature.
The goal for this PR is to add a key to each possible value to be filtered, so that we can use it in the URL. I want to uncouple the label from the key in the URL, so that we can change the label in the future but keep old URLs working.

This specific PR shouldn't have any user-facing change.

* Commit 1: don't look at it, it is #807 
* Commit 2: small type changes to make it easier to work with the table columns configuration types
* Commit 3: add keys to the `possibleValues` entries but otherwise do not use them too much
* Commit 4: use the keys in the react state

In the next PR, it will be straightforward to add the keys from the URL. The alternative would have been to store the labels in the react state (like it is now), and translate between keys and labels when saving to the URL and reading from the URL, but it seemed less straightforward.


src/components/CompareResults/ResultsTable.tsx
```diff
@@ -16,17 +16,25 @@ import ResultsControls from './ResultsControls';
...
+    possibleValues: [
+      { label: 'Windows', key: 'windows' },
+      { label: 'OSX', key: 'osx' },
+      { label: 'Linux', key: 'linux' },
+      { label: 'Android', key: 'android' },
+    ],
+    matchesFunction(result, valueKey) {
+      const label = this.possibleValues.find(
+        ({ key }) => key === valueKey,
+      )?.label;
```
julienw, line 35: Initially I wasn't a big fan of these `find` operations, because this function is called in a loop on each of the results.
But then I profiled and found this to be negligible compared to the full react rendering, so I decided that we may optimize this later...

beatrice-acasandrei (approved):
Looks good to me 👍

julienw:
Thanks for the review!

PR #812, opened 2024-12-05
Adjust table styles and alignments

julienw:
This patch should be easier to review by ignoring whitespace changes.

I tried to split that in separate commits but the last one is still a bit "catch-all". So there are 3 commits:

### Commit 1: fixes alignment in both main results page and subtests page
This changes the table configurations for a better alignment.
Especially it increases the value for Confidence because the header takes too much space with the button, and use hardcoded pixel values for the buttons. I also removed the "%" string in the Delta title because it's also present in all cells, so it looked superfluous. I used the devtools' grid tool to check the alignments.

### Commit 2: simplify how the background color is specified in the rows.
Instead of specifying it for each cell, specify it just once.

There should be no user change for that commit.

### Commit 3: Various unordered changes, especially to the subtests view
In the subtests view I tried to simplify the DOM by removing some containers here and there. In the subtests view It also fixes the space between rows ([bug 1917856](https://bugzilla.mozilla.org/show_bug.cgi?id=1917856)).

It also changes how the expandable row right margin is implemented (previously: with `width:97%`, now with a real pixel-based margin with the same value as the button). In the future we might want to change the "expandable" part is implemented, by putting the button outside of the row and having the row + expanded part in the same container. This would avoid hardcoding the same value in a lot of places.

I'm sorry that this commit is a bit of a catch-all but I didn't know really how to split it further...

There should be no visible change in the main results page from commit 3 (except for the border-radius in the content of an expanded row).

### Some preview links

Main page: [deploy preview](https://deploy-preview-812--mozilla-perfcompare.netlify.app/compare-results?baseRev=d140333670bcd3103a668a5ec04ed438bb192368&baseRepo=mozilla-central&newRev=250be4ca3c669db9a397456402f68249aa15d8d5&newRepo=mozilla-central&framework=1) / [main branch](https://main--mozilla-perfcompare.netlify.app/compare-results?baseRev=d140333670bcd3103a668a5ec04ed438bb192368&baseRepo=mozilla-central&newRev=250be4ca3c669db9a397456402f68249aa15d8d5&newRepo=mozilla-central&framework=1)
Subtests page: [deploy preview](https://deploy-preview-812--mozilla-perfcompare.netlify.app/subtests-compare-results?baseRev=d140333670bcd3103a668a5ec04ed438bb192368&baseRepo=mozilla-central&newRev=250be4ca3c669db9a397456402f68249aa15d8d5&newRepo=mozilla-central&framework=13&baseParentSignature=92557&newParentSignature=92557) / [main branch](https://main--mozilla-perfcompare.netlify.app/subtests-compare-results?baseRev=d140333670bcd3103a668a5ec04ed438bb192368&baseRepo=mozilla-central&newRev=250be4ca3c669db9a397456402f68249aa15d8d5&newRepo=mozilla-central&framework=13&baseParentSignature=92557&newParentSignature=92557)

#### Main results page:

Before:
image

After:
image

#### Subtests page

Before:
image

After:
image


src/components/CompareResults/RevisionRowExpandable.tsx
```diff
@@ -37,11 +37,13 @@ function RevisionRowExpandable(props: RevisionRowExpandableProps) {
...
       backgroundColor: themeColor200,
       padding: Spacing.Medium,
-      width: '97%',
+      borderRadius: `0px 0px ${Spacing.Small}px ${Spacing.Small}px`,
+      marginInlineEnd: 34 /* This value needs to be synchronized with the expand icon size. */,
     }),
     content: style({
       backgroundColor: contentThemeColor,
       padding: Spacing.Medium,
+      borderRadius: Spacing.xSmall,
```
julienw, line 46: Looks like the border-radius was missing here

src/components/CompareResults/SubtestsResults/SubtestsRevisionRow.tsx
```diff
@@ -51,18 +51,11 @@ function getStyles(themeMode: string) {
           justifyContent: 'start',
           paddingInlineStart: '15%',
         },
-        '.expand-button-container': {
-          justifyContent: 'right',
-        },
         '.subtests': {
           borderRadius: '4px 0 0 4px',
-          paddingLeft: Spacing.xLarge,
+          paddingLeft: Spacing.Medium, // Synchronize with its header
```
julienw, line 56: To align with the header

src/components/CompareResults/SubtestsResults/SubtestsTableContent.tsx
```diff
@@ -1,31 +1,23 @@
-import { style } from 'typestyle';
-
-import { Spacing } from '../../../styles';
 import type { CompareResultsItem } from '../../../types/state';
 import SubtestsRevisionRow from './SubtestsRevisionRow';
 
-const styles = {
-  tableBody: style({
-    marginTop: Spacing.Large,
-  }),
```
julienw, line 10: This was making the extra space between rows.

src/components/CompareResults/TableHeader.tsx
```diff
@@ -154,7 +154,7 @@ function TableHeader({
       background:
         themeMode == 'light' ? Colors.Background100 : Colors.Background300Dark,
       borderRadius: '4px',
-      padding: Spacing.Small,
```
julienw, line 157: By removing the padding left and right, this makes the gris align better.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -112,8 +112,9 @@ const cellsConfiguration: CompareResultsTableConfig = [
 
     gridWidth: '1fr',
   },
-  { key: 'buttons', gridWidth: '2fr' },
-  { key: 'expand', gridWidth: '0.2fr' },
+  // We use the real pixel value for the buttons, so that everything is better aligned.
+  { key: 'buttons', gridWidth: `calc(3.5 * 34px)` }, // 2 or 3 buttons, so at least 3*34px, but give more so that it can "breathe"
```
julienw, line 116: I wasn't sure if we should add some more space, but I find that value doesn't look too bad.

It's especially useful when the "tags" are numerous on the row above, so that the sight can catch the run values easily (see the screenshots for the "main page").

src/components/CompareResults/RevisionRowExpandable.tsx
```diff
@@ -37,11 +37,13 @@ function RevisionRowExpandable(props: RevisionRowExpandableProps) {
     expandedRow: style({
       backgroundColor: themeColor200,
       padding: Spacing.Medium,
-      width: '97%',
+      borderRadius: `0px 0px ${Spacing.Small}px ${Spacing.Small}px`,
```
julienw, line 40: Note that the border-radius from ExpandableRow's style wasn't applied on the right element... That's the problem with too many nested elements :-)

beatrice-acasandrei (approved):

PR #815, opened 2024-12-06
Sort by delta

julienw:
It's by no means ready, but I wanted to give you folks a sneak peak.

Currently I'm focusing on only the subtests view.

### DIFF
--- a/src/components/CompareResults/ResultsTable.tsx
+++ b/src/components/CompareResults/ResultsTable.tsx
@@ -20,7 +20,6 @@ import TableHeader from './TableHeader';
 const cellsConfiguration: CompareResultsTableConfig = [
   {
     name: 'Platform',
-    disable: true,
     filter: true,
     key: 'platform',
     gridWidth: '2fr',
@@ -56,7 +55,6 @@ const cellsConfiguration: CompareResultsTableConfig = [
   },
   {
     name: 'Status',
-    disable: true,
     filter: true,
     key: 'status',
     gridWidth: '1.5fr',
@@ -83,7 +81,6 @@ const cellsConfiguration: CompareResultsTableConfig = [
   },
   {
     name: 'Confidence',
-    disable: true,
     filter: true,
     key: 'confidence',
     gridWidth: '1.5fr',
--- a/src/components/CompareResults/ResultsTable.tsx
+++ b/src/components/CompareResults/ResultsTable.tsx
@@ -36,6 +36,11 @@ const cellsConfiguration: CompareResultsTableConfig = [
       const platformName = getPlatformShortName(result.platform);
       return platformName === label;
     },
+    sortFunction(resultA, resultB) {
+      const platformA = getPlatformShortName(resultA.platform);
+      const platformB = getPlatformShortName(resultB.platform);
+      return platformA.localeCompare(platformB);
+    },
   },
   {
     name: 'Base',
@@ -78,6 +83,9 @@ const cellsConfiguration: CompareResultsTableConfig = [
     name: 'Delta',
     key: 'delta',
     gridWidth: '1fr',
+    sortFunction(resultA, resultB) {
+      return resultA.delta_percentage - resultB.delta_percentage;
+    },
   },
   {
     name: 'Confidence',
@@ -101,6 +109,9 @@ const cellsConfiguration: CompareResultsTableConfig = [
           return result.confidence_text === label;
         }
       }
+    },
+    sortFunction(resultA, resultB) {
+      return (resultA.confidence || 0) - (resultB.confidence || 0);
     },
   },
   {
--- a/src/components/CompareResults/ResultsTable.tsx
+++ b/src/components/CompareResults/ResultsTable.tsx
@@ -189,6 +189,9 @@ export default function ResultsTable() {
           filters={tableFilters}
           onToggleFilter={onToggleFilter}
           onClearFilter={onClearFilter}
+          sortDirection={null}
+          sortColumn={null}
+          onToggleSort={() => {}}
         />
       </Box>
       {/* Using a key in Suspense makes it that it displays the fallback more
--- a/src/components/CompareResults/SubtestsResults/SubtestsResultsTable.tsx
+++ b/src/components/CompareResults/SubtestsResults/SubtestsResultsTable.tsx
@@ -65,7 +65,6 @@ const cellsConfiguration: CompareResultsTableConfig = [
   },
   {
     name: 'Status',
-    disable: true,
     filter: true,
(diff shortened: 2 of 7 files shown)


src/types/types.ts
```diff
@@ -14,14 +19,21 @@ export type CompareResultsTableFilterableCell = {
...
-      name?: string;
-      filter?: false;
-      key: string;
-      gridWidth?: string;
-    }
-  | CompareResultsTableFilterableCell;
+  | BasicCell
+  | (BasicCell & SortableCell)
+  | CompareResultsTableFilterableCell
+  | (CompareResultsTableFilterableCell & SortableCell);
```
julienw, line 36: I thought that splitting the types into subtypes would made it more readable

src/components/CompareResults/SubtestsResults/SubtestsResultsTable.tsx
```diff
@@ -134,7 +134,7 @@ function resultMatchesColumnFilter(
   const cellConfiguration = cellsConfiguration.find(
     (cell) => cell.key === columnId,
   );
-  if (!cellConfiguration || !cellConfiguration.filter) {
+  if (!cellConfiguration || !('filter' in cellConfiguration)) {
```
julienw, line 159: This is for Typescript. Type refining works with the `in` operator but not with the form we were gusin before, curiously.  I asked them and it seems to be made on purpose (see https://github.com/microsoft/TypeScript/issues/51181#issuecomment-2517826823).

src/hooks/useTableSort.ts
```diff
@@ -0,0 +1,26 @@
+import { useState } from 'react';
+
+// This hook handles the state that handles table sorting.
+const useTableSort = () => {
```
julienw: In later commit this hook will be more complex, because it will handle also the URL change.

src/types/types.ts
Carla-Moz: Thanks for this, but maybe more documentation is needed later to understand how we're successfully capturing all the possible type cases for `CompareResultsTableCell` so it doesn't produce an error? Is it standard practice to have all these possible options listed like this for a type or would it be too much to find the type through a function? Just wondering.

Carla-Moz:
I'm somewhat concerned with having the same dropdown to filter _and_ sort, esp for the confidence column - it's looking pretty busy. 

Usually, we'd like one dropdown or button to do one thing either filter or sort, especially since we have so many options for filtering. 

I suggest the `confidence` sorting be separated from the dropdown filter and be an arrow up or down button like you have for the `delta`. That would mean we could have the caret down icon beside the confidence (and other columns) to open up the filter dropdown. What do you think?

Screenshot 2024-12-17 at 14 41 32

Screenshot 2024-12-17 at 14 44 51

The sorting icon and caret down button should do separate things.

Carla-Moz (changes requested):
Thanks for an easy to follow review! I'd like to hear your thoughts on separating the concerns for the confidence filter/sort dropdown/icon.

src/types/types.ts
julienw: It's clearly not ideal especially if there are even more combinations later. I thought this could be a problem for later _if_ more types are indeed needed :-) Definitely I can add some comments to explain the types, I should have done that before the review indeed!

Also I could do something such as:
```ts
export type CompareResultsTableFilterableCell = BasicCell & FilterableCell;
type MaybeSortable<C extends BasicCell> = C | (C & SortableCell);
export type CompareResultsTableCell =
  | MaybeSortable<BasicCell>
  | MaybeSortable<CompareResultsTableFilterableCell>;
```
to factor the union with `SortableCell` into a separate type. Would that be better?


I'm curious what you have in mind when you write `find the type through a function`. Can you provide an example (even a simpler one)?

julienw:
> I suggest the confidence sort action be separate from the dropdown filter actions. The confidence sorting icon button can behave like the delta sorting icon btn. What do you think?

yeah that's a good point, i'll see if I can come up with something!

src/types/types.ts
Carla-Moz: The `MaybeSortable` utility type is more aligned with what I was thinking when I said `find the type through a function`, which would require more effort and be more confusing. 
Example: 

```
//you'd have to disable the any tsc eslint error
const findType = (cell: any} => {
  if ('filter' in cell) {
    return 'filterable';
  }
  if ('sortFunction' in cell) {
    return 'sortable';
  }
  return 'basic';
};

// pass down findType
//set state and the appropriate types
// do things based on findType
// more work and have to figure out other things :(
```

-- also I think my potential type error concerns are already handled through differentiating between the cell types so it would be redundant [TableHeader.tsx#L376](https://github.com/mozilla/perfcompare/blob/df5efb51832d0dbe1492b8b63c24b1136f9287bf/src/components/CompareResults/TableHeader.tsx#L376)

src/types/types.ts
```diff
@@ -1,42 +1,62 @@
...
-  | (CompareResultsTableFilterableCell & SortableCell);
+  | FilterableCell
+  | SortableCell
+  | FilterableAndSortableCell;
 
+// The full configuration for a results table.
 export type CompareResultsTableConfig = CompareResultsTableCell[];
 
+/* --- End of types for configuring the behavior and styles of the results tables columns --- */
+
```
Carla-Moz, line 60: This is so much better! Thank you for taking the time to document it so clearly

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -17,7 +17,7 @@ import { Framework } from '../../types/types';
 import type { CompareResultsTableConfig } from '../../types/types';
 import { getPlatformShortName } from '../../utils/platform';
 
-const cellsConfiguration: CompareResultsTableConfig = [
+const columnsConfiguration: CompareResultsTableConfig = [
   {
```
Carla-Moz, line 21: +1 changing to 'column'

Carla-Moz (approved):
This looks complete to me. Thank you for taking the time to document, improve the design, and write clear tests. Excited for this one ^^

src/components/CompareResults/TableHeader.tsx
```diff
@@ -114,27 +108,19 @@ function FilterableColumn({
...
+    ? `${name} (Click to filter values. Some filters are active.)`
+    : `${name} (Click to filter values)`;
 
   return (
     <>
       <Button
-        color='tableHeaderButton'
         {...bindTrigger(popupState)}
+        color='tableHeaderButton'
+        size='small'
```
julienw, line 120: note: this makes the font size the same as for the columns without a button, I find that this looks better then.

PR #816, opened 2024-12-09
PCF-562 Consider adding the OS version to the name of the platform in the "Platform" column

esanuandra:
[Jira link](https://mozilla-hub.atlassian.net/browse/PCF-562)
[Bugzilla link](https://bugzilla.mozilla.org/show_bug.cgi?id=1918565)

[Localhost testing link](http://localhost:3000/compare-results?baseRev=78d8afbe5767fd5e579d3235d46a02d1c255e3cf&baseRepo=mozilla-central&newRev=01166de9b45b0edb29c9c97a8d9e2ca9023da37a&newRepo=mozilla-central&framework=13)
[Deploy testing link](https://deploy-preview-816--mozilla-perfcompare.netlify.app/compare-results?baseRev=78d8afbe5767fd5e579d3235d46a02d1c255e3cf&baseRepo=mozilla-central&newRev=01166de9b45b0edb29c9c97a8d9e2ca9023da37a&newRepo=mozilla-central&framework=13)

Screenshot 2024-12-18 at 16 06 34


src/utils/platform.ts
```diff
@@ -16,6 +16,23 @@ export const getPlatformShortName = (
...
+      .split('-')[0]
+      .replace('win', '')}`;
+  if (platform.startsWith('macosx'))
+    return `${platformShortName}\u00a0${platform
+      .split('-')[0]
+      .replace('macosx', '')}`;
+  if (platform.startsWith('linux'))
+    return `${platformShortName}\u00a0${platform
+      .split('-')[0]
+      .replace('linux', '')}`;
```
esanuandra, line 35: to add comments with example

src/types/types.ts
```diff
@@ -83,6 +83,25 @@ export type FilteredResults = {
   isFiltered: boolean;
 };
 
+export type PlatformOS =
```
julienw, line 86: I'd remove the type from here, as it's used only once there's no reason to have an exported type. It's used only as a tool in platform.ts. So instead I'd do this in platform.ts:
```ts
type PlatformOS = keyof typeof osMapping;
```
So that we don't have to maintain the data in both places.

src/utils/platform.ts
```diff
@@ -14,12 +15,42 @@ export const getPlatformShortName = (
...
+const extractPlatformWithOs = (platform: string) => {
+  const name = osMapping[platform.split('-')[0] as PlatformOS];
+
+  if (name) return name;
+
+  return getPlatformShortName(platform);
+};
+
 export const getPlatformAndVersion = (platform: string): string => {
   const platformShortName = getPlatformShortName(platform);
```
julienw, line 47: you can move the call to `getPlatformShortName` inside the if block now, as it's used only there.
(be careful to add the brackets :D)

julienw (approved):
Thanks, this looks pretty good to me!
I only have some nits for maintenance and readability, but feel free to merge without a new review after you fix them. As long as the test don't change that's good :-)

PR #818, opened 2024-12-10
PCF-576 Add checkboxes to the table filters dropdown

beatrice-acasandrei:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/PCF-576)

Before:
Screenshot 2024-12-10 at 18 52 43

After:
Screenshot 2024-12-13 at 14 17 13

[deploy](https://deploy-preview-818--mozilla-perfcompare.netlify.app/compare-results?baseRev=e3b3fa7a28a0e217ef910e1ec71d903eb01c7970&baseRepo=mozilla-central&newRev=70cf5d90c346d5954f08633c22f5eb6b6cd90d66&newRepo=mozilla-central&framework=1) vs [production](https://perf.compare/compare-results?baseRev=e3b3fa7a28a0e217ef910e1ec71d903eb01c7970&baseRepo=mozilla-central&newRev=70cf5d90c346d5954f08633c22f5eb6b6cd90d66&newRepo=mozilla-central&framework=1)

### DIFF
--- a/src/components/CompareResults/TableHeader.tsx
+++ b/src/components/CompareResults/TableHeader.tsx
@@ -1,6 +1,6 @@
-import CheckIcon from '@mui/icons-material/Check';
 import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
 import Button from '@mui/material/Button';
+import Checkbox from '@mui/material/Checkbox';
 import Divider from '@mui/material/Divider';
 import Menu from '@mui/material/Menu';
 import MenuItem from '@mui/material/MenuItem';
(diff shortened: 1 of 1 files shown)


beatrice-acasandrei:
@julienw @Carla-Moz can you take a look at the description and let me know your thoughts?

julienw:
Hey beatrice!

I think you need to remove the padding and use ListItemIcon so that this looks good. Currently there's too much space between the items, too much space at the left, and not enough space between the checkbox and the text. Using `ListItemIcon` should fix that, + manually set padding to 0 with `sx`. I think they're also a bit too big, so you need `size="small"` to make them smaller.

FYI I checked about the accessibility (do we need a proper label for the checkbox?) but we don't need to deal with it here because the list item is a `menuitemcheckbox` which effectively hides its children.


Thanks

julienw (approved):
I pushed an update to not use ListItemButton but use MenuItem directly, with a ListItemIcon inside. I also simplified the code. You can have a look at the last commit if you're curious.

I did that so that the text is better aligned when we'll have the "sort" capabilities in the same menu that are using the same `ListItemIcon`. (and also because the code is simpler as a result)

PR #819, opened 2024-12-10
Upgrade typescript, eslint configuration and prettier

julienw:
In this PR I'm updating a bunch of things. My initial nudge has been that I wanted to use a new Web API in my upcoming patch, and TypeScript didn't know about it, so I went on with upgrading TypeScript. This gave a warning with the typescript eslint plugin, so I decided to upgrade it as well. And therefore I had to update more things.
I also removed the "airbnb" config, because it's deprecated and I thought that using the recommended options from eslint itself were good enough.
I also updated prettier, this wasn't 100% necessary but this was easy.

As a result running `npm run lint` and `npm run tsc` are faster now, by approximately 20%.


Carla-Moz (approved):

julienw:
Thanks for the quick approval!

PR #821, opened 2024-12-11
PCF-582 PerfCompare link in the graphs view fails with: Error: r[0] is undefined

beatrice-acasandrei:


src/components/CompareResults/SubtestsResults/SubtestsResultsTable.tsx
```diff
@@ -222,8 +221,6 @@ function SubtestsResultsTable({
           rowGridTemplateColumns={rowGridTemplateColumns}
         />
       ))}
-
-      {processedResults.length == 0 && <NoResultsFound />}
```
julienw, line 226: I think it would be good to keep that here, for when the user search or filter filters everything out.

src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
```diff
@@ -81,6 +68,34 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
     updateRawSearchParams(rawSearchParams);
   };
 
+  if (!results.length) {
+    return (
+      <React.Fragment>
```
julienw, line 73: no need to change, but as a reminder current versions of React accept to use `<>` instead of `React.Fragment`, I find it less noisy and faster to type, but it's really a stylistic nit so this can also stay that way.

julienw (approved):
Thanks, this looks good to me! Just one small comment :-)
Do you think you could add a quick test for that?

Also for next time, please remember to add a relevant deploy preview link :-)
Thanks!

PR #822, opened 2024-12-11
add input background color to select input when focused

Netacci:
[Bugzilla link](https://bugzilla.mozilla.org/show_bug.cgi?id=1925640)
[Deploy preview link](https://deploy-preview-822--mozilla-perfcompare.netlify.app)

Before
When select input is focused, the default MUI background color doesn't cover the padded section. The padded section retains the select input background color. 

focus

After
To resolve this, I had two ideas in mind, 

1. Add a negative margin(left and right) and padding to the focused input which would make the bgColor cover everywhere 
2. Override the MUI select input(focus) default styles, using the input background color so that it retains same color when focused.

I went with the latter option because adding a negative margin seemed hacky.

color


src/theme/components.js
```diff
@@ -249,6 +249,12 @@ const components = {
           outline: '1px auto',
         },
       },
+      select: {
+        '&:focus': {
+          backgroundColor: Colors.Background300Dark,
```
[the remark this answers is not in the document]
Carla-Moz: Maybe unique classes would work better here. What do you think? 

```
    select: {
        '&.select-light': {
          '&:focus': {
            backgroundColor: Colors.SecondaryDefault,
          },
          '&:hover': {
            padding: 0,
            backgroundColor: Colors.SecondaryHover,
          },
        },
        '&.select-dark': {
          '&:focus': {
            backgroundColor: Colors.Background300Dark,
          },
          '&:hover': {
            padding: 0,
            backgroundColor: Colors.SecondaryHoverDark,
          },
        },
      },
```

But for this to work, you'd have to update both the dropdown components that use `Select`, so it's more work but at least it will move us away from `typestyle`.

https://github.com/Netacci/perfcompare/blob/6aa592eac274eef1fb470b86144aace635560fb3/src/components/Search/SearchDropdown.tsx#L61C10-L61C12

You'd have to add an `inputProps `with the new class to all the `Select` components for example:

```
inputProps={{
   classes: {
     select: mode === 'light' ? 'select-light' : 'select-dark',
      },
}}

```

Carla-Moz (changes requested):
Thanks for the work here. I added some suggestions to get us more in line with the future regarding how we would like to style moving on.

src/theme/components.js
```diff
@@ -249,6 +249,12 @@ const components = {
...
+          borderRadius: '4px',
```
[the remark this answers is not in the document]
Netacci: > This borderRadius style is not necessary; it is already taken care of in `src/styles/DropdownMenu.ts` Speaking of that file, I noticed the wrong color bug happens because of this. line
> 
> https://github.com/Netacci/perfcompare/blob/6aa592eac274eef1fb470b86144aace635560fb3/src/styles/DropDownMenu.ts#L68.
> 
> You can go ahead and delete the focus styles for both dark and light in this file because we will handle them below in the component file instead.

Oh alright. I missed that file during my troubleshooting :)

src/theme/components.js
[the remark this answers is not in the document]
Netacci: There's this flickering that happens when I hover over the select. Removing the `padding:0` seems to fix the issue

https://github.com/user-attachments/assets/05da47f5-1932-448d-a8d7-dd6d1fe46428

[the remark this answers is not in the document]
Carla-Moz: Ahh, you're right! Okay, thanks for catching.

src/theme/components.js
```diff
@@ -250,9 +250,21 @@ const components = {
...
       },
       select: {
-        '&:focus': {
-          backgroundColor: Colors.Background300Dark,
-          borderRadius: '4px',
+        '&.select-light': {
+          '&:focus': {
+            backgroundColor: Colors.SecondaryDefault,
+          },
+          '&:hover': {
```
Carla-Moz, line 257: You can remove the hover styles here.

Carla-Moz, line 257: Could add a padding of: 

`padding: '3px 0px 5px',`

src/theme/components.js
```diff
@@ -250,9 +250,21 @@ const components = {
...
+            backgroundColor: Colors.SecondaryHover,
+          },
+        },
+        '&.select-dark': {
+          '&:focus': {
+            backgroundColor: Colors.Background300Dark,
+          },
+          '&:hover': {
+            backgroundColor: Colors.SecondaryHoverDark,
```
Carla-Moz, line 266: Same here:
padding: '3px 0px 5px',

Carla-Moz (changes requested):

src/theme/components.js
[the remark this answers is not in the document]
Carla-Moz: Yes, that's because we have separate hover styles for the `selected:focus`. However, on regular hover, there's extra padding. 
Screenshot 2024-12-17 at 15 34 06

Could you set the hover padding to : padding: '3px 0px 5px' ? And see if that makes it look perfect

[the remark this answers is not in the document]
Carla-Moz: There's only a flick when you hover on the sides, but that's ok for now. What's important that it stays one color on focus. Some other style finishes I'll take care of when I revamp the styles later.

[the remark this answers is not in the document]
Netacci: Alright. will do

src/theme/components.js
```diff
@@ -249,6 +249,24 @@ const components = {
           outline: '1px auto',
         },
       },
+      select: {
+        '&.select-light': {
```
julienw, line 253: there's a much easier way than using classes and editing everything everywhere. Indeed MUI knows about the theme!
Look at how we're doing with MuiTooltip below: using a function that takes `({theme})` as a parameter. This function can return an object with the 2 properties `&:focus` and `&:hover` depending on the theme.

Netacci: Okay. This works too. Created a `Select.ts` file under styles and moved two properties for light and dark.

julienw: You haven't pushed the update yet, but this sounds good to me :) Carla will review it once you'll push.
Just make sure to revert all the changes that add the classes everywhere :-)

Carla-Moz: Ah, yes that's so true, lol. I don't know why I keep going back to the longer method. Thanks @julienw.

src/styles/Select.ts
```diff
@@ -0,0 +1,28 @@
+import { Colors } from './Colors';
+
+export const SelectStyles = {
+  light: {
+    '&:focus': {
+      backgroundColor: Colors.SecondaryDefault,
+      paddingTop: '0px',
```
Carla-Moz, line 7: Almost done! You only have to remove the` paddingTop and paddingBottom` light and dark styles for 'focus' and replace them with `minWidth: '160px'` This will eliminate the flashing on hover
Screenshot 2024-12-18 at 10 53 20

Netacci: I still experience the flashing when I add `minWidth:'160px' ` to `focus`

https://github.com/user-attachments/assets/fa454379-32be-4d39-bc88-4bd68eaf7ea9

The only time I don't see the flashing is when I remove `paddingTop` and `paddingBottom` from both `focus` and `hover`(Hover and focus have only backgroundColor)

https://github.com/user-attachments/assets/668361a7-72a3-4da3-8d5b-0dbf382a1ae8

Carla-Moz: That's fine for now. The different color on the padded left is for `on hover when focused` goes beyond the scope of this PR, which is to `keep it one solid color when focused`. We need to fix the padding of MuiInputBase to fix that clash and that can be a quick separate PR. 

Also, we have to leave the hover padding styles as is for now because of the issue I mentioned above:

image

Carla-Moz: So once you push those final changes, we can get this merged. :)

julienw:
Thanks!

Netacci:
I know this is closed but I observed this also. 

https://github.com/user-attachments/assets/dbe3696f-0863-4a7f-8cbc-82391da4aa11

Is this part of the fix that will be done in a separate PR? @Carla-Moz

Carla-Moz:
> I know this is closed but I observed this also.
> 19.12.2024_13.38.15_REC.mp4
> 
> Is this part of the fix that will be done in a separate PR? @Carla-Moz

Could you please file a bugzilla big for this and try opening a PR for a fix?  Next time, we'll check all the UI elements when playing around with the select element. Here's the link: https://bugzilla.mozilla.org/enter_bug.cgi?product=Testing&component=PerfCompare&status_whiteboard=%5Bpcf%5D

julienw:
That's the issue when dealing with the global styles! You'll need to think and decide which behavior is the most generic one and put that to the global styles, and then override them in the specific elements using `sx`.

Carla-Moz:
@julienw So true!

PR #825, opened 2024-12-16
Allow text copying from the edit revision panel

Netacci:
[Bugzilla link](https://bugzilla.mozilla.org/show_bug.cgi?id=1925000)
[Local preview link](http://localhost:3000/compare-results?baseRev=38ad50fbdf916b73179812f7f9b361178ac1f7e5&baseRepo=mozilla-central&newRev=a861e5955acebded52e664b0ca290e040fa933fb&newRepo=mozilla-central&framework=13)
[Deploy preview link](https://deploy-preview-825--mozilla-perfcompare.netlify.app/compare-results?baseRev=38ad50fbdf916b73179812f7f9b361178ac1f7e5&baseRepo=mozilla-central&newRev=a861e5955acebded52e664b0ca290e040fa933fb&newRepo=mozilla-central&framework=13)

### Overview
User can't select or copy a text on the edit revision panel. This was happening because the area is a button and texts in buttons can't be selected or copied. 

An immediate fix would have been to add `user-select: "text"` to the button style, but since this button doesn't perform any action, removing it was the ideal way to go.

To fix this, `listItemButton` element was changed to `Box` element, and the style was updated so it doesn't affect the UI.

Before

https://github.com/user-attachments/assets/2a5deb8b-4962-4e42-8aea-f7486c7bc78c

After

https://github.com/user-attachments/assets/c3f41b04-8660-40de-8e6d-df46ae6bb68f


src/styles/SelectedRevsStyle.ts
```diff
@@ -52,19 +52,17 @@ export const SelectRevsStyles = (mode: string) => {
...
 
-    listItemButton: {
-      cursor: 'auto !important',
-      padding: '0 !important',
-      alignItems: 'flex-start !important',
+    selectedRevision: {
+      display: 'flex',
+      cursor: 'auto',
+      padding: '0',
+      alignItems: 'flex-start',
```
julienw, line 59: you don't need `cursor` `padding` and `alignItems`, these values are the default values for these properties. It was necessary before to override the button's css properties.

src/styles/SelectedRevsStyle.ts
```diff
@@ -52,19 +52,17 @@ export const SelectRevsStyles = (mode: string) => {
...
+      minWidth: '0px',
       $nest: {
         '&:hover': {
-          backgroundColor: 'transparent !important',
+          backgroundColor: 'transparent',
```
julienw, line 63: I think it's not needed either, this was also used to override the button's css properties.

src/styles/SelectedRevsStyle.ts
```diff
@@ -94,10 +92,6 @@ export const SelectRevsStyles = (mode: string) => {
                 : Colors.SecondaryTextDark,
             },
 
-            svg: {
-              marginRight: `${Spacing.xSmall}px`,
-              fontSize: '1rem',
```
julienw, line 99: I think this is still needed, otherwise the icons are too small and too close to the text. There might be other ways to do it later (like providing them on the SVG icon themselves) but let's keep it for now.

See:
Before:
image

After:
image

src/styles/SelectedRevsStyle.ts
```diff
@@ -52,19 +52,17 @@ export const SelectRevsStyles = (mode: string) => {
...
       $nest: {
         '&:hover': {
-          backgroundColor: 'transparent !important',
+          backgroundColor: 'transparent',
         },
 
-        '.search-revision-item-icon': {
-          minWidth: '0',
-          marginLeft: `${Spacing.Small + 6}px`,
-        },
```
julienw, line 67: Thanks for removing this unused style!

julienw (changes requested):
Thanks for the patch! Here are some comments to fix before landing, tell me what you think! :-)

src/styles/SelectedRevsStyle.ts
Netacci: Oh okay I see that now

src/styles/SelectedRevsStyle.ts
Netacci: Okay I missed that :) I thought it was unused. Adding it back now

julienw (approved):
Thanks, this looks great!

The next step would be to add a "copy" icon to copy this easily. I filed a dedicated bug with some steps in https://bugzilla.mozilla.org/show_bug.cgi?id=1938295 if you want to take a look :-)

PR #826, opened 2024-12-17
Add aria expanded to revision row expand button

Netacci:
[Bugzilla link](https://bugzilla.mozilla.org/show_bug.cgi?id=1935720)
[Bugzilla link 2](https://bugzilla.mozilla.org/show_bug.cgi?id=1935719)

[Deploy preview link](https://deploy-preview-826--mozilla-perfcompare.netlify.app/)

### Description
This PR improves the accessibility of the expand button for the revision row by adding proper ARIA attributes.
- Added `aria-expanded` to dynamically set to true or false based on the expanded state of the row.
- Added `aria-controls` that point to the `id` of the expandable section (`revision-row-expandable`)
- Changed the `div` being used for the expandable row to a `section`

### DIFF
--- a/src/components/CompareResults/RevisionRowExpandable.tsx
+++ b/src/components/CompareResults/RevisionRowExpandable.tsx
@@ -1,3 +1,4 @@
+import { Box } from '@mui/material';
 import Divider from '@mui/material/Divider';
 import { style } from 'typestyle';
 
@@ -71,9 +72,10 @@ function RevisionRowExpandable(props: RevisionRowExpandableProps) {
   };
 
   return (
-    <div
+    <Box
       id='revision-row-expandable'
-      aria-label='revision-row'
+      aria-label='Revision Row Details'
+      component={'section'}
       className={`${styles.expandedRow}`}
     >
       <div className={`${styles.content}`}>
@@ -133,7 +135,7 @@ function RevisionRowExpandable(props: RevisionRowExpandableProps) {
           </div>
         )}
       </div>
-    </div>
+    </Box>
   );
 }
 
--- a/src/components/CompareResults/RevisionRowExpandable.tsx
+++ b/src/components/CompareResults/RevisionRowExpandable.tsx
@@ -1,4 +1,3 @@
-import { Box } from '@mui/material';
 import Divider from '@mui/material/Divider';
 import { style } from 'typestyle';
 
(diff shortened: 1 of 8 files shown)


Carla-Moz:
@Netacci Would you mind rebasing this branch before I start looking this over?

Netacci:
> @Netacci Would you mind rebasing this branch before I start looking this over?

This is done now

Carla-Moz (approved):
This looks good too! Thanks!

src/components/CompareResults/RevisionRowExpandable.tsx
```diff
@@ -71,7 +72,12 @@ function RevisionRowExpandable(props: RevisionRowExpandableProps) {
   };
 
   return (
-    <div className={`${styles.expandedRow}`}>
+    <Box
+      id='revision-row-expandable'
```
julienw, line 76: I have a few comments :)

1. the most important: the id needs to be unique, here it's not unique! You need to understand that it's currently the same for each line. In this case, because we don't have a unique id or a key for a result, I believe the easiest way would be the `useId` hook (see https://react.dev/reference/react/useId). You can call it in `RevisionRow` and pass it to `RevisionRowExpandable`. You'll also need to do the same in `SubtestsRevisionRow` BTW.
2. you don't need a `Box`, just use a `section` element directly. (I understand you picked the `section` so that you can add an `aria-label`. I'm not fully sure it's useful but why not...)

julienw (commented):
I left some more comments, please look at them before landing :-)

Especially you need to handle SubtestsRevisionRow as well!

src/components/CompareResults/SubtestsResults/SubtestsRevisionRow.tsx
```diff
@@ -122,6 +122,7 @@ function determineSign(baseMedianValue: number, newMedianValue: number) {
 }
 
 function SubtestsRevisionRow(props: RevisionRowProps) {
+  const id = useId();
```
julienw, line 125: you also need to add the aria-expanded etc in this component, like in the other one :-)

Netacci: ah missed that. fixing now

julienw:
Thanks, this looks good to me now, but you need to merge the main branch and update the snapshots to resolve the conflicts :) when it's done I'll happily merge.

julienw (approved):
thanks!

Netacci:
> Thanks, this looks good to me now, but you need to merge the main branch and update the snapshots to resolve the conflicts :) when it's done I'll happily merge.

Done now

PR #828, opened 2024-12-19
Format numbers in revision and subtests revision rows

Netacci:
[Bugzilla link](https://bugzilla.mozilla.org/show_bug.cgi?id=1920829)

[Deploy preview link]()

### Description
This PR formats all numbers on Revision and Subtest revision row.

**Before**

Revision Row

unformated

Expanded Revision Row

unformated-exandable

Subtests Revision Row

unformated-subtests

Changes
I created a function called `formatNumber` in the `format.ts` file. This function formats displayed numbers in the Revision and Subtests rows using `Intl.NumberFormat`

**After**

Revision Row

formatted

Expanded Revision Row

formatted-expandable

Subtests Revision Row

formatted-subtests


src/utils/format.ts
```diff
@@ -9,3 +9,7 @@ export function formatDate(date: Date) {
 export function formatDateRange(date1: Date, date2: Date) {
   return formatDate(date1) + ' - ' + formatDate(date2);
 }
+
+export const formatNumber = (value: number) => {
+  return Intl.NumberFormat('en-US').format(value);
```
julienw, line 14: It's a lot more performant to instantiante `Intl.NumberFormat` only once outside of the function, and use this single instance inside the function.
Indeed the constructor of `Intl.NumberFormat` does a ton of things.

Netacci: Oh alright. The same way the `dateFormatter` was implemented.

Netacci: I couldn't use the function here https://github.com/Netacci/perfcompare/blob/184dfb1bfcddc121399c129668ddd0cdcd0440ad/src/components/CompareResults/RevisionRowExpandable.tsx#L111 because `medianDifference` is a string. Was that done deliberately?

Carla-Moz (approved):
Looks like you were able to resolve this. LGTM! Thanks.

src/components/CompareResults/RevisionRowExpandable.tsx
```diff
@@ -29,12 +30,17 @@ function RevisionRowExpandable(props: RevisionRowExpandableProps) {
     new_is_better: newIsBetter,
   } = result;
 
+  const NumberFormat = (value: number) => {
+    return new Intl.NumberFormat('en-US', {
+      maximumFractionDigits: 2,
+    }).format(value);
```
julienw, line 36: Please create Intl.NumberFormat outside of the component, so that the instance is reused for each render. It's very very important for performance.

Netacci: Okay noted

src/components/CompareResults/RevisionRowExpandable.tsx
```diff
@@ -6,10 +6,15 @@ import { useAppSelector } from '../../hooks/app';
...
 import { Colors, Spacing } from '../../styles';
 import type { CompareResultsItem } from '../../types/state';
+import { formatNumber } from './../../utils/format';
 
 const strings = Strings.components.expandableRow;
 const { singleRun } = strings;
-
+const NumberFormat = (value: number) => {
+  return new Intl.NumberFormat('en-US', {
+    maximumFractionDigits: 2,
```
julienw, line 15: same here: the NumberFormat formatter needs to be instanciated only once

src/components/CompareResults/RevisionRowExpandable.tsx
```diff
@@ -34,7 +39,7 @@ function RevisionRowExpandable(props: RevisionRowExpandableProps) {
   let medianDifference = '';
   let medianPercentage = '';
   if (baseMedian && newMedian) {
-    medianDifference = (newMedian - baseMedian).toFixed(2);
+    medianDifference = NumberFormat(newMedian - baseMedian);
     medianPercentage = (((newMedian - baseMedian) / baseMedian) * 100).toFixed(
```
julienw, line 43: the formatter should be used here as well

src/components/CompareResults/RevisionRowExpandable.tsx
julienw: After you create a formatter I think you won't need the function anymore, just call the formatter directly in the component.

Netacci: Updated now!

src/components/CompareResults/RunValues.tsx
```diff
@@ -41,11 +42,12 @@ function RunValues(props: RunValuesProps) {
...
       {application ? (
         <div>
-          <b>{name}:</b> {median} {measurementUnit} ({application})
+          <b>{name}:</b> {formatNumber(median)} {measurementUnit} ({application}
+          )
         </div>
       ) : (
         <div>
-          <b>{name}:</b> {median} {measurementUnit}
+          <b>{name}:</b> {formatNumber(median)} {measurementUnit}
```
julienw, line 50: it's a bug that we were using median here, can you please move to the `avg` value instead?

src/components/CompareResults/RunValues.tsx
```diff
@@ -47,7 +47,7 @@ function RunValues(props: RunValuesProps) {
         </div>
       ) : (
         <div>
-          <b>{name}:</b> {formatNumber(median)} {measurementUnit}
+          <b>{name}:</b> {formatNumber(avg)} {measurementUnit}
```
julienw, line 50: you also need to do it above Neta

Netacci: Oh okay I didn't realize they should be the same. Thanks for pointing that out

src/components/CompareResults/RunValues.tsx
```diff
@@ -42,8 +42,7 @@ function RunValues(props: RunValuesProps) {
     <div className={styles.container}>
       {application ? (
         <div>
-          <b>{name}:</b> {formatNumber(median)} {measurementUnit} ({application}
-          )
+          <b>{name}:</b> {formatNumber(avg)} {measurementUnit} ({application})
```
Carla-Moz, line 45: From what I can see here, looks like all the necessary values have been formatted. This looks ready to merge. What do you say @julienw

julienw (approved):
I added a commit to rename the `numberFormat` function, for more clarity, but otherwise this looks good to me!

PR #829, opened 2024-12-19
Silence the warning generated by the Breadcrumbs component by inserting a span instead of a fragment

julienw:
This warning was driving me crazy for months :D

The warning (only displayed when running tests or in development mode) was: 
```
    console.error
      MUI: The Breadcrumbs component doesn't accept a Fragment as a child.
      Consider providing an array instead.
```

There should be no user-visible change.

[Deploy preview](https://deploy-preview-829--mozilla-perfcompare.netlify.app/subtests-compare-results?baseRev=d140333670bcd3103a668a5ec04ed438bb192368&baseRepo=mozilla-central&newRev=250be4ca3c669db9a397456402f68249aa15d8d5&newRepo=mozilla-central&framework=13&baseParentSignature=92557&newParentSignature=92557) / [main branch](https://main--mozilla-perfcompare.netlify.app/subtests-compare-results?baseRev=d140333670bcd3103a668a5ec04ed438bb192368&baseRepo=mozilla-central&newRev=250be4ca3c669db9a397456402f68249aa15d8d5&newRepo=mozilla-central&framework=13&baseParentSignature=92557&newParentSignature=92557)


src/components/CompareResults/SubtestsResults/SubtestsBreadcrumbs.tsx
```diff
@@ -59,8 +59,9 @@ function SubtestsBreadcrumbs({ view }: SubtestsBreadcrumbsProps) {
         marginBottom: 3,
       }}
     >
-      [<></>,<Link href='/'>Home</Link>,
-      <Link href={allResultsURL}>All results</Link>, ]
```
julienw, line 63: I think that when this was implemented Beatrice tried to use an array (with the `[]`), but this doesn't work as expected here: indeed this is not an array (because it should be in `{}`), here the brackets are just handled as normal text and then filtered out by the Breadcrumb component in https://github.com/mui/material-ui/blob/17026fc203a8b413ec7343b6eeb267a50519d515/packages/mui-material/src/Breadcrumbs/Breadcrumbs.js#L174

The main issue here is the fragment as the first item. This is done this way so that there's a separator already before the `Home` item.

Instead I'm inserting an empty `span`. I don't love it but the check that I mentioned above wants a real React component. An alternative could be to create a component that just returns `null`... but this would still generate an empty `li`, so I thought a `span` is just as good.

Carla-Moz (approved):
Thanks! I've been meaning to make an issue for this console bug too T-T

PR #832, opened 2024-12-19
Add tooltips to revision column headers

Netacci:
[Bugzilla link](https://bugzilla.mozilla.org/show_bug.cgi?id=1912026)

[Deploy link](url)

### Description

This PR adds a tooltip to the Revision Column header names to give users more description of what each columns do. The tooltip description was obtained from the bugzilla report.

Platform - No tooltip
Base - A summary of all values from Base runs using a median.
New- A summary of all values from New runs using a median.
Status- No tooltip
Delta- The percentage difference between the Base and New values
Confidence- Calculated using a Student's T-test comparison. Low is anything under a T value of 3, Medium is between 3 and 5, and High is anything higher than 5.
Total runs- The total number of tasks/jobs that ran for this metric.

Before screenshot

tooltip-before

After screenshot
tooltip

A description is also added to the test name on hover. I used same text used for the `aria-label`

tooltipname

### DIFF
--- a/src/components/CompareResults/ResultsTable.tsx
+++ b/src/components/CompareResults/ResultsTable.tsx
@@ -42,6 +42,8 @@ const cellsConfiguration: CompareResultsTableConfig = [
     name: 'Base',
     key: 'base',
     gridWidth: '1fr',
+    tooltip: true,
+    tooltipContent: 'A summary of all values from Base runs using a median.',
   },
   {
     key: 'comparisonSign',
@@ -51,8 +53,9 @@ const cellsConfiguration: CompareResultsTableConfig = [
   {
     name: 'New',
     key: 'new',
-
     gridWidth: '1fr',
+    tooltip: true,
+    tooltipContent: 'A summary of all values from New runs using a median.',
   },
   {
     name: 'Status',
@@ -80,6 +83,8 @@ const cellsConfiguration: CompareResultsTableConfig = [
     name: 'Delta',
     key: 'delta',
     gridWidth: '1fr',
+    tooltip: true,
+    tooltipContent: 'The percentage difference between the Base and New values',
   },
   {
     name: 'Confidence',
@@ -87,6 +92,9 @@ const cellsConfiguration: CompareResultsTableConfig = [
     filter: true,
     key: 'confidence',
     gridWidth: '1.5fr',
+    tooltip: true,
+    tooltipContent:
+      "Calculated using a Student's T-test comparison. Low is anything under a T value of 3, Medium is between 3 and 5, and High is anything higher than 5.",
     possibleValues: [
       { label: 'No value', key: 'none' },
       { label: 'Low', key: 'low' },
@@ -109,8 +117,9 @@ const cellsConfiguration: CompareResultsTableConfig = [
   {
     name: 'Total Runs',
     key: 'runs',
-
     gridWidth: '1fr',
+    tooltip: true,
+    tooltipContent: 'The total number of tasks/jobs that ran for this metric.',
   },
   // We use the real pixel value for the buttons, so that everything is better aligned.
   { key: 'buttons', gridWidth: `calc(3.5 * 34px)` }, // 2 or 3 buttons, so at least 3*34px, but give more so that it can "breathe"
--- a/src/components/CompareResults/TableHeader.tsx
+++ b/src/components/CompareResults/TableHeader.tsx
@@ -1,5 +1,6 @@
+import InfoIcon from '@mui/icons-material/InfoOutlined';
 import KeyboardArrowDownIcon from '@mui/icons-material/KeyboardArrowDown';
-import { ListItemIcon, ListItemText } from '@mui/material';
+import { ListItemIcon, ListItemText, Tooltip } from '@mui/material';
 import Button from '@mui/material/Button';
 import Checkbox from '@mui/material/Checkbox';
 import Divider from '@mui/material/Divider';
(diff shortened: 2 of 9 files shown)


Carla-Moz:
Thanks for your work! I have some comments

The column titles have a heavier font weight. They got set to 700 somewhere. We we need to revert to original font weight 

Screenshot 2024-12-20 at 1 04 01 PM

Carla-Moz (changes requested):
Let's try removing the information icon, the bolder text, and only set the cursor to pointer for the column header requiring tooltips . I say this because the header is quite busy with icons and we don't want too much info overload.

Carla-Moz:
The tooltip is also the wrong background color when switching back to light mode. 

Screenshot 2024-12-20 at 1 10 55 PM

Carla-Moz:
I like this arrow down caret to show which column title the tool-tip is referring too. That's why I think it's ok to remove the information icons.

Screenshot 2024-12-20 at 1 20 23 PM

src/components/CompareResults/TestHeader.tsx
```diff
@@ -79,7 +81,7 @@ function createTitle(
           underline='hover'
           target='_blank'
           href={docsURL}
-          title='Link to suite documentation'
+          title={Strings.components.revisionRow.title.suiteLink}
         >
```
Carla-Moz, line 84: How about:

```
const suiteLink = Strings.components.revisionRow.title.suiteLink

...

<Link
          aria-label='link to suite documentation'
          aria-label='link to suite documentation'
          underline='hover'
          underline='hover'
          target='_blank'
          target='_blank'
          href={docsURL}
          href={docsURL}
          title='Link to suite documentation'
          title={suiteLink}
        >
            {result.suite}
</Link>
...

```

Netacci:
> Thanks for your work! I have some comments.
> 
> The column titles have a heavier font weight. They got set to 700 somewhere. Let's revert to original font weight.
> 
> Screenshot 2024-12-20 at 1 04 01 PM


This image is from on perf.compare they seem to have the `fontweight: 700` set there already. what is the original font weight?
perff

Netacci:
> Let's try removing the information icon, the bolder text, and only set the cursor to pointer for the column header requiring tooltips . I say this because the header is quite busy with icons and we don't want too much info overload.

Yes I agree that the header is quite busy. I'll remove the info icon

Carla-Moz (approved):
This looks good to me. Thanks!

src/components/CompareResults/TableHeader.tsx
```diff
@@ -311,21 +310,50 @@ function TableHeader({
...
 
   function renderColumnHeader(header: CompareResultsTableColumn) {
-    if ('sortFunction' in header && 'filter' in header) {
-      return (
-        <ButtonGroup
-          variant='contained'
-          color='tableHeaderButton'
-          disableElevation
-        >
+    const content = (() => {
```
julienw, line 313: please create a separate named function, it's much more readable :) thanks

src/components/CompareResults/TableHeader.tsx
```diff
@@ -336,39 +364,23 @@ function TableHeader({
...
+      return header.name;
+    })();
+    return header.tooltip ? (
+      <Tooltip
+        title={header.tooltip}
+        placement='top'
+        arrow
+        classes={{ tooltip: styles.typography }}
+      >
+        <Box sx={{ cursor: 'pointer' }}>{content}</Box>
```
julienw, line 379: Why is the Box needed? Can you please add a comment explaining why?

src/components/CompareResults/TableHeader.tsx
julienw: alternatively, use a variable:
```js
let content = header.name;
if (...) {
  content = ...
} else if (...) {
  content = ...
} else...

if (header.tooltip) {
  return <Tooltip>...</Tooltip>
}
return content;
```

julienw: It's important to remember that "simple" (and "readable") is more important than "clever".

src/components/CompareResults/TableHeader.tsx
Netacci: Box is used because tooltip expects a single valid element but sometimes content is a string (header.name) I can use a span here too.

julienw: Ah I see, sounds good. Can you add a comment to the code about that ?

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -41,6 +41,7 @@ const columnsConfiguration: CompareResultsTableConfig = [
     name: 'Base',
     key: 'base',
     gridWidth: '1fr',
+    tooltip: 'A summary of all values from Base runs using a median.',
```
julienw, line 44: Ah, we should have seen that earlier: this isn't the `median` but the `mean`.

src/components/CompareResults/ResultsTable.tsx
```diff
@@ -50,8 +51,8 @@ const columnsConfiguration: CompareResultsTableConfig = [
   {
     name: 'New',
     key: 'new',
-
     gridWidth: '1fr',
+    tooltip: 'A summary of all values from New runs using a median.',
```
julienw, line 55: same

julienw:
Thanks, this looks good to me, but this now needs a merge to solve the conflicts just like #826.

julienw (approved):
Thanks!

PR #833, opened 2024-12-19
Fix for select element resizing for framework and revision dropdown

Netacci:
[Bugzilla link](https://bugzilla.mozilla.org/show_bug.cgi?id=1938396)

This PR fixes the `framework` and `revision` dropdown from residing when hovered.

Before
hover

After
hover-fix


Carla-Moz (approved):
Looks good! Thank you!

PR #836, opened 2025-01-06
Removed repetitive style references

Carla-Moz:
To duplicate:

1. Go to perf.compare on Firefox browser (v133.0.3)
2. Switch to dark mode
3. Hover over any repository drop-down
4. Observe the slight expanded size on hover
5. Note, this is only observed on the Home Page

Desired outcome:
There should be no change in size of the select element on hover.

Update: This is only observed on Firefox browser (v133.0.3) not on Nightly or Chrome.

I'd like to see if others see this strange hover effect too. If not, I'll just merge the commit removing the redundant style references. 

Screenshot 2025-01-06 at 13 15 19


julienw (approved):
I don't see the same issue as you, but I don't mind cleaning up the styles. Thanks!

Note that after this change it looks like that `DropDownItems` isn't used anywhere anymore and could be deleted.

Then I also digged down the other exported things from the `DropDownMenu` file: `DropDownItemRaw` and `DropDownMenuRaw`. They're used in `theme/components.js`... but with classes that don't seem to be used anywhere at the moment (`paper-dark` and `paper-light`). So this could be cleaned up as well, if I'm not mistaken. It would be good to do it in this PR or another PR as you prefer.

Carla-Moz:
@julienw, nice catch. You're right `paper-dark/light` aren't used anywhere.

PR #837, opened 2025-01-07
Small adjustments on the sort functionality

julienw:
This PR contains 2 simple commits:
1. For the "delta" column, this uses the absolute value instead of the signed value. Indeed it makes sense that when sorting descending (for example) the high negative values are also shown close to the top. This was an oversight from the initial implementation.
2. When clicking on the sorting button first, the direction is "descending" instead of "ascending", so that the highest values are shown at the top. I believe this is what the users are looking for usually, that's why it makes sense to get this with less clicks. Of course it's still possible to sort "ascending" with another click.

[deploy preview](https://deploy-preview-837--mozilla-perfcompare.netlify.app/subtests-compare-results?baseRev=d140333670bcd3103a668a5ec04ed438bb192368&baseRepo=mozilla-central&newRev=250be4ca3c669db9a397456402f68249aa15d8d5&newRepo=mozilla-central&framework=13&baseParentSignature=92557&newParentSignature=92557)

Tell me what you think @Carla-Moz !

### DIFF
--- a/src/__tests__/CompareResults/SubtestsResultsView.test.tsx
+++ b/src/__tests__/CompareResults/SubtestsResultsView.test.tsx
@@ -209,7 +209,7 @@ describe('SubtestsResultsView Component Tests', () => {
       // Initial view (alphabetical ordered, even if "sort by subtests" isn't specified
       expect(summarizeVisibleRows()).toEqual([
         'dhtml.html: 1.14 %, Low',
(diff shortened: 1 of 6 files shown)


Carla-Moz (approved):
Yes, setting the sort to 'desc' on first click is what I've also understood most people wanted so good change there. Thanks!

julienw:
Thanks!

PR #848, opened 2025-01-28
Make results table title editable

Carla-Moz:
Closes issues [Allow users to change the page title of PerfCompare and have the title persist in the url](https://mozilla-hub.atlassian.net/browse/PCF-540)

[Deploy preview](https://deploy-preview-848--mozilla-perfcompare.netlify.app/compare-results?baseRev=321052fb2ab50a954f9d31d198275891b5fed615&baseRepo=mozilla-central&newRev=7f5fd9b4d3452ee1132204bda40042e760dfeef4&newRepo=mozilla-central&framework=1)

**UPDATED**. Please read before reviewing. The [last commit](https://github.com/mozilla/perfcompare/pull/848/commits/335cff0279290e51db90b1cb6c482262dbb56303) replaces the `form` in `EditTitleInput` with `Form` from` react-router-dom`  and adds a new method on `onSaveSubmit` to better handle the form submission and error case.

In response to first review, please check the following commits:

- [x] [moved the edit title feature down to the results table to preserve the PerfCompare title ](https://github.com/mozilla/perfcompare/pull/848/commits/2745b3874684f451e3f97ebcd898c09a81365511)
- [x] [fixed the cancel button and added corresponding tests ](https://github.com/mozilla/perfcompare/pull/848/commits/c999a13c02dfdb22f6dc0d4f82b3bd97b4ee8805)
- [x] [Moved save and cancel button to the edit title input comp](https://github.com/mozilla/perfcompare/pull/848/commits/f45bcdf57d3b31cb2ff7e3050aa4fa3ea36d67a8)

The following below are handled but this [last commit](https://github.com/mozilla/perfcompare/pull/848/commits/335cff0279290e51db90b1cb6c482262dbb56303):
- [x] Refactored the form and submission in `EditTitleInput`
- [x] Fix truncation of the last letter 
- [x] Allow for updates when the user key presses the enter button 
- [x] Esc key press for cancel
- [x] Moved error handling to edit title input prop
- [x] add remaining tests for coverage

Updated preview shots:

Screenshot 2025-02-03 at 16 45 05

Screenshot 2025-02-03 at 16 45 24

Screenshot 2025-02-03 at 16 46 37

**Outdated:** This is split into 5 commits atm to make it easier to review. I'd like a review now before adding the logic to the other Results View pages
- [x] [Add edit title text and icon on Compare with Base Results page](https://github.com/mozilla/perfcompare/pull/848/commits/81de371fa472bc2cfd73c04eae7b89fab9c4327d)
- [x] [Add the input component](https://github.com/mozilla/perfcompare/pull/848/commits/cbdee6fc96c9d5dfaa29af9775622334233762fa)
- [x] [Add editable title to other results page views](https://github.com/mozilla/perfcompare/pull/848/commits/3e84ec0a8e95079f3e76f920316c363fb11ae641)
- [x] [Add a save button](https://github.com/mozilla/perfcompare/pull/848/commits/c01739a7ecfef02202e05fa10b111432f482a5ab)
- [x] [Add logic to save title in url and show in UI in With Base Results](https://github.com/mozilla/perfcompare/pull/848/commits/c68cf18c37f5b4dd2a507d652f2bdea4e6b628fa)
Note: regarding my last commit, I forgot to put the `titleMessageError` below the input. Will do in next commit. 
- [x] Add test coverage

### DIFF
--- a/src/components/CompareResults/EditTitleInput.tsx
+++ b/src/components/CompareResults/EditTitleInput.tsx
@@ -8,6 +8,7 @@ import TextField from '@mui/material/TextField';
 interface EditTitleInputProps {
   compact: boolean;
   value: string;
+  refInput: React.RefObject<HTMLInputElement>;
   onCancel: () => unknown;
   onChange: (titleText: string) => unknown;
   onSave: () => unknown;
@@ -16,6 +17,7 @@ interface EditTitleInputProps {
 function EditTitleInput({
   compact,
   value,
+  refInput,
   onChange,
   onSave,
   onCancel,
@@ -60,6 +62,7 @@ function EditTitleInput({
         focused
       >
         <TextField
+          inputRef={refInput}
           sx={{ minWidth: '568px' }}
           onFocus={(e) => e.currentTarget.select()}
           placeholder={inputPlaceholder}
--- a/src/components/CompareResults/EditTitleInput.tsx
+++ b/src/components/CompareResults/EditTitleInput.tsx
@@ -1,0 +1,28 @@
+import FormControl from '@mui/material/FormControl';
+import TextField from '@mui/material/TextField';
+
+interface EditTitleInputProps {
+  compact: boolean;
+  onChange: (titleText: string) => unknown;
+}
+
+function EditTitleInput({ compact, onChange }: EditTitleInputProps) {
+  const size = compact ? 'small' : undefined;
+  const inputPlaceholder = 'Write a title for this comparison';
+
+  return (
+    <FormControl fullWidth sx={{ paddingRight: 1 }}>
+      <TextField
+        fullWidth
+        placeholder={inputPlaceholder}
+        inputProps={{ 'aria-label': inputPlaceholder }}
+        id='results-title-input'
+        onChange={(e) => onChange(e.currentTarget.value)}
+        size={size}
+        className='edit-results-title-text-field'
+      />
+    </FormControl>
+  );
+}
+
+export default EditTitleInput;
--- a/src/components/CompareResults/EditTitleInput.tsx
+++ b/src/components/CompareResults/EditTitleInput.tsx
@@ -3,10 +3,17 @@ import TextField from '@mui/material/TextField';
 
 interface EditTitleInputProps {
   compact: boolean;
+  titleError: string | null;
+  comparisonTitleName: string | undefined;
   onChange: (titleText: string) => unknown;
 }
 
-function EditTitleInput({ compact, onChange }: EditTitleInputProps) {
+function EditTitleInput({
+  compact,
+  comparisonTitleName,
+  onChange,
+  titleError,
+}: EditTitleInputProps) {
   const size = compact ? 'small' : undefined;
   const inputPlaceholder = 'Write a title for this comparison';
 
@@ -14,12 +21,19 @@ function EditTitleInput({ compact, onChange }: EditTitleInputProps) {
     <FormControl fullWidth sx={{ paddingRight: 1 }}>
       <TextField
         fullWidth
-        placeholder={inputPlaceholder}
-        inputProps={{ 'aria-label': inputPlaceholder }}
+        placeholder={
+          comparisonTitleName ? comparisonTitleName : inputPlaceholder
+        }
+        inputProps={{
+          'aria-label': comparisonTitleName
+            ? comparisonTitleName
+            : inputPlaceholder,
+        }}
         id='results-title-input'
         onChange={(e) => onChange(e.currentTarget.value)}
         size={size}
         className='edit-results-title-text-field'
+        error={Boolean(titleError)}
       />
     </FormControl>
   );
--- a/src/components/CompareResults/EditTitleInput.tsx
+++ b/src/components/CompareResults/EditTitleInput.tsx
@@ -1,10 +1,9 @@
-import { useEffect, useState, useCallback } from 'react';
+import { useEffect, useState } from 'react';
 
 import Box from '@mui/material/Box';
 import Button from '@mui/material/Button';
 import FormControl from '@mui/material/FormControl';
 import TextField from '@mui/material/TextField';
-import { Form } from 'react-router-dom';
 
 interface EditTitleInputProps {
   compact: boolean;
@@ -26,20 +25,19 @@ function EditTitleInput({
   const [titleError, setTitleError] = useState(false);
   const titleErrorMsg = 'Title cannot be empty';
 
-  const handleEscKeypress = useCallback((e: KeyboardEvent) => {
-    if (e.key === 'Escape') {
-      onCancel();
-    }
-  }, []);
-
   useEffect(() => {
+    const handleEscKeypress = (e: KeyboardEvent) => {
+      if (e.key === 'Escape') {
+        onCancel();
+      }
+    };
+
     document.addEventListener('keydown', handleEscKeypress);
     return () => {
       document.removeEventListener('keydown', handleEscKeypress);
     };
-  }, [handleEscKeypress]);
+  }, [onCancel]);
 
-  //Pr notes: added to better handle form submission and error
   const onSaveSubmit = (e: React.FormEvent) => {
     if (!value.trim()) {
       e.preventDefault();
@@ -51,8 +49,7 @@ function EditTitleInput({
   };
 
   return (
-    //PR notes: changed to Form from react router dom to handle form submission and enter key
-    <Form
+    <form
       className='edit-title-form'
       aria-label='edit results table title'
       onSubmit={onSaveSubmit}
@@ -85,27 +82,15 @@ function EditTitleInput({
           className='edit-title-btns'
           sx={{ display: 'flex', justifyContent: 'flex-end', paddingLeft: 1 }}
         >
-          <Button
-            name='cancel-title'
-            aria-label='cancel title'
-            className='cancel-btn'
-            variant='text'
-            onClick={onCancel}
-          >
+          <Button variant='text' onClick={onCancel}>
             Cancel
           </Button>
-          <Button
-            name='save-title'
-            aria-label='save title'
-            className='save-btn'
-            variant='text'
-            type='submit'
-          >
+          <Button variant='text' type='submit'>
             Save
           </Button>
         </Box>
       </FormControl>
-    </Form>
+    </form>
   );
 }
 
--- a/src/components/CompareResults/EditTitleInput.tsx
+++ b/src/components/CompareResults/EditTitleInput.tsx
@@ -63,7 +63,8 @@ function EditTitleInput({
         focused
       >
         <TextField
-          sx={{ minWidth: '568PX' }}
+          sx={{ minWidth: '568px' }}
+          onFocus={(e) => e.currentTarget.select()}
           placeholder={inputPlaceholder}
           inputProps={{
             'aria-label': inputPlaceholder,
--- a/src/components/CompareResults/EditTitleInput.tsx
+++ b/src/components/CompareResults/EditTitleInput.tsx
@@ -6,7 +6,6 @@ import FormControl from '@mui/material/FormControl';
 import TextField from '@mui/material/TextField';
 
 interface EditTitleInputProps {
-  compact: boolean;
   value: string;
   refInput: React.RefObject<HTMLInputElement>;
   onCancel: () => unknown;
@@ -15,14 +14,12 @@ interface EditTitleInputProps {
 }
 
 function EditTitleInput({
-  compact,
   value,
   refInput,
   onChange,
   onSave,
   onCancel,
 }: EditTitleInputProps) {
-  const size = compact ? 'small' : undefined;
   const inputPlaceholder = 'Write a title for this comparison';
   const [titleError, setTitleError] = useState(false);
   const titleErrorMsg = 'Title cannot be empty';
@@ -41,12 +38,11 @@ function EditTitleInput({
   }, [onCancel]);
 
   const onSaveSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
     if (!value.trim()) {
-      e.preventDefault();
       setTitleError(true);
       return;
     }
-    e.preventDefault();
     onSave();
   };
 
@@ -74,7 +70,7 @@ function EditTitleInput({
             setTitleError(false);
             onChange(e.target.value);
           }}
-          size={size}
+          size='small'
           className='edit-results-title-text-field'
           error={titleError}
           value={value}
--- a/src/components/CompareResults/EditTitleInput.tsx
+++ b/src/components/CompareResults/EditTitleInput.tsx
@@ -1,41 +1,110 @@
+import { useEffect, useState, useCallback } from 'react';
+
+import Box from '@mui/material/Box';
+import Button from '@mui/material/Button';
 import FormControl from '@mui/material/FormControl';
 import TextField from '@mui/material/TextField';
+import { Form } from 'react-router-dom';
 
 interface EditTitleInputProps {
   compact: boolean;
-  titleError: string | null;
-  comparisonTitleName: string | undefined;
+  value: string;
+  handleToggle: () => unknown;
   onChange: (titleText: string) => unknown;
+  onSave: () => unknown;
 }
 
 function EditTitleInput({
   compact,
-  comparisonTitleName,
+  value,
   onChange,
-  titleError,
+  onSave,
+  handleToggle,
 }: EditTitleInputProps) {
   const size = compact ? 'small' : undefined;
   const inputPlaceholder = 'Write a title for this comparison';
+  const [titleError, setTitleError] = useState(false);
+  const titleErrorMsg = 'Title cannot be empty';
+
+  const handleEscKeypress = useCallback((e: KeyboardEvent) => {
+    if (e.key === 'Escape') {
+      handleToggle();
+    }
+  }, []);
+
+  useEffect(() => {
+    document.addEventListener('keydown', handleEscKeypress);
+    return () => {
+      document.removeEventListener('keydown', handleEscKeypress);
+    };
+  }, [handleEscKeypress]);
+
+  //Pr notes: added to better handle form submission and error
+  const onSaveSubmit = (e: React.FormEvent) => {
+    if (!value.trim()) {
+      e.preventDefault();
+      setTitleError(true);
+      return;
+    }
+    e.preventDefault();
+    onSave();
+  };
 
   return (
-    <FormControl fullWidth sx={{ paddingRight: 1 }}>
-      <TextField
+    //PR notes: changed to Form from react router dom to handle form submission and enter key
+    <Form
+      className='edit-title-form'
+      aria-label='edit results table title'
+      onSubmit={onSaveSubmit}
+    >
+      <FormControl
         fullWidth
-        placeholder={
-          comparisonTitleName ? comparisonTitleName : inputPlaceholder
-        }
-        inputProps={{
-          'aria-label': comparisonTitleName
-            ? comparisonTitleName
-            : inputPlaceholder,
-        }}
-        id='results-title-input'
-        onChange={(e) => onChange(e.currentTarget.value)}
-        size={size}
-        className='edit-results-title-text-field'
-        error={Boolean(titleError)}
-      />
-    </FormControl>
+        sx={{ flexDirection: 'row', width: 'auto' }}
+        focused
+      >
+        <TextField
+          sx={{ minWidth: '568PX' }}
+          placeholder={inputPlaceholder}
+          inputProps={{
+            'aria-label': inputPlaceholder,
+          }}
+          id='results-title-input'
+          onChange={(e) => {
+            setTitleError(false);
+            onChange(e.target.value);
+          }}
+          size={size}
+          className='edit-results-title-text-field'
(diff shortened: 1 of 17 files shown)


Carla-Moz:
> thanks for working on it! I couldn't look at the code, but there are a few bugs I could see:
> 
>     1. the URL doesn't have the latest version of the text, it's the text minus one letter
> 
>     2. pressing  doesn't work, one has to go and click "save"
> 
>     3. pressing "cancel" doesn't work (at least when there was a previous text), the text is replaced by the one currently entered, not the previous one
> 
>     4. when editing when we already entered a text before, the previous text is written as a placeholder, it should be a preselected value (see the behavior at the top left of [the profiler UI](https://profiler.firefox.com/public/hqstgh4kre9beta1gq2q68dnp0g3v50q5gprjk0/calltree/?globalTrackOrder=0w3&hiddenGlobalTracks=1w3&hiddenLocalTracksByPid=347182-0wk~345944-0w5~345990-0w3~0-0&profileName=before%20change&thread=0&timelineType=category&v=10))
> 
>     5. The "edit title" seems a bit prominent, I would think that just the iconbutton (with a title of course) would be good enough here. I'd also center it better but for that I'm not so sure.
> 
> 
> Otherwise I'm not sure if the title should be here or rather at the "Results" title part (or next to the results part). But also I think it's pretty fine and I don't have a real good opinion, so if you prefer it there I'm fine with it too.
> 
> Thanks again!

Thanks for the review! My last two commits address concerns 3-5. Cancel button behaves as intended with corresponding test. And I stored the previous text as a value, not placeholder. I agree about moving the edit title feature down to the table. However, I'd like to keep the text "Edit title" to stay consistent with our editing UI re 'Edit entry'

I'll handle the URL and enter key presses in future commits. With that being said, would you like to see the 'save' button removed or should we allow for both to save the new title in the URL? I think both is better.

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -92,18 +98,154 @@ function ResultsMain() {
...
+            onChange={onComparisonTitleChange}
+            value={comparisonTitleName}
+          />
+        ) : (
+          <Grid component='h2' item className={styles.title}>
+            {comparisonTitleName}
+          </Grid>
+        )}
+
+        {renderEditSaveCancelBtns()}
```
julienw, line 234: IMO the save/cancel buttons should go in `EditTitleInput` directly, you'd save a lot of comparisons.

Also if you can make it part of a `<form>`, you'll need to just handle the `submit` event for both the save (button type="submit") and the enter key (automatic), with the added benefit of cleaner semantics.

I'd like to add that it would be good if the "Escape" button would do a "cancel" too. You can use the [`key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) property of the keydown event to check for this.

src/components/Shared/PerfCompareHeader.tsx
```diff
@@ -26,6 +26,7 @@ function PerfCompareHeader({ isHome }: PerfCompareHeaderProps) {
           variant='h1'
           align='center'
           gutterBottom
+          pr={isHome ? '0' : 1}
```
julienw, line 29: what is `pr` again?

julienw:
> I agree about moving the edit title feature down to the table. However, I'd like to keep the text "Edit title" to stay consistent with our editing UI re 'Edit entry'

If it doesn't stay at the top, I think "Edit title" would be fine.
I think it was too visible and present at the top for its usefulness, but if we make it less prominent then it's OK.

src/components/Shared/PerfCompareHeader.tsx
Carla-Moz: paddingRight

julienw:
I didn't have the time to look at the code, but from the deploy preview I like it much more than before, good work!.

A couple comments:
1. when starting the edition with the button, can you make it so that the text is fully selected and the input is focused? You should be able to do it with `useEffect` and a reference to the input (`input.focus(); input.select();`);
2. also I noticed that the stylized underline for the main "perfcompare" title at the top seems incorrectly moved to the right.

Now or as a follow-up:
* Change window.title to account for the new title, so that the users can find it in their history more easily (which is IMO a goal)

As a follow-up, once the implementation has been approved as well:
* Do the same thing for the subtests page

Carla-Moz:
Ahhhh, @julienw, unfortunately I rebased _then_ squashed my commits. I apologize. Hopefully, it's not too hard to review. :-/

Carla-Moz:
#Now or as a follow-up:

    Change window.title to account for the new title, so that the users can find it in their history more easily (which is IMO a goal)

As a follow-up, once the implementation has been approved as well:

    Do the same thing for the subtests page

I'll take care of the above in follow-up PRs, @julienw

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -91,18 +96,106 @@ function ResultsMain() {
...
+    updateRawSearchParams(rawSearchParams);
+    showEditComparisonTitle(!editComparisonTitleInputVisible);
+  };
+
+  const deSlugify = (slug: string | undefined) => {
+    if (!slug) return '';
+    return slug.replace(/-/g, ' ').replace(/^\w/, (char) => char.toUpperCase());
+  };
+
+  const deSlugifiedTitle = deSlugify(comparisonTitleName);
```
julienw, line 151: just use the original title coming from the URL :-)

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -91,18 +96,106 @@ function ResultsMain() {
...
+            ) : (
+              <>
+                <Grid component='h2' item className={styles.title}>
+                  {deSlugifiedTitle}
+                </Grid>
+                <Button
+                  name='edit-title'
+                  aria-label='edit title'
+                  startIcon={buttonIcon}
+                  className='edit-title-btn'
```
julienw, line 184: same as above, you don't need name/aria-label/className here

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -91,18 +96,106 @@ function ResultsMain() {
...
+  const initialComparisonTitle = rawSearchParams.get('title') ?? 'Results';
+  const [comparisonTitleName, setComparisonTitleName] = useState(
+    initialComparisonTitle,
+  );
+  const [previousComparisonTitle, setPreviousComparisonTitle] =
+    useState('Results');
+  const [editComparisonTitleInputVisible, showEditComparisonTitle] =
+    useState(false);
+
+  const handleEditInputToggle = () => {
```
julienw, line 110: It would be easier to have 2 different functions, because it's used for 2 different things:
* display when not displayed -> `onEditButtonClick` (for example)
* canceled -> `onEditCancel` (for example)

(when it's saved it's another function, which adds to the confusion IMO)

This would remove the if condition, make the code more linear and easier to follow.

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -91,18 +96,106 @@ function ResultsMain() {
...
   }
 
+  /********** Edit Results Title Section **********/
+  const [rawSearchParams, updateRawSearchParams] = useRawSearchParams();
+  const initialComparisonTitle = rawSearchParams.get('title') ?? 'Results';
+  const [comparisonTitleName, setComparisonTitleName] = useState(
+    initialComparisonTitle,
+  );
+  const [previousComparisonTitle, setPreviousComparisonTitle] =
+    useState('Results');
```
julienw, line 106: Usually if a state isn't used for the rendering, it's a good sign it shouldn't be a state. Indeed changing a state triggers a rerender, but that's not needed here. Therefore this should be a ref instead of a state.
Think of a `ref` just like a variable attached to this component instance.
More info about it: https://react.dev/reference/react/useRef

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -91,18 +96,106 @@ function ResultsMain() {
...
+        .toLowerCase()
+        .replace(/[^a-z0-9]+/g, '-') // Convert non-alphanumeric to hyphen
+        .replace(/^-+|-+$/g, '') // Trim hyphens from start and end
+    );
+  };
+
+  const onComparisonTitleChange = (value: string) => {
+    //PR notes: fix for truncated letter at the end of entry
+    const slug = slugifyComparisonTitle(value);
+    rawSearchParams.set('title', slug);
```
julienw, line 137: I would remove it, and just set it in `OnComparisonTitleSave` directly, someting like this:
```js
rawSearchParams.set("title", comparisonTitleName)
updateRawSearchParams(rawSearchParams)
```

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -91,18 +96,106 @@ function ResultsMain() {
...
+    setComparisonTitleName(value);
+  };
+
+  const OnComparisonTitleSave = () => {
```
julienw, line 141: nit: please start with a small letter

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -91,18 +96,106 @@ function ResultsMain() {
...
+  };
+
+  const deSlugifiedTitle = deSlugify(comparisonTitleName);
+
+  const buttonIcon = (
+    <img
+      id='edit-title-icon'
+      className='icon icon-edit'
+      src={themeMode === 'light' ? pencil.toString() : pencilDark.toString()}
+      alt='edit-title-icon'
```
julienw, line 158: nit: this alt is useless IMO
rather add `role="presentation"` for this presentational image

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -91,18 +96,106 @@ function ResultsMain() {
...
+  const deSlugify = (slug: string | undefined) => {
+    if (!slug) return '';
+    return slug.replace(/-/g, ' ').replace(/^\w/, (char) => char.toUpperCase());
+  };
+
+  const deSlugifiedTitle = deSlugify(comparisonTitleName);
+
+  const buttonIcon = (
+    <img
+      id='edit-title-icon'
```
julienw, line 155: is the id useful?

src/components/CompareResults/EditTitleInput.tsx
```diff
@@ -0,0 +1,112 @@
...
+import Box from '@mui/material/Box';
+import Button from '@mui/material/Button';
+import FormControl from '@mui/material/FormControl';
+import TextField from '@mui/material/TextField';
+import { Form } from 'react-router-dom';
+
+interface EditTitleInputProps {
+  compact: boolean;
+  value: string;
+  handleToggle: () => unknown;
```
julienw, line 12: It might be more readable and easier to reason about to call this `onCancel` (which it is :-) ).

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -91,18 +96,106 @@ function ResultsMain() {
...
+  const onComparisonTitleChange = (value: string) => {
+    //PR notes: fix for truncated letter at the end of entry
+    const slug = slugifyComparisonTitle(value);
+    rawSearchParams.set('title', slug);
+    setComparisonTitleName(value);
+  };
+
+  const OnComparisonTitleSave = () => {
+    updateRawSearchParams(rawSearchParams);
+    showEditComparisonTitle(!editComparisonTitleInputVisible);
```
julienw, line 143: nit: just use `false` directly

julienw (changes requested):
Thanks, the behavior looks better!

I left a few small comments, tell me what you think.
Especially I don't understand the dance with the `slug`.

Also, I still don't see this:

> when starting the edition with the button, can you make it so that the text is fully selected and the input is focused? You should be able to do it with useEffect and a reference to the input (`input.focus(); input.select();`).

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -91,18 +96,106 @@ function ResultsMain() {
...
+    return (
+      title
+        /**
+         *please see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize#examples
+         */
+        .normalize('NFD') // Normalize to decompose diacritics (e.g., é -> e)
+        .replace(/[\u0300-\u036f]/g, '') // Remove accents
+        .toLowerCase()
+        .replace(/[^a-z0-9]+/g, '-') // Convert non-alphanumeric to hyphen
+        .replace(/^-+|-+$/g, '') // Trim hyphens from start and end
```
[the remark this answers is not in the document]
Carla-Moz: This is for the title in the url not the UI: https://developers.google.com/search/docs/crawling-indexing/url-structure

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: @julienw I can't in this case because it's slugified in the URL. Let me show you what I mean.

If I create a new title called `Test Title` in the UI, the title in Search Params will be slugified to produce > `test-title`

```
const onComparisonTitleChange = (value: string) => {
    const slug = slugifyComparisonTitle(value); // outputs to test-title
    rawSearchParams.set('title', slug); 
    setComparisonTitleName(value);
  };

```

If I use the title from the URL, it would show `test-title` in the UI not `Test Title`. That's why I deslugify it. I think it's a good idea to slugify the title because people might have spaces, special symbols, colons, or accents, or who knows what. This way the title is standardized in the url - always lower case with hyphens to replace spaces. What do you think? Also, let me know if I'm misunderstanding you.

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: Looks like I've been mixing aria-lable with aria-description :/ I'll make note of this for the future.

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: Delete this whole function?

Carla-Moz: Ok, I see what you mean.

src/components/CompareResults/ResultsMain.tsx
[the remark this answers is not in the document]
Carla-Moz: Okay, I see that URLSearchParams does do that. I'll delete.

src/components/CompareResults/ResultsMain.tsx
Carla-Moz: Okay, I see what you mean. I see that

src/components/CompareResults/EditTitleInput.tsx
```diff
@@ -0,0 +1,112 @@
...
+          FormHelperTextProps={{ sx: { padding: 0, margin: 0 } }}
+        />
+        <Box
+          className='edit-title-btns'
+          sx={{ display: 'flex', justifyContent: 'flex-end', paddingLeft: 1 }}
+        >
+          <Button
+            name='cancel-title'
+            aria-label='cancel title'
+            className='cancel-btn'
```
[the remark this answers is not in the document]
Carla-Moz: Why don't we want to have a name for our buttons? Doesn't it make it easier to find for tests?

src/components/CompareResults/EditTitleInput.tsx
```diff
@@ -0,0 +1,112 @@
...
+      setTitleError(true);
+      return;
+    }
+    e.preventDefault();
+    onSave();
+  };
+
+  return (
+    //PR notes: changed to Form from react router dom to handle form submission and enter key
+    <Form
```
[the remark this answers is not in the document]
Carla-Moz: Hmm, true.

Carla-Moz:
> Also, I still don't see this:
> 
>     when starting the edition with the button, can you make it so that the text is fully selected and the input is focused? You should be able to do it with useEffect and a reference to the input (input.focus(); input.select();).
> 

Doesn't this line solve it? https://github.com/Carla-Moz/perfcompare/blob/3e5025b4e8747221bd4ddd1b8cf1fb3cb0396ba5/src/components/CompareResults/EditTitleInput.tsx#L67

julienw:
> ```
> Also, I still don't see this:
> when starting the edition with the button, can you make it so that the text is fully selected and the input is focused? 
> You should be able to do it with useEffect and a reference to the input (input.focus(); input.select();).
> ```
> 
> @julienw Doesn't this line solve making sure the input is focused and the text is selected? https://github.com/Carla-Moz/perfcompare/blob/3e5025b4e8747221bd4ddd1b8cf1fb3cb0396ba5/src/components/CompareResults/EditTitleInput.tsx#L67

This makes it so that the input is selected _when_ it's focused (and it's good!), but not that it's focused when the user presses "edit". This makes that the user needs an extra click to edit the title, which we can avoid.

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -57,11 +60,13 @@ function ResultsMain() {
...
-      gap: '9px',
-      margin: `0 0 ${Spacing.Medium}px 0`,
-    }),
+  };
+
+  const titleContainerSx = {
+    alignItems: 'center',
+    gap: '9px',
+    margin: `0 0 ${Spacing.Medium}px 0`,
+    //PR notes: removed the maxWidth sinec it interfered with styles
```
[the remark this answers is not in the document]
julienw: It's common that I add comments on github directly, and then I think "mmm this would be better directly in the code". 

We need to find the right balance between:
"this is useful for future reader, because it explains why we're doing something" => comments in the code
"this is useful only for the PR reviewer, because it explains the reasoning behind a change" => comments in the PR

Thanks!

src/components/CompareResults/ResultsMain.tsx
julienw: yeah, `aria-label` is a way to set an "accessible name", but it's already set by computation here thanks to the button text :-) the devtools accessibility panel can be very helpful to look at that computation.

Remember the first rule of aria: do not use aria if you can avoid it :-)

src/components/CompareResults/ResultsMain.tsx
julienw: sorry, yes, it was a comment just for this only line, I always forget that github doesn't show on which lines a comment applies despite I selected it or them before...

Carla-Moz:
> > ```
> > Also, I still don't see this:
> > when starting the edition with the button, can you make it so that the text is fully selected and the input is focused? 
> > You should be able to do it with useEffect and a reference to the input (input.focus(); input.select();).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @julienw Doesn't this line solve making sure the input is focused and the text is selected? https://github.com/Carla-Moz/perfcompare/blob/3e5025b4e8747221bd4ddd1b8cf1fb3cb0396ba5/src/components/CompareResults/EditTitleInput.tsx#L67
> 
> This makes it so that the input is selected _when_ it's focused (and it's good!), but not that it's focused when the user presses "edit". This makes that the user needs an extra click to edit the title, which we can avoid.

Not to be annoying, why is it necessary to have the input focused when the user presses the `Edit button` vs when the user hovers over, clicks on, or starts typing in the Edit input?  In terms of accessibility, we already have an aria label that lets users know this input is for editing the comparison title.

Carla-Moz:
> Think aboug the whole process of editing the title: When using the mouse to click the edit button, this is one less click. When using the keyboard to click the edit button, this is also much more easier to use as it doesn't require a backwards tabulation.
> 
> Note that I'm not saying to remove your "select on focus" behavior ! Only to add a "focus on edit" behavior. Is there a reason you don't like it? I'd like to understand :-)
> 
> Thanks

I suppose.  I'll go ahead and add the focus on edit click but I'd like to talk about it when you return lol.

src/components/CompareResults/ResultsMain.tsx
```diff
@@ -90,19 +94,87 @@ function ResultsMain() {
...
   }
+  /************************************************/
+  /********** Edit Results Title Section **********/
+  /************************************************/
+  const [rawSearchParams, updateRawSearchParams] = useRawSearchParams();
+  const initialComparisonTitle = rawSearchParams.get('title') ?? 'Results';
+  const [comparisonTitleName, setComparisonTitleName] = useState(
+    initialComparisonTitle,
+  );
+  const previousComparisonTitleRef = useRef('Results');
```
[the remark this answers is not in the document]
julienw: Mmm actually the initial value isn't important because it's set later in `onEditButtonClick`.

Therefore I'd do something like that (if you want):
```suggestion
  // The initial value isn't important because it's set later in `onEditButtonClick`
  const previousComparisonTitleRef = useRef('');
```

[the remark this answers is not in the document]
Carla-Moz: I'll just set it to initialComparisonTitle so we don't need the comment.

src/components/CompareResults/EditTitleInput.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+function EditTitleInput({
+  compact,
+  value,
+  refInput,
+  onChange,
+  onSave,
+  onCancel,
+}: EditTitleInputProps) {
+  const size = compact ? 'small' : undefined;
```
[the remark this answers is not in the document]
Carla-Moz: Oh yeah, that makes sense lol. Good catch!

PR #849, opened 2025-02-03
Implement loading spinner for Subtests Results Page 

sumairq:
# Add Loading Spinner for Subtests Results  
[Bug 1917858](https://bugzilla.mozilla.org/show_bug.cgi?id=1917858) 

## Description  
This PR introduces a loading state for the `SubtestsResultsMain` component, improving the user experience while results are being fetched. Previously, the component relied on the raw promise resolution, which could lead to a blank screen before data was available. Now, a `Skeleton` loader and a loading indicator are displayed while the results are being fetched.  

## Changes  
- Implemented `useState` hooks for `isLoading` and `error` states.  
- Wrapped the results promise with `.then()` and `.catch()` to handle loading and error states properly within a useEffect().  
- Displayed a `Skeleton` loader and UI placeholders while data is being retrieved.
- Made the `result` prop optional for the retrigger button so the button can be displayed as a placeholder even when the results are not available.
- Ensured proper cleanup using `isMounted` to avoid state updates on unmounted components.  
- Maintained existing search functionality and table rendering.  

## Testing  
- Verified that the spinner appears while waiting for results.  
- Checked that data loads correctly once available.  
- Ensured proper error handling and UI fallback in case of failures.  

## Screenshots  

| Spinner in action|
|--------|
| header-only-skeleton2 |

### DIFF
--- a/src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
+++ b/src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
@@ -1,16 +1,15 @@
-import { useState } from 'react';
+import { useState, Suspense } from 'react';
 
-import { Grid, Link } from '@mui/material';
+import { Grid, Skeleton, Stack, Link } from '@mui/material';
 import Alert from '@mui/material/Alert';
 import { Container } from '@mui/system';
-import { useLoaderData } from 'react-router-dom';
+import { useLoaderData, Await } from 'react-router-dom';
 import { style } from 'typestyle';
 
 import SubtestsBreadcrumbs from './SubtestsBreadcrumbs';
 import SubtestsResultsTable from './SubtestsResultsTable';
 import SubtestsRevisionHeader from './SubtestsRevisionHeader';
 import DownloadButton from '.././DownloadButton';
-import NoResultsFound from '.././NoResultsFound';
 import SearchInput from '.././SearchInput';
 import { subtestsView, subtestsOverTimeView } from '../../../common/constants';
 import { useAppSelector } from '../../../hooks/app';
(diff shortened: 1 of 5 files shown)


julienw:
The behavior looks good to me.
@Carla-Moz would you be able to look at the code? I did a quick overview and this looks good, but I'd like more scrutiny from you :-) thanks

src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
```diff
@@ -67,32 +87,71 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
...
+            <>
+              <header>
+                <SubtestsBreadcrumbs view={view} />
+
+                <Alert severity='info' className={styles.title}>
+                  A Perfherder link is available for the same results
+                  {'.'}
+                </Alert>
+                <Stack spacing={1} sx={{ marginBottom: '12px' }}>
+                  <Skeleton
```
julienw, line 104: the color is a little bit dark, I wonder if we can make it somewhat lighter? But no need to think too much

julienw:
Hey Sumair, thanks for this work!
I'm surprised because this doesn't use `<Suspense>` and `<Await>` as it's done in the main results view, as you can see in https://github.com/mozilla/perfcompare/blob/0be94afab8a1a1bb9fc53ac87de081b0122fa4e6/src/components/CompareResults/ResultsTable.tsx#L213-L235. These utilities make it a lot easier to handle and maintain rather than managing some states manually ourselves.
`Suspense` is coming with `React` while `Await` is coming with `react-router`.

It's also good to mention that React 19 also supports [`use`](https://react.dev/reference/react/use) for a similar use case, but we don't have React 19 in this project yet.

Can you please change this PR to use `Suspense` and `Await`, by taking the other components as an example?
Thanks!

julienw:
just a note that I started looking at it but haven't finished, sorry. Will try to finish it by tomorrow.

src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
```diff
@@ -68,102 +64,107 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
...
-            <DownloadButton resultsPromise={[results]} />
-          </Grid>
-          <Grid item xs='auto'>
-            <RetriggerButton result={results[0]} variant='text' />
-          </Grid>
-        </Grid>
+
+        <Suspense
+          fallback={
+            <div>
```
julienw, line 74: We don't need the `div` just use a fragment `<></>`.

src/utils/subtestsUtils.ts
```diff
@@ -0,0 +1,114 @@
...
+
+type SubtestsResults = {
+  key: string;
+  // By construction, there should be only one item in the array. But if more
+  // than one subtests share the same name, then there will be more than one item.
+  // Can this happen? We're not sure.
+  value: CompareResultsItem[];
+};
+
+export function getSubtestsHeaderAndPerfherderURL(
```
julienw, line 20: There's no need to have just one function for this. I think it would be better to have 2: one for the header, one for the URL. As you'll see in another comment, we shouldn't use the header to compute the URL anyway.

src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
```diff
@@ -68,102 +64,107 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
...
+                  <SearchInput
+                    defaultValue={initialSearchTerm}
+                    onChange={onSearchTermChange}
+                  />
+                </Grid>
+                <Grid item xs='auto'>
+                  <DownloadButton resultsPromise={[]} />
+                </Grid>
+                <Grid item xs='auto'>
+                  <RetriggerButton variant='text' />
```
julienw, line 106: as said above, just use a normal disabled button

src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
```diff
@@ -68,102 +64,107 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
...
+              </Stack>
+              <Grid container spacing={1}>
+                <Grid item xs={12} md={6} sx={{ marginInlineEnd: 'auto' }}>
+                  <SearchInput
+                    defaultValue={initialSearchTerm}
+                    onChange={onSearchTermChange}
+                  />
+                </Grid>
+                <Grid item xs='auto'>
+                  <DownloadButton resultsPromise={[]} />
```
julienw, line 103: use a normal disabled button

src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
```diff
@@ -68,102 +64,107 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
...
+                <Grid item xs='auto'>
+                  <RetriggerButton variant='text' />
+                </Grid>
+              </Grid>
+            </div>
+          }
+        >
+          <Await resolve={results}>
+            {(loadedResults: CompareResultsItem[]) => {
+              {
```
julienw, line 114: why 2 blocks??

julienw: It wuold be better to extract this component, along with all the related logic, to another component (in the same file).

I think the new component could be called something such as `SubtestsResultsMainImpl` or `SubtestsResultsMainInternal`.

julienw (changes requested):
Thanks,
I think that in terms of behavior this looks pretty good! This is just like I believe it should behave.

I'm suggesting here some code changes to make the code easier to read, follow, and maintain. 
I think the biggest change I'm asking would be how we compute the perfherder URL. Otherwise this is mostly some moving code around :-)

Tell me what you think!

src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
```diff
@@ -145,25 +134,75 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
...
+        <Suspense
+          fallback={
+            <>
+              <Stack spacing={1} sx={{ marginBottom: '12px' }}>
+                <Skeleton
+                  variant='rounded'
+                  sx={{
+                    // This fontSize value makes the skeleton match the height of the replaced element.
+                    fontSize: '2.18em',
+                    backgroundColor: secondary.main,
```
julienw, line 147: just use "secondary.main" as a string, then material UI knows how to do it inrternally :-)

src/components/CompareResults/SubtestsResults/SubtestsResultsMain.tsx
```diff
@@ -68,74 +118,8 @@ function SubtestsResultsMain({ view }: SubtestsResultsMainProps) {
...
-  }
-
   return (
-    <Container
-      maxWidth={false}
-      sx={{ maxWidth: '1300px' }}
-      className={styles.container}
-      data-testid='subtests-main'
-    >
+    <Container className={styles.container} data-testid='subtests-main'>
```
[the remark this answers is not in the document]
julienw: if you can do a PR with just this change, that would be great :) so that I can then deploy the new changes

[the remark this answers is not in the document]
sumairq: https://github.com/mozilla/perfcompare/pull/891

PR #853, opened 2025-02-06
Add tooltip to subtest platform name

Netacci:
[Bugzilla number](https://bugzilla.mozilla.org/show_bug.cgi?id=1946315)

[Deploy link]()

This PR adds a tooltip to the subtests' platform name. Similar to what was done for the results view here [PR](https://github.com/mozilla/perfcompare/pull/832)

SDD

SDD

### DIFF
--- a/src/components/CompareResults/SubtestsResults/SubtestsRevisionHeader.tsx
+++ b/src/components/CompareResults/SubtestsResults/SubtestsRevisionHeader.tsx
@@ -78,6 +78,7 @@ function getSuite(
   docsURL: string,
   isLinkSupported: boolean,
 ) {
+  const suiteLink = Strings.components.revisionRow.title.suiteLink;
   if (isLinkSupported) {
     return (
       <>
@@ -86,6 +87,7 @@ function getSuite(
           underline='hover'
           target='_blank'
           href={docsURL}
+          title={suiteLink}
         >
           {header.suite}
         </Link>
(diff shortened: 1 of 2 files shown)


Carla-Moz (approved):
LGTM! Honestly, I think it's better when a tooltip is placed above the element so as not to be blocked by the cursor, esp if the element is small and the cursor is bigger than normal.

Netacci:
> LGTM! Honestly, I think it's better when a tooltip is placed above the element so as not to be blocked by the cursor, esp if the element is small and the cursor is bigger than normal. 

Yea I see that. I wanted to remain consistent with the way it was done on the revisions page and also with the hash numbers. But maybe another pr can be opened to ammend all of them.

Carla-Moz:
@Netacci you can update this and I'll merge it.

julienw:
Hey, maybe I wasn't clear in the initial bug :) but what I meant initially is adding a tooltip on "Linux 18.04", to show the full platform name. Just like the left cell in the table in the main results table.

Netacci:
> Hey, maybe I wasn't clear in the initial bug :) but what I meant initially is adding a tooltip on "Linux 18.04", to show the full platform name. Just like the left cell in the table in the main results table.

Ohh okay. I guess I didn't interpret it that way. I'll add that to this pr too.

src/components/CompareResults/SubtestsResults/SubtestsRevisionHeader.tsx
```diff
@@ -156,8 +157,13 @@ function SubtestsRevisionHeader(props: SubtestsRevisionHeaderProps) {
         <strong>{getSuite(header, docsURL, isLinkSupported)}</strong> |
         {baseInfo}
         {getRevLink(header.new_rev, header.new_repo, '- New')} | {framework} |{' '}
-        {platformIcon}
-        <span>{getPlatformAndVersion(header.platform)}</span> |{' '}
+        <Tooltip placement='bottom' title={header.platform} arrow>
+          <span>
+            {platformIcon}
+            <span>{getPlatformAndVersion(header.platform)}</span>
```
Carla-Moz, line 163: Can you please add styles to set the cursor to pointer for the platform header? When I hover over it, the cursor is text which is misleading.

Carla-Moz (commented):
Thanks! Just one small change needed.

src/components/CompareResults/SubtestsResults/SubtestsRevisionHeader.tsx
Netacci: Okay will also do same for the main results table

Carla-Moz (approved):
Thanks!

PR #854, opened 2025-02-06
add zindex to filter menu

Netacci:
[Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1946046)
[Deploy link](https://deploy-preview-854--mozilla-perfcompare.netlify.app/compare-results?baseRev=731d4b2dcdc5955ba826a55077d5d5ba4b8e1ea8&baseRepo=mozilla-central&newRev=5041c20ccabb22016a82288a40bc13d845dfe2fe&newRepo=mozilla-central&framework=1&filter_confidence=high)

This fixes the issue of tooltip overlapping the confidence filter menu when it's open.

I added a zindex to the filter menu so that it stacks higher than the tooltip. This would also work for other filters if we decide to add a tooltip to them later.

Update: When the filter menu is open, the tooltip still pops up. This happens because the tooltip is wrapping the whole `FiterableColumnHeader` component which includes the filter menu. I had to remove the tooltip from wrapping the whole component to just wrapping only the button. This solves the issue and also fixes the overlay issue without needing a `zindex`

Before

be

After

https://github.com/user-attachments/assets/1ec2fab8-4c96-4616-8e48-5ab1a36d9083

### DIFF
--- a/src/components/CompareResults/TableHeader.tsx
+++ b/src/components/CompareResults/TableHeader.tsx
@@ -134,6 +134,7 @@ function FilterableColumnHeader({
         </Box>
         <KeyboardArrowDownIcon />
       </Button>
+      {/* zindex of filter menu has to be greater than tooltip's zindex which is 1500 */}
       <Menu sx={{ zIndex: 2000 }} {...bindMenu(popupState)}>
         <MenuItem dense={true} onClick={onClearFilter}>
           Select all values
--- a/src/components/CompareResults/TableHeader.tsx
+++ b/src/components/CompareResults/TableHeader.tsx
@@ -134,7 +134,7 @@ function FilterableColumnHeader({
         </Box>
         <KeyboardArrowDownIcon />
       </Button>
-      <Menu {...bindMenu(popupState)}>
+      <Menu sx={{ zIndex: 2000 }} {...bindMenu(popupState)}>
         <MenuItem dense={true} onClick={onClearFilter}>
           Select all values
         </MenuItem>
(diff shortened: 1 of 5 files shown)


Carla-Moz:
@Netacci thanks for this. Just one observation. The tooltip stays open even though the mouse is away from the "Confidence" name. Can you please make sure the tooltip immediately disappears when the mouse is away from it like the other tooltips?

Screenshot 2025-02-10 at 17 04 18

src/components/CompareResults/TableHeader.tsx
```diff
@@ -134,7 +134,7 @@ function FilterableColumnHeader({
         </Box>
         <KeyboardArrowDownIcon />
       </Button>
-      <Menu {...bindMenu(popupState)}>
+      <Menu sx={{ zIndex: 2000 }} {...bindMenu(popupState)}>
```
Carla-Moz, line 138: Hmm, is this the only place we need this because I noticed the tooltips aren't in the subtests table view either. It might help to add this to the Menu MUI overrides in the component file

Carla-Moz (changes requested):

Carla-Moz (changes requested):
Please see my comment above. Thanks!

Netacci:
> @Netacci thanks for this. Just one observation. The tooltip stays open even though the mouse is away from the "Confidence" name. Sometimes it triggers open even if the user isn't hovering over the "Confidence" and is trying to select an option from the dropdown menu. Can you please make sure the tooltip immediately disappears when the mouse is away from it like the other tooltips?
> 
> Screenshot 2025-02-10 at 17 04 18

Yes you are right. This is happening because tooltip is wrapping the whole `FilterableColumnHeader` component which includes the filter menu, so when we hover around the menu list it prompts the tooltip to reappear. To fix this, i've taken a different approach to render tooltip for the filterable columns. I've made tooltip to wrap only the button. This also fixes the overlay issue and therefore we don't need `zindex` anymore. I'll push this so you can see.

Carla-Moz (commented):
LGTM! Thanks for cleaning this up

Carla-Moz (approved):

PR #844, opened 2025-01-23
PCF-561 When there are a lot of replicates, the expanded row with all the runs is too long

esanuandra:
Jira: PCF-561

[localhost test link results page](http://localhost:3000/compare-over-time-results?baseRepo=try&selectedTimeRange=1209600&newRev=301fd7df21a146720ac180a15dbfd007b090dcaf&newRepo=try&framework=1)
[localhost test link subtests page](http://localhost:3000/subtests-compare-over-time-results?baseRepo=try&newRev=301fd7df21a146720ac180a15dbfd007b090dcaf&newRepo=try&framework=1&selectedTimeRange=1209600&baseParentSignature=4764992&newParentSignature=4764992)
[deploy test link](https://deploy-preview-844--mozilla-perfcompare.netlify.app/compare-over-time-results?baseRepo=try&selectedTimeRange=1209600&newRev=301fd7df21a146720ac180a15dbfd007b090dcaf&newRepo=try&framework=1)

PR #859, opened 2025-02-12
Follow up PCF-561 Add copy run values to extended row

esanuandra:
This is a follow up PR for #844 
It adds the `Copy results` button inside the extended row.

[deploy test link](https://deploy-preview-859--mozilla-perfcompare.netlify.app/compare-over-time-results?baseRepo=try&selectedTimeRange=1209600&newRev=301fd7df21a146720ac180a15dbfd007b090dcaf&newRepo=try&framework=1) 

Screenshot 2025-02-12 at 17 07 19

### DIFF
--- a/src/__tests__/CompareResults/RevisionRow.test.tsx
+++ b/src/__tests__/CompareResults/RevisionRow.test.tsx
@@ -115,4 +115,29 @@ describe('Expanded row', () => {
 
     expect(showLessButton).toBeInTheDocument();
   });
+
+  it('should copy run values when "Copy values" is clicked', async () => {
+    const writeTextMock = jest
+      .spyOn(navigator.clipboard, 'writeText')
+      .mockResolvedValue();
+    const user = userEvent.setup({ delay: null });
+    const { testCompareData } = getTestData();
+    const baseRuns = testCompareData[0].base_runs.toString();
+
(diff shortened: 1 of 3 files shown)


src/__tests__/CompareResults/RevisionRow.test.tsx
```diff
@@ -115,4 +115,29 @@ describe('Expanded row', () => {
 
     expect(showLessButton).toBeInTheDocument();
   });
+
+  it('should copy run values when "Copy values" is clicked', async () => {
```
julienw, line 119: thanks for the test!

src/components/CompareResults/RunValues.tsx
```diff
@@ -50,6 +50,10 @@ function RunValues(props: RunValuesProps) {
     setExpanded(!expanded);
   };
 
+  const onCopyValues = () => {
+    void navigator.clipboard.writeText(values.toString());
```
julienw, line 54: I'm not fully sure of what we should copy. I was initially thinking separated by spaces (like how it is displayed) would have been better, but trying it with google spreadsheet convinced me that commas are a better option. So it's good to me!

julienw (approved):
looks good to me, thanks!

PR #860, opened 2025-02-14
Fix incorrectly rendered platform on results table

Netacci:
[Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1943334)
[Deploy link](http://localhost:3000/compare-results?baseRev=0874fcba521fcf935ace34a3507f284de9bc4d79&baseRepo=autoland&newRev=e4734cd9848c0b570af98a9708da8fe4a2263b10&newRepo=autoland&framework=2&search=installer&filter_platform=windows)

This PR fixes incorrectly rendered platform names on the results table

Before

When macOS is filtered, we can still see windows as part of the result. [Live link](https://perf.compare/compare-results?baseRev=0874fcba521fcf935ace34a3507f284de9bc4d79&baseRepo=autoland&newRev=e4734cd9848c0b570af98a9708da8fe4a2263b10&newRepo=autoland&framework=2&search=installer&filter_platform=osx)

macos

After

The platform name for Windows is `windows` so this wasn't capturing some windows platform with platform name as `win64-nightlyasrelease` 

To fix this I updated the `getPlatformShortName` to return Windows if platform name starts with `win` This way we cover both `win` and `windows` platforms.

I also updated `OS X` to `macOS` on the results table. This was updated on the `osMapping` object.

download (19)

### DIFF
--- a/src/__tests__/CompareResults/ResultsTable.test.tsx
+++ b/src/__tests__/CompareResults/ResultsTable.test.tsx
@@ -164,9 +164,9 @@ describe('Results Table', () => {
     expect(summarizeVisibleRows()).toEqual([
       'a11yr dhtml.html opt e10s fission stylo webrender',
       '  rev: spam',
-      '  - OS X 10.15, Improvement, 1.08 %, Low',
+      '  - macOS 10.15, Improvement, 1.08 %, Low',
       '  rev: devilrabbit',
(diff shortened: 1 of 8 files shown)


Carla-Moz (approved):
LGTM, thanks!

PR #861, opened 2025-02-20
Update the project to node v22

julienw:
I'm updating all the projects I work on to use node v22 instead of 18 or 20.
In this project this mostly means that some test snapshots had to be regenerated, because the output of some locale-dependent strings changed with the node update.

But otherwise everything seems to work fine. Locally I tried:
* the development server
* building the app

I also checked the netlify deploy preview because this is the closest of our production environment.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ orbs:
 executors:
   node:
     docker:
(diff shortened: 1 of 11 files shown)


Carla-Moz (approved):
This works for me locally too when I tested, building as well. Thanks.

PR #864, opened 2025-02-25
Add browser name to result header

Netacci:
[Bugzilla number](https://bugzilla.mozilla.org/show_bug.cgi?id=1937794)

This PR adds browser name to results header. 
When going through compare results , it's hard to know which browser is being compared. The only way to know is to expand the row. 

Before
download (26)

After
I created a function that checks base and new apps and returns the browser name when they both aren't `firefox` or `fenix`. 

[Different browser comparison](https://deploy-preview-864--mozilla-perfcompare.netlify.app/compare-results?baseRev=804f06781c29e65269decaa33e0ddf60f2e5da46&baseRepo=mozilla-central&newRev=f0f8a9427002ddea9023cc4ab3f4ca0c7c3d9947&newRepo=try&framework=13&filter_status=improvement%2Cregression)

download (24)

When both `base_app` and `new_app` are `firefox` or `fenix`, nothing gets displayed

[Same browser comparison](https://deploy-preview-864--mozilla-perfcompare.netlify.app/compare-results?baseRev=632d1b46eeba0e6950cb84fd950eceb086142d23&baseRepo=try&newRev=7c9e39d87387c3ed4c829ad5300ceb13705026e0&newRepo=try&framework=12)

When the row gets expanded, the browser name is removed from the header since we already have them on the expanded row.

download (25)


Netacci:
After rebasing, this tests keep failing here even though they all pass on local. I'll close the PR and try to open a new branch with same solution

Untitled

PR #865, opened 2025-02-27
Update Search to use `search` and `hash` 

Netacci:
[Bugzilla link](https://bugzilla.mozilla.org/show_bug.cgi?id=1941460)

[Deploy Preview](https://deploy-preview-865--mozilla-perfcompare.netlify.app/)

### Description

This PR improves the search functionality by allowing users to do a full-text search by bug number, author, comments e.t.c

This is still a work in progress because the backend needs to be updated before this can be merged

Changes:
- removed logic that handles `author_contains` and replace with `search` 
- Check if a search is `hash` and then pass it to `?revision=`
- Update tests to handle new search update

Before

https://github.com/user-attachments/assets/091d3954-bbd7-4bc8-a2dd-363ddf85d29b

After 

Search by bug number, hash, author name, email and comment

https://github.com/user-attachments/assets/047e7be3-2f0a-4228-8112-c46384cdf6a6


src/resources/Strings.tsx
```diff
@@ -57,13 +57,15 @@ export const Strings = {
           },
           base: {
             selectLabel: 'Base',
-            inputPlaceholder: 'Search base by ID number or author email',
+            inputPlaceholder:
+              'Search base by ID number, author email or bug number',
```
Carla-Moz, line 61: And comment too, yes? 

`Search base by ID number, author email, bug number, or comment.`

Carla-Moz: Same for below for new

src/__tests__/Search/fetchRevisionsBySearch.test.tsx
```diff
@@ -50,11 +50,69 @@ describe('SearchView/fetchRevisionsByAuthor', () => {
...
+    );
+
+    await screen.findAllByText('Bug 123456 Fuzzy She turned me into a newt!');
+    expect(
+      screen.getAllByText("it's just a flesh wound")[0],
+    ).toBeInTheDocument();
+  });
+
+  it('should fetch revisions by phrase', async () => {
+    const { testData } = getTestData();
```
Carla-Moz, line 83: By `phrase` you mean `comments`, yes? If so, lets use the same language to stay consistent: comments

Carla-Moz (changes requested):
Awesome job! So happy you've gotten closer to the finish line. I've add some comments above.

Carla-Moz:
Also, searching by bug number works with `try` repo vs `mozilla central`

src/resources/Strings.tsx
Netacci: Alright will do

Netacci: > And author name and comment too, yes?
> 
> `Search base by ID number, author name or email, bug number, or comments.`

This text is too long for the input. Should I reduce the font size? or we could shorten the text

julienw: What's "ID number"? if it is the revision hash, I believe we should mention that.
Also we can remove "base" and "revision", I think this makes no sense in this context.
What about:
"Search by revision ID, author, bug number and comments"

Netacci: Updated now

Carla-Moz (approved):
Thanks for the update, @Netacci ! Congrats on completing this super helpful work.

julienw:
Hey folks,

This is working well and the code looks pretty good to me, but unfortunately I think the server-side code is still too slow to enable this by default.

We would need to put this work behind a flag so that we can land it and it doesn't bitrot.
The easiest could be to add a parameter (such as "useFulltextSearch") to the URL, like we have for `fakedata` (see `src/components/CompareResults/loader.ts`), that would trigger this code.
This means that we need to keep the old code too, because that would still be the default.

Netacci:
> Hey folks,
> 
> This is working well and the code looks pretty good to me, but unfortunately I think the server-side code is still too slow to enable this by default.
> 
> We would need to put this work behind a flag so that we can land it and it doesn't bitrot. The easiest could be to add a parameter (such as "useFulltextSearch") to the URL, like we have for `fakedata` (see `src/components/CompareResults/loader.ts`), that would trigger this code. This means that we need to keep the old code too, because that would still be the default.

Okay I understand. I wonder why it's still slow though since the ginindex is working.

julienw:
> Okay I understand. I wonder why it's still slow though since the ginindex is working.

I think the generated query is still pretty complex. And there are a ton of data.
Remember that I found out a similar SQL query that was faster but that we didn't manage to be able to generate it in the first patch. Maybe it would be easier now that we have the index?
Maybe we'll need to add even less data in the ginindex (slice the comments even more, ideally keeping just the first line or the first 50 bytes)?
Maybe it would be easier to add a separate endpoint so that we can have a more simple query?

Lots of unknowns!

Carla-Moz:
> Hey folks,
> 
> This is working well and the code looks pretty good to me, but unfortunately I think the server-side code is still too slow to enable this by default.
> 
> We would need to put this work behind a flag so that we can land it and it doesn't bitrot. The easiest could be to add a parameter (such as "useFulltextSearch") to the URL, like we have for `fakedata` (see `src/components/CompareResults/loader.ts`), that would trigger this code. This means that we need to keep the old code too, because that would still be the default.

@Netacci, can you please implement @julienw 's suggestion above and rebase? It would be great if this could be checked in. We can trigger this once we can improve the speed of the backend code.

Netacci:
Alright will do

Netacci:
> > Hey folks,
> > This is working well and the code looks pretty good to me, but unfortunately I think the server-side code is still too slow to enable this by default.
> > We would need to put this work behind a flag so that we can land it and it doesn't bitrot. The easiest could be to add a parameter (such as "useFulltextSearch") to the URL, like we have for `fakedata` (see `src/components/CompareResults/loader.ts`), that would trigger this code. This means that we need to keep the old code too, because that would still be the default.
> 
> @Netacci, can you please implement @julienw 's suggestion above and rebase? It would be great if this could be checked in. We can trigger this once we can improve the speed of the backend code.

This can be checked here https://deploy-preview-865--mozilla-perfcompare.netlify.app/?useFulltextSearch

Carla-Moz:
> Hey folks,
> 
> This is working well and the code looks pretty good to me, but unfortunately I think the server-side code is still too slow to enable this by default.
> 
> We would need to put this work behind a flag so that we can land it and it doesn't bitrot. The easiest could be to add a parameter (such as "useFulltextSearch") to the URL, like we have for `fakedata` (see `src/components/CompareResults/loader.ts`), that would trigger this code. This means that we need to keep the old code too, because that would still be the default.

@Neta, can you implement  @julien's suggestion so this can be rebased and landed? I'd prefer not to have this get too old with incoming patches.

PR #867, opened 2025-03-18
Fix: Graph spills over the container on first load

sumairq:
### Description:
This PR fixes the issue where the `CommonDistribution` graph was spilling over its container on the initial load. The issue was caused by using relative units for the container width. I updated it to an absolute width of `390px`, similar to how it's implemented for the `RunValues` graph. This ensures consistent rendering without overflow.  

Additionally, I refined the approach for maintaining the graph's aspect ratio. Instead of setting `maintainAspectRatio: false`, which caused the chart to expand vertically when more height was available, I explicitly defined `aspectRatio: 1.7` in the options. Since `maintainAspectRatio` defaults to `true`, this ensures proper scaling within the container while maintaining a consistent aspect ratio.

latest


Carla-Moz:
@sumairq Thanks, Sumair. It's a bit tedious, but a few more things before we land this. The font-size of the info-caption is still too small. Here's how to fix the styles when increasing the font-size to 16. 

Before:

fontsize

Go to :

1.https://github.com/mozilla/perfcompare/blob/79dfc92a5c253661a7fc3e54412db74a75831724/src/styles/SelectedRevsStyle.ts#L46
3. Reduce from `191px` to `140px`
4. https://github.com/mozilla/perfcompare/blob/79dfc92a5c253661a7fc3e54412db74a75831724/src/styles/SelectedRevsStyle.ts#L55
5. Add `fontSize: FontSizeRaw.Normal.fontSize` to `.info-caption`
6. Delete this line here: https://github.com/mozilla/perfcompare/blob/79dfc92a5c253661a7fc3e54412db74a75831724/src/styles/SelectedRevsStyle.ts#L83
7. Delete this line here: https://github.com/mozilla/perfcompare/blob/79dfc92a5c253661a7fc3e54412db74a75831724/src/styles/SelectedRevsStyle.ts#L94
8. Delete this line here: https://github.com/mozilla/perfcompare/blob/79dfc92a5c253661a7fc3e54412db74a75831724/src/styles/SelectList.ts#L10

After:
Screenshot 2025-03-19 at 16 49 14

src/components/CompareResults/CommonGraph.tsx
```diff
@@ -39,6 +39,7 @@ function CommonGraph({
       },
     },
     responsive: true,
+    maintainAspectRatio: false,
```
julienw, line 42: I think we need a max-height somewhere
see the behavior for the graphs in https://deploy-preview-867--mozilla-perfcompare.netlify.app/subtests-compare-results?baseRev=6237f52367ba70569588068ce9c00e2607bca802&baseRepo=mozilla-central&newRev=0bc4d1520de177e298ad4d051b99d0ac435e2b3c&newRepo=mozilla-central&framework=13&baseParentSignature=5273367&newParentSignature=5273367 for example

sumairq:
@julienw I’ve fixed the height issue of the graph by adjusting the `aspectRatio` option instead of disabling `maintainAspectRatio`.

Previously, setting `maintainAspectRatio: false` caused the chart to expand vertically whenever more height was available, leading to inconsistent scaling. A better approach was to explicitly define `aspectRatio: 1.7` in the options and keep `maintainAspectRatio` set to true (or simply remove it, since true is the default).
image

julienw (approved):
Thanks, this works for me. In my future patch I'll be able to reuse this (possibly a bit different) but this is already a good improvement.

PR #872, opened 2025-04-07
Bug 1949264 - Add git migration compatibility for ./mach try perf

Andrej1198:
As we are moving away from hg to git ./mach try perf we will no longer be able to get the try url and will be unable to generate perfcompare links. This patch along with the 2 others on treeherder and mozilla central add a hash to the commit message and allows us to get a perfcompare url that finds the commits via that hash This patch is still a work in progress as to properly test I will need the treeherder one to merge first to test properly as currently I am blocked with cross origin requests being not being allowed

### DIFF
--- a/src/components/CompareResults/loader.ts
+++ b/src/components/CompareResults/loader.ts
@@ -4,6 +4,7 @@ import {
   fetchCompareResults,
   fetchFakeCompareResults,
   memoizedFetchRevisionForRepository,
+  fetchRevisionFromHash,
 } from '../../logic/treeherder';
 import { Changeset, CompareResultsItem, Repository } from '../../types/state';
 import { FakeCommitHash, Framework } from '../../types/types';
@@ -181,10 +182,21 @@ export async function loader({ request }: { request: Request }) {
   }
 
   const baseRevFromUrl = url.searchParams.get('baseRev');
+  const baseHashFromUrl = url.searchParams.get('originalHash');
+  if (baseRevFromUrl == null && baseHashFromUrl != null) {
+    const promises = await fetchRevisionFromHash(baseHashFromUrl, 'try');
+    console.log(promises)
+  }
+
   const baseRepoFromUrl = url.searchParams.get('baseRepo') as
     | Repository['name']
     | null;
   const newRevsFromUrl = url.searchParams.getAll('newRev');
+  const newHashFromUrl = url.searchParams.get('newHash');
+  if (newRevsFromUrl == null && newHashFromUrl != null) {
+    const promises = await fetchRevisionFromHash(newHashFromUrl, 'try');
+    console.log(promises)
+  }
   const newReposFromUrl = url.searchParams.getAll(
     'newRepo',
   ) as Repository['name'][];
--- a/src/logic/treeherder.ts
+++ b/src/logic/treeherder.ts
@@ -1,7 +1,7 @@
 import moize from 'moize';
 
 import { JobInformation } from '../types/api';
-import { CompareResultsItem, Repository, Changeset } from '../types/state';
+import { CompareResultsItem, Repository, Changeset, CommitToHash} from '../types/state';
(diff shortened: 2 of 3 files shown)


Andrej1198:
No need to review, I added an 'on hold', I need  to properly test the patch and to do that I need to get the treeherder changes merged first and pass all linting checks

src/components/CompareResults/loader.ts
```diff
@@ -180,11 +181,23 @@ export async function loader({ request }: { request: Request }) {
     };
   }
 
-  const baseRevFromUrl = url.searchParams.get('baseRev');
+  let baseRevFromUrl = url.searchParams.get('baseRev');
+  let newRevsFromUrl = url.searchParams.getAll('newRev');
+  const baseHashFromUrl = url.searchParams.get('originalHash');
```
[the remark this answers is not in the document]
Andrej1198: I don't know why I did this, I'm re-writing to fix it here and in the treeherder patch

src/components/CompareResults/loader.ts
```diff
@@ -180,11 +181,28 @@ export async function loader({ request }: { request: Request }) {
     };
   }
 
-  const baseRevFromUrl = url.searchParams.get('baseRev');
+  let baseRevFromUrl = url.searchParams.get('baseRev');
+  let newRevsFromUrl = url.searchParams.getAll('newRev');
+  const baseHashFromUrl = url.searchParams.get('baseHash');
```
Carla-Moz, line 186: Could please add a brief comment on why these changes are necessary?

Carla-Moz, line 186: Where are `baseHashfromUrl` and `newRevsFromUrl` coming from? Not from the inputs in `Compare with Base`, so I'm confused here. I feel these values will always be null and never reach the if condition below to `fetchRevisionFromHash`.

Carla-Moz (changes requested):
Hi Andrej, thanks for this PR. I have some questions below to help me better review this patch.

src/components/CompareResults/loader.ts
Carla-Moz: Resolved through zoom meeting.

Andrej1198:
Hey @Carla-Moz I added some comments in my code and tests to better explain things
I also added a better commit message to properly explain context, and why I am doing this
Finally I also reverted a change we spoke about and re-defined `baseRevFromUrl` and `newRevsFromUrl` back to constants
Let me know if you have any issues with my work :)

src/components/CompareResults/loader.ts
```diff
@@ -180,22 +180,32 @@ export async function loader({ request }: { request: Request }) {
...
   const newHashFromUrl = url.searchParams.get('newHash');
+
+  // This loader is the only one used by ./mach try perf, and due to
+  // recent changes mach try perf can't get instant push links to try
+  // when we push, we must attach a hash to the commit message and
+  // search commits for that hash, and return the commit associated
+  // with that hash and update the baseRev and newRev
+  let baseRevsFromHash = baseRevFromUrl;
+  let newRevsFromHash = newRevsFromUrl;
+  let usingRevisionsOrHash = true;
```
Carla-Moz, line 195: Nit: Wondering if the naming here could be clearer. Something like  `notUsingHashToFindRevisions` or something of the sort.

Carla-Moz (approved):
@Andrej1198 Thanks for the updates. I just have one nit but LGTM!

src/components/CompareResults/loader.ts
Andrej1198: I've change to use:  `notUsingHashToFindRevisions`

julienw:
I'm coming a bit late: have you considered using a different endpoint? That would make it possible to use a different loader instead of overloading the existing one (while still using the exacty same react components to render the results).

But without trying it I don't know if this is more or less complex.

PR #877, opened 2025-04-24
PCF-523 Results table: add nowrap to tags list

esanuandra:
[Jira ticket](https://mozilla-hub.atlassian.net/browse/PCF-523)

[Deploy test link](https://deploy-preview-877--mozilla-perfcompare.netlify.app/compare-results?baseRev=232849eaaf94591746a518bcfbb1973d544022e0&baseRepo=mozilla-central&newRev=8cfe77187cb705982852f38bb1b918a2f916aa28&newRepo=mozilla-central&framework=1) there are some examples down on the page with longer test names. I grouped the revision with the copy revision button so that the button does not go alone on the next row.

Screenshot 2025-05-22 at 17 56 08

### DIFF
--- a/src/components/CompareResults/LinkToRevision.tsx
+++ b/src/components/CompareResults/LinkToRevision.tsx
@@ -14,7 +14,7 @@ const styles = {
     fontSize: '16px',
     lineHeight: '1.5',
   }),
-  align: style({
+  nowrap: style({
     display: 'flex',
     flexWrap: 'nowrap',
   }),
@@ -29,7 +29,7 @@ export default function LinkToRevision(props: LinkToRevisionProps) {
   const { result } = props;
   const shortHash = truncateHash(result.new_rev);
   return (
-    <div className={styles.align}>
+    <div className={styles.nowrap}>
       <Link
         href={getTreeherderURL(result.new_rev, result.new_repository_name)}
         target='_blank'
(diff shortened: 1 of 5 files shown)


Carla-Moz (approved):
Thanks, Andra. LGTM ! I know @julienw mentioned adding it to the items themselves (correct if I'm wrong) but this solution works fine too, imho.

julienw:
I was looking at a mobile resolution using the responsive mode. But I tried my approach and it gives the same result. Also mobile has so much more issues. So I'm fine with the solution.

One idea though would be to use this instead:
```
display: flex;
flex-wrap: wrap;
```
Then the tags themselves could flow on several lines.

This could go with this on the same element:
```
margin: 0 4px;
gap: 4px;
```
while removing the `margin: 4px` we have on the indivdual tags themselves.

And tada:
image

src/components/CompareResults/LinkToRevision.tsx
```diff
@@ -14,6 +14,10 @@ const styles = {
     fontSize: '16px',
     lineHeight: '1.5',
   }),
+  align: style({
```
julienw, line 17: (optional nit) Maybe call that "nowrap" for a more self-describing name :)

```suggestion
  nowrap: style({
```

julienw (approved):
looks good to me, thanks

PR #882, opened 2025-04-29
Change how the filters are implemented so that the behavior is a bit better

julienw:
Previously the internal state was keeping the _unchecked_ checkboxes, now it keeps the _checked_ checkboxes. This makes the code easier to follow.
    
The user-visible change is that unknown values are not filtered out as expected when the user selects specific filters. They are still visible when all filters are checked. This was the motivation to write this patch.

You can test with these links:
[main branch](https://main--mozilla-perfcompare.netlify.app/compare-results?baseRev=ba56f7ee55792aaf974891449b64ace32bc9e1b8&baseRepo=mozilla-central&newRev=30d833468eba8358571c55ec3e7e8991f4952fb9&newRepo=mozilla-central&framework=2&search=fetch_content) / [deploy preview](https://deploy-preview-882--mozilla-perfcompare.netlify.app/compare-results?baseRev=ba56f7ee55792aaf974891449b64ace32bc9e1b8&baseRepo=mozilla-central&newRev=30d833468eba8358571c55ec3e7e8991f4952fb9&newRepo=mozilla-central&framework=2&search=fetch_content)

This is using the build_metrics framework, and searching for `fetch_content` which has some of these "unknown" platforms. For example `updatebot` near the top. 

Now you can try selecting other platforms from the menu. It is especially visible when using one of the `Select only XXX` options. When using one of these options the user expects to see _only_ these, but in production the user sees also the unknown platforms. Now with this patch they disappear as I'd expect as a user.

Tell me what you think!

### DIFF
--- a/src/components/CompareResults/TableHeader.tsx
+++ b/src/components/CompareResults/TableHeader.tsx
@@ -72,7 +72,7 @@ type FilterableColumnHeaderProps = {
 
   /* Properties for filtering */
   possibleValues: FilterableColumn['possibleValues'];
-  uncheckedValues?: Set<string>;
+  checkedValues?: Set<string>;
   onToggleFilter: (checkedValues: Set<string>) => unknown;
   onClearFilter: () => unknown;
   tooltip?: string;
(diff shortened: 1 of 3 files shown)


beatrice-acasandrei (approved):

src/components/CompareResults/TableHeader.tsx
```diff
@@ -82,35 +82,33 @@ function FilterableColumnHeader({
...
 
   const onClickOnlyFilter = (valueKey: string) => {
-    const newUncheckedValues = new Set(
-      possibleValues
-        .filter(({ key }) => key !== valueKey)
-        .map(({ key }) => key),
-    );
-    onToggleFilter(newUncheckedValues);
+    const newCheckedValues = new Set([valueKey]);
+    onToggleFilter(newCheckedValues);
```
Carla-Moz, line 107: Thanks for this update. It makes a lot more sense reading this vs before.

Carla-Moz:
Ah okay, so your patch makes sure the filters _do not_ include the `unknown`, for example, re platform selection. Makes sense. I've tested your deploy preview and so far, it works as expected. And it works when other filters are selected too like Confidence.

julienw:
> Ah okay, so your patch makes sure the filters _do not_ include the `unknown`, for example, re platform selection. Makes sense. I've tested your deploy preview and so far, it works as expected. And it works when other filters are selected too like Confidence.

Yeah that's it! If all checkboxes are selected, we see them, but as soon as one is unchecked (that is the user is interested only in some of the items), we don't show the unknowns anymore.

Maybe in the future we'll want to generate the filters from the actual values, but I also like the current situation where only the important values are shown. But we might want to hide the filters that are not present at all (eg ios is rarely present). That's work for another day!

PR #885, opened 2025-05-14
Update eslint to latest version, and migrate its config to flat config

julienw:
This PR updates eslint to its latest version. Since v9 the flat config format is mandatory. It is a bit cumbersome but I managed to convert it, and I think it's also clearer now.

Here is the migration guide for more information: https://eslint.org/docs/latest/use/configure/migration-guide

This also enabled the cache, which should make eslint invocations a lot faster after the first run.


package.json
```diff
@@ -23,14 +23,11 @@
     "@reduxjs/toolkit": "^1.8.3",
     "@types/react": "^18.0.15",
     "@types/react-dom": "^18.0.6",
-    "@typescript-eslint/eslint-plugin": "^8.18.0",
-    "@typescript-eslint/parser": "^8.18.0",
```
julienw, line 27: They're not used anymore, now we need to import `typescript-eslint` directly.

package.json
```diff
@@ -23,14 +23,11 @@
...
     "assert": "^2.1.0",
     "buffer": "^6.0.3",
     "chart.js": "^4.3.0",
     "crypto-browserify": "^3.12.0",
     "dayjs": "^1.11.7",
-    "eslint-plugin-jest-formatting": "^3.1.0",
```
julienw, line 33: it's been folded in `eslint-plugin-jest` in recent versions.

Carla-Moz (approved):
Thanks for taking on this cumbersome work, @julienw ! Deploy and local work fine for me.

julienw:
Thank you for the review!

PR #886, opened 2025-05-14
Update build-related packages

julienw:
This one is an easy one: it updates a bunch of packages, but otherwise everything should work like before.

It would be good to double check the deploy preview as well as the local development environment. On my side it seems to work well.


webpack/webpack.common.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
       assert: require.resolve('assert'),
       buffer: require.resolve('buffer'),
       crypto: require.resolve('crypto-browserify'),
+      process: require.resolve('process'),
```
julienw, line 16: webpack can find it automaticallly because it has the same name, but I find it better to mention it explicitely, so that we know more easily that it's actually being used.

package.json
```diff
@@ -46,11 +46,9 @@
     "react-router-dom": "^6.26.2",
     "react-virtuoso": "^4.9.0",
     "redux": "^4.2.0",
-    "run": "^1.4.0",
     "stream-browserify": "^3.0.0",
     "taskcluster-client-web": "^60.4.0",
     "taskcluster-lib-scopes": "^11.0.0",
-    "ts-node": "^10.9.1",
```
julienw, line 53: same for that

webpack/webpack.development.js
```diff
@@ -4,6 +4,5 @@ module.exports = {
   devServer: {
     port: 3000,
     historyApiFallback: true,
-    magicHtml: true,
```
julienw, line 7: this has been removed from webpack-dev-server... but this wasn't useful anyway.

Carla-Moz (approved):
LGTM, checked the preview and my local setup.

PR #887, opened 2025-05-16
Update jest, express, and other develoment-related packages

julienw:
Continuing my quest :-)

It could be easier to look at it commit by commit, but if you want to look at it all at once this could work too, as I added some comments for the most surprising things.

Tell me what you think!

### DIFF
--- a/src/__tests__/utils/test-utils.tsx
+++ b/src/__tests__/utils/test-utils.tsx
@@ -15,11 +15,9 @@ import { store } from './setupTests';
 import SnackbarCloseButton from '../../components/Shared/SnackbarCloseButton';
 import getProtocolTheme from '../../theme/protocolTheme';
 
-type ChildrenProps = { children: React.ReactElement };
-
 type ThemeConfig = Partial<Theme> | null;
 export function render(ui: React.ReactElement, themeConfig?: ThemeConfig) {
-  function Wrapper({ children }: ChildrenProps) {
+  function Wrapper({ children }: { children: React.ReactNode }) {
     const { protocolTheme } = getProtocolTheme('light');
     const theme = themeConfig ? createTheme(themeConfig) : protocolTheme;
     return (


package.json
```diff
@@ -121,18 +122,18 @@
...
-    "jest-axe": "^6.0.0",
-    "jest-environment-jsdom": "^30.0.0-alpha.3",
-    "jest-resolve": "^28.1.3",
-    "jest-watch-typeahead": "^2.0.0",
-    "npm-run-all2": "^6.0.6",
+    "jest": "^29.7.0",
+    "jest-axe": "^10.0.0",
+    "jest-environment-jsdom": "^30.0.0-alpha.7",
+    "jest-resolve": "^29.7.0",
+    "jest-watch-typeahead": "^2.2.2",
```
julienw, line 129: I personally don't use this extension to jest, but I still kept it to be sure...

server.js
```diff
@@ -6,12 +6,15 @@ var app = express();
...
 });
 
 app.set('port', process.env.PORT || 8080);
 
-var server = app.listen(app.get('port'), function () {
-  console.log('listening on port ', server.address().port);
+var server = app.listen(app.get('port'), function (error) {
+  if (error) {
+    throw error;
+  }
```
julienw, line 18: The new version of express now reports the errors to the callback, previously it was throwing directly.

server.js
```diff
@@ -6,12 +6,15 @@ var app = express();
 
 const rootDir = path.join(__dirname, 'build');
 app.use(express.static(rootDir));
-app.get('/*', (req, res) => {
+app.get('/*splat', (req, res) => {
```
julienw, line 9: The new version of express requests (no pun intended) this.

See the full migration guide at https://expressjs.com/en/guide/migrating-5.html

src/__tests__/App.test.tsx
```diff
@@ -54,7 +54,7 @@ describe('App', () => {
 
   test('Should switch between dark mode and light mode on toggle', async () => {
     // set delay to null to prevent test time-out due to useFakeTimers
-    const user = userEvent.setup({ delay: null });
+    const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime });
```
julienw, line 57: `advanceTimers` is a newish option, that integrates better with fake timers. This makes it possible to remove `delay: null`, and therefore to be closer to how a user actually interacts with the app.

see https://testing-library.com/docs/user-event/options#advancetimers for more info.

src/__tests__/CompareResults/__snapshots__/ResultsTable.test.tsx.snap
```diff
@@ -2431,7 +2431,7 @@ exports[`Results Table can sort the table and persist the sort parameters to the
 exports[`Results Table can sort the table and persist the sort parameters to the URL 2`] = `
 <button
   aria-label="Delta (Currently sorted by this column. Click to change)"
-  class="MuiButtonBase-root MuiButton-root MuiButton-contained MuiButton-containedTableHeaderButton MuiButton-sizeSmall MuiButton-containedSizeSmall MuiButton-disableElevation MuiButton-root MuiButton-contained MuiButton-containedTableHeaderButton MuiButton-sizeSmall MuiButton-containedSizeSmall MuiButton-disableElevation css-cdzw26-MuiButtonBase-root-MuiButton-root"
+  class="MuiButtonBase-root MuiButton-root MuiButton-contained MuiButton-containedTableHeaderButton MuiButton-sizeSmall MuiButton-containedSizeSmall MuiButton-disableElevation Mui-focusVisible MuiButton-root MuiButton-contained MuiButton-containedTableHeaderButton MuiButton-sizeSmall MuiButton-containedSizeSmall MuiButton-disableElevation css-cdzw26-MuiButtonBase-root-MuiButton-root"
```
julienw, line 2434: It's not super clear where the changes to the snapshot come from, but I believe it is related to the focus handling. It's possible a newer jsdom changes this, or that `await user.click()` and friends wait a little more than before. As a result the snapshot changes show focus-related changes (like the ripple effect).

src/__tests__/Search/CompareOverTime.test.tsx
```diff
@@ -155,19 +155,19 @@ describe('Compare Over Time', () => {
       name: 'mozilla-central',
     });
     await user.click(mozRepoItem);
-    expect(mozRepoItem).toBeInTheDocument();
+    expect(baseDropdown).toHaveTextContent('mozilla-central');
```
julienw, line 158: Another case where I believe `await user.click()` resolves later than before.

Previously the option we just clicked was still present... but I believe this was a mistake! as the dropdown disappears after a click. I replaced by checking whether the button now reads the new selected option.

(same below)

src/__tests__/utils/test-utils.tsx
```diff
@@ -15,11 +15,9 @@ import { store } from './setupTests';
 import SnackbarCloseButton from '../../components/Shared/SnackbarCloseButton';
 import getProtocolTheme from '../../theme/protocolTheme';
 
-type ChildrenProps = { children: React.ReactElement };
-
 type ThemeConfig = Partial<Theme> | null;
 export function render(ui: React.ReactElement, themeConfig?: ThemeConfig) {
-  function Wrapper({ children }: ChildrenProps) {
+  function Wrapper({ children }: { children: React.ReactNode }) {
```
julienw, line 20: I think the previous typing was incorrect, but now it's flagged.

Carla-Moz:
Will look at this soon!

beatrice-acasandrei (approved):
Looks good, thank you!

julienw:
Thanks beatrice!

PR #888, opened 2025-05-16
Update material UI

julienw:
I followed the migration guides in https://mui.com/material-ui/migration/upgrade-to-v6/ and https://mui.com/material-ui/migration/upgrade-to-v7/. Fortunately they provided a bunch of codemods which helped a lot.

You can follow commit by commit, here are the changes:
1. Remove the dependency to `@emotion/react` because we're not using it directly. It's still present by transitivity in the end, so not a big deal. Material UI wants us to depend directly on `@emotion/styled`.
2. Update the packages themselves.
3. Update to the new Grid component -- it's fairly similar but has a bunch of property changes. I could have kept using the old Grid component, but decided to move forward and convert it. I find the new API much more consistent and straightforward to understand. See https://mui.com/material-ui/migration/upgrade-to-grid-v2/. I used a codemod for that, but it would be good to double check there's no issue with it in the end. I couldn't find any myself.
4. Changes in how the dark mode is implemented. My understanding is that with these changes, a theme change won't rerender as much as before. See https://mui.com/material-ui/migration/upgrade-to-v6/#color-mode-theme-utility. There are deeper changes to make but didn't want to make them here, related to the theme. See https://mui.com/material-ui/migration/upgrade-to-v7/#theme-behavior-changes for the other things we could do.
5. System properties deprecated in favor of `sx`. See https://mui.com/material-ui/migration/migrating-from-deprecated-apis/#system-props
6. `slotProps` being used more, see all of https://mui.com/material-ui/migration/migrating-from-deprecated-apis/. I did that one manually, I didn't see there was a general codemod at the start of this file 😅. Typescript helped a lot though.
7. Simple type change, coming from https://mui.com/material-ui/migration/upgrade-to-v7/#package-layout-updated
8. Finally test changes. The biggest change is due to the fact that `Select` now renders a `combobox` role instead of a `button`, with a different accessible name, so I had to update a bunch of tests here and there. The snapshot changes are in the next commit.
9. I split out the snapshot changes in this extra commit.


src/components/CompareResults/RevisionSelect.tsx
```diff
@@ -41,6 +41,7 @@ function RevisionSelect() {
         defaultValue=''
         displayEmpty
         size='small'
+        inputProps={{ 'aria-label': 'Filter by revision' }}
```
julienw, line 44: Without this new prop, the select didn't have any label anymore, and we couldn't target it in tests. This is also better for accessibility.

src/__tests__/CompareResults/RevisionSelect.test.tsx
```diff
@@ -27,8 +27,8 @@ describe('Revision select', () => {
     expect(await screen.findByTestId('revision-select')).toBeInTheDocument();
     expect(document.body).toMatchSnapshot();
 
-    const selectButton = await screen.findByRole('button', {
-      name: 'All revisions',
+    const selectButton = await screen.findByRole('combobox', {
+      name: 'Filter by revision',
```
julienw, line 31: as you see, now the accessible name is more like a label, but before it was its value (I think because we didn't add an `aria-label`)

beatrice-acasandrei (approved):
I tested the PR locally, couldn't notice any issue. Thanks!

julienw:
Thanks for the review!

PR #889, opened 2025-05-20
In the tests for SubtestsRevisionRow, wait until the page is rendered before asserting

julienw:
This also fixes a warning about a missing `act`.

### DIFF
--- a/src/__tests__/CompareResults/SubtestsRevisionRow.test.tsx
+++ b/src/__tests__/CompareResults/SubtestsRevisionRow.test.tsx
@@ -83,6 +83,9 @@ describe('SubtestsRevisionRow Component', () => {
       />,
     );
 
+    // wait until all elements render correctly
+    await screen.findByText('browser.html');
+
     expect(document.body).toMatchSnapshot();
     expect(await screen.findByText(/firefox/i)).toBeInTheDocument();
     expect(await screen.findByText(/chrome/i)).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


src/__tests__/CompareResults/SubtestsRevisionRow.test.tsx
```diff
@@ -83,6 +83,9 @@ describe('SubtestsRevisionRow Component', () => {
       />,
     );
 
+    // wait until all elements render correctly
+    await screen.findByText('browser.html');
+
     expect(document.body).toMatchSnapshot();
```
julienw, line 89: See the changes below, you'll see that the snapshots were empty before, because the component wasn't rendered.

src/__tests__/CompareResults/SubtestsRevisionRow.test.tsx
```diff
@@ -98,6 +101,9 @@ describe('SubtestsRevisionRow Component', () => {
       />,
     );
 
+    // wait until all elements render correctly
+    await screen.findByText('dhtml.html');
```
julienw, line 105: It is even more important in this test.

Indeed in the previous test, we were using `await` until the text for either `firefox` or `chrome` was showing up. So at the end of the test, the component was rendered.

But in _this_ test, we're using `queryByText` which doesn't wait. So:
1. Indeed the page wasn't rendered at all, so of course the text wasn't there,
2. but also the page was still being rendered _after_ the test. This is what triggered the warning about `act`: something was rendered when the test wasn't waiting anything.

Netacci: Okay this makes sense

Netacci (approved):

PR #891, opened 2025-06-04
Increase container maxWidth to 1300px to match main results table

sumairq:
**Related PR https://github.com/mozilla/perfcompare/pull/849**

## Description
This PR updates the maxWidth property in `styles.container` from `"xl" / (1200px)` to a fixed value of `1300px` to match the layout used in the main results table. The previous implementation used a hardcoded maxWidth of `1300px` directly on the Container component, but this was lost when we switched to using the `styles.container` class with the `maxWidth: "xl"` setting.


julienw (approved):
Thank you!

PR #894, opened 2025-06-06
Update chart.js, react, and redux

julienw:
This should have been a straightforward update, but it wasn't as easy after all.

You can look at the commits one by one and it shouldn't that bad. :-)


src/components/Accessibility/SkipLink.tsx
```diff
@@ -18,7 +20,6 @@ const SkipLink: React.FC<SkipLinkProperties> = (props) => {
 
   return React.cloneElement(props.children, {
     onClick,
-    ref: props.containerRef,
```
julienw, line 21: I think this was a mistake...

src/components/Accessibility/SkipLink.tsx
```diff
@@ -1,8 +1,10 @@
 import React from 'react';
 
 interface SkipLinkProperties {
-  containerRef: React.RefObject<HTMLElement>;
-  children: React.ReactElement;
+  containerRef: React.RefObject<HTMLElement | null>;
+  children: React.ReactElement<{
+    onClick: (e: React.SyntheticEvent) => unknown;
```
julienw, line 6: Fixes an issue with the call to cloneElement below. Basically we need to pass the expected props of the element;

src/__tests__/Search/SearchContainer.test.tsx
```diff
@@ -21,7 +21,7 @@ function setupTestData() {
 
 function renderComponent() {
   setupTestData();
-  const ref: React.RefObject<HTMLInputElement> = React.createRef();
+  const ref: React.RefObject<HTMLInputElement | null> = React.createRef();
```
julienw, line 24: It looks like all refs need to have `null` as an explicit option now.

webpack/webpack.common.js
```diff
@@ -54,9 +54,10 @@ module.exports = {
   plugins: [
     new webpack.ProvidePlugin({
       Buffer: ['buffer', 'Buffer'],
-    }),
-    new webpack.ProvidePlugin({
-      process: 'process/browser',
+
+      // We need to use require.resolve, otherwise we get this issue with
+      // react-redux: https://github.com/reduxjs/redux-toolkit/issues/4990
+      process: require.resolve('process/browser'),
```
julienw, line 60: I got a build error without that..

I think we need `process` initially because of taskcluster-client-web otherwise.

julienw:
multumesc!

PR #896, opened 2025-06-12
Update the react router package

julienw:
This should not change anything.

You can follow commit by commit:
1.    Update the react-router-dom package and replace it with the react-router package directly
The new version of react-router merged everything into one package.

2.    Update import directives    
All packages were merged to react-router, so we need to update all
imports to import from there.
Only ReactProvider is special as they have 2 different versions,
depending whether we're in a DOM environment or in a different
environment.

3.    Update tsconfig moduleResolution so that it picks up the right imports
The previous configuration "node" is quite outdated. This commit
replaces it with "bundler" which seems appropriate in our case.

4.    Add the hydrateFallback information to avoid a warning
Currently I added it with an empty component, so that the behavior isn't
changed. In the future we could decide to replace this with a spinner if
needed. (But I don't like spinners much...)

5.    Update snapshots

6.    Change how the types of the loader function is specified with useLoaderData
The old way was triggering an issue with the rule @typescript-eslint/no-unnecessary-type-assertion.
See https://github.com/typescript-eslint/typescript-eslint/issues/11298

I did test the app a lot locally, but please test as well as this might impact various cases.


src/components/CompareResults/ResultsTable.tsx
```diff
@@ -228,7 +228,7 @@ export default function ResultsTable() {
           {(resolvedResults) => (
             <TableContent
               columnsConfiguration={columnsConfiguration}
-              results={resolvedResults as CompareResultsItem[][]}
```
julienw, line 231: `resolvedResults` seems to be properly typed automatically now

PR #897, opened 2025-06-12
Update fetch-mock-jest to @fetch-mock/jest

julienw:
Our version of fetch-mock-jest was very outdated, and was triggering some security advisories.
Now they changed the package names and changed some syntaxes.
It's not a very big deal though, mostly tedious changes here and there.

The biggest syntax change is that we now have 2 different objects: `fetch` is the mock handler while `fetchMock` is the mock configurator. In `expect` we can pass one or the other, both work. Previously both responsibilities were in the same object `fetch`.

Finally the last commit contains a change that _look_ unrelated, but this was producing an `act` warning after the update. I'm not sure why this was appearing now, because I'm pretty sure that the issue was preexisting. Fortunately the new library makes it easier to wait for the fetch to finish (because here we don't have any visible update that we could use).

Tell me how you like it :-)


src/__tests__/TaskclusterAuth/TaskclusterCallback.test.tsx
```diff
@@ -63,28 +61,34 @@ describe('Taskcluster Callback', () => {
...
+    const tokenCall = fetchMock.callHistory.lastCall(
+      'https://firefox-ci-tc.services.mozilla.com/login/oauth/token',
+    );
+    const body = tokenCall?.options.body;
+    if (!(body instanceof URLSearchParams)) {
+      throw new Error(
+        'Unexpected body passed to the endpoint /login/oauth/token',
+      );
+    }
+    const requestBody = Object.fromEntries(body);
```
julienw, line 83: FYI this part is the less "automatic" part, where I had to change more code than needed. We don't have access to the direct calls now, but we can queries calls with the callHistory API https://www.wheresrhys.co.uk/fetch-mock/docs/API/CallHistory

PR #899, opened 2025-06-12
Move some colors to protocolTheme and use a Stack in the expanded RevisionRow

julienw:
This moves the colors that change in light/dark mode in the revision rows to protocolTheme, and simplifies these components in the process.

In RevisionRowExpandable this also uses Stack for the layout now. Doing this, the spacing around the title of the row (the platform) changed, but looking at our figma design, it looks like the new spacing is more inline with the initial plan.

The motivation is that I'm working on these components for another patch and did this while working on that, but it's fairly orthogonal to the other work.

It's probably easier to look at this PR commit by commit, and ignoring whitespaces.

Also you'll notice that I did some more work to remove typestyle completely in RevisionRowExpandable but I didn't do it for RevisionRow.

Before:
image

After:
image

Deploy preview:
https://deploy-preview-899--mozilla-perfcompare.netlify.app/subtests-compare-over-time-results?baseRepo=mozilla-central&newRev=5223faf9adce23b50640f7aca9baf395526094b9&newRepo=mozilla-central&framework=12&selectedTimeRange=86400&baseParentSignature=4769453&newParentSignature=4769453


src/theme/protocolTheme.ts
```diff
@@ -35,6 +35,12 @@ const lightMode = {
     improvement: Colors.Background500,
     regression: Colors.Background400,
   },
+  revisionRow: {
+    background: Colors.Background200,
+  },
+  expandedRow: {
```
julienw, line 41: `expandedRow` will be used in the next commit.

beatrice-acasandrei (approved):

PR #900, opened 2025-06-13
Update Jest to v30

julienw:
I also removed jest-watch-typeahead that's not working with jest 30 at the moment. Tell me if you were using it as I've seen there's a PR to make it work with jest 30 so we could wait for it before upgrading.

Otherwise there's a small snapshot change coming from, I believe, a change in jsdom. Note that jsdom is updated to v26 (it was v22 before), so we might have more recent APIs and behavior implemented.


julienw:
Yay the automatic review works :-)

PR #908, opened 2025-06-26
Do not do a request to taskcluster if there's no job to retrigger

julienw:
While I was testing #907 I realized that we're doing a request to treeherder even when we don't have a job id to retrigger (this is when we don't have a run for the selected revision). That request was returning a 500 error. When testing locally, webpack was showing a full page error (fortunately that wasn't happening in production).

Ideally we should also be able to start a run by copying the other run's configuration (PCF-71), but that's some more work.
Also ideally the select input in the modal should be disabled with some information given to the user, but similarly this was more work than I was willing to do right now.

So the goal of this PR is simply to avoid sending requests when we don't have a job to retrigger, because this request was incorrect (had "undefined" as the job id in its URL), and was returning an error 500.

### DIFF
--- a/src/components/CompareResults/Retrigger/RetriggerButton.tsx
+++ b/src/components/CompareResults/Retrigger/RetriggerButton.tsx
@@ -128,21 +128,27 @@ export function RetriggerButton({ result, variant }: RetriggerButtonProps) {
   }) => {
     setStatus('pending');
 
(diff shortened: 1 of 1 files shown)


beatrice-acasandrei (approved):

julienw:
> @julienw Should I file tickets for the follow-up improvements that you mentioned in the description?

yes why not !

PR #924, opened 2025-08-13
#1931317 Show N/A string instead of 0 deviation

misspran:
I'm a bit unsure about understanding the business logic correctly on the bug. Or what we would want as the string response for the standard deviation but I figure I'll just do the PR and get eyes on it quickly. 

Screenshot 2025-08-13 at 2 07 35 PM

[Deploy preview](https://deploy-preview-924--mozilla-perfcompare.netlify.app/compare-results?baseRev=7d880ff29269122bc521b3f69a643a3b7158a8eb&baseRepo=mozilla-central&newRev=31224efda311facc4ea54446610fd7c52d5d22a1&newRepo=try&framework=13&search=wikipedia)

### DIFF
--- a/src/components/CompareResults/RunValues.tsx
+++ b/src/components/CompareResults/RunValues.tsx
@@ -88,7 +88,7 @@ function RunValues(props: RunValuesProps) {
       <div className={styles.deviation}>
         {values.length > 1
           ? `${stddev} ${unit} = ${stddevPercent}% standard deviation`
-          : `${stddev} ${unit} = N/A standard deviation`}
+          : 'N/A standard deviation'}
       </div>
     </>
   );
--- a/src/components/CompareResults/RunValues.tsx
+++ b/src/components/CompareResults/RunValues.tsx
@@ -37,7 +37,6 @@ function RunValues(props: RunValuesProps) {
   const displayedValues = 100;
   const firstValues = values.slice(0, displayedValues);
   const lastValues = values.slice(displayedValues);
-
   const [expanded, setExpanded] = useState(false);
   const toggleIsExpanded = () => {
     setExpanded(!expanded);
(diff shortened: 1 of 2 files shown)


src/components/CompareResults/RunValues.tsx
```diff
@@ -87,7 +86,9 @@ function RunValues(props: RunValuesProps) {
         {measurementUnit}
       </div>
       <div className={styles.deviation}>
-        {stddev} {unit} = {stddevPercent}% standard deviation
+        {values.length > 1
+          ? `${stddev} ${unit} = ${stddevPercent}% standard deviation`
+          : `${stddev} ${unit} = N/A standard deviation`}
```
canova, line 91: Currently we still print `${stddev} ${unit}` but I think we should remove it completely and just keep `N/A standard deviation`. Because `stddev` is the standard deviation value, and seeing it as `0 MS = N/A` might still confuse some people. Zero here means that we failed to calculate it.

canova (commented):
Thanks for the changes and for the screenshot! For making the reviews easier we usually put a deploy preview link so people can also load the link and see directly. You can see the deploy preview base url in the netlify comment. For this PR it is: https://deploy-preview-924--mozilla-perfcompare.netlify.app/ (only the part that changes is the PR number, so you can also type manually)

You can change the link Greg gave in the bug to point to this deploy preview like: https://deploy-preview-924--mozilla-perfcompare.netlify.app/compare-results?baseRev=7d880ff29269122bc521b3f69a643a3b7158a8eb&baseRepo=mozilla-central&newRev=31224efda311facc4ea54446610fd7c52d5d22a1&newRepo=try&framework=13&search=wikipedia

(I added this link to your initial comment now so people can see it)


The changes look good! I have one comment about the way we still print `stddev` below.

src/components/CompareResults/RunValues.tsx
beatrice-acasandrei: @gmierz This looks good to me, after the PR is rebased in my opinion we could merge it. Do you have other feedback?

gmierz: This looks great to me, thanks! :)

beatrice-acasandrei (approved):

PR #935, opened 2025-09-23
Bug 1915775 - Change spacing between "PerfCompare" header and "Home" back button

log101:
Changed styling of "PerfCompare" header and added a new style for the "Home" button to match Figma design. Now the spacing between the home button and header is 42px and the spacing between the home button and the comparison card is 26px as in the Figma design.


src/components/Shared/LinkToHome.tsx
```diff
@@ -1,19 +1,31 @@
...
-        direction='row'
-        sx={{
-          alignItems: 'center',
-        }}
-      >
-        <ChevronLeftIcon fontSize='small' />
-        <p>Home</p>
-      </Stack>
-    </Link>
+    <Box className={`${styles.box}`}>
```
kala-moz, line 17: This can be `className={styles.box}` No need for the template literals here since we don't have a string name for the className

kala-moz (changes requested):
Thanks for fixing this issue. I've requested a small change. After you've rebased, we can check this in.

kala-moz (approved):
Thanks!

PR #1024, opened 2026-04-06
Bug-1986319: Allow auto selection of full hash matches

moijes12:
Updated `SearchInputAndResults` to:
* Track input value as state
* Check if the input is a hash and if it matches an existing hash
* Auto select the hash if it matches the full hash

@kala-moz @beatrice-acasandrei 

Fixes [Bug-1986319](https://bugzilla.mozilla.org/show_bug.cgi?id=1986319)

### DIFF
--- a/src/components/Search/SearchInputAndResults.tsx
+++ b/src/components/Search/SearchInputAndResults.tsx
@@ -34,6 +34,8 @@ export default function SearchInputAndResults({
   );
   const [searchError, setSearchError] = useState(null as null | string);
 
+  const [inputValue, setInputValue] = useState('');
+
   // The last used searchTerm is kept in a ref, so that it's possible to use it
   // in an effect when the repository prop changes. It's not stored in a state
   // because it's not used for rendering DOM.
@@ -66,8 +68,15 @@ export default function SearchInputAndResults({
     }
   }, []);
 
+  // Helper to check if the search term matches a full 40 character revision hash
+  const isFullRevisionHash = (searchTerm: string) =>
+    /^[a-f0-9]{40}$/i.test(searchTerm.trim());
+
   const searchRecentRevisions = useCallback(
-    async (searchTerm: string) => {
+    async (
+      searchTerm: string,
+      { autoSelect }: { autoSelect: boolean } = { autoSelect: false },
+    ) => {
       // If the URL has a parameter useFulltextSearch, use it to determine the search type.
       const urlParams = new URLSearchParams(window.location.search);
       const useFulltextSearch = urlParams.has('useFulltextSearch');
@@ -116,6 +125,29 @@ export default function SearchInputAndResults({
 
       try {
         const results = await fetchRecentRevisions(searchParameters);
+
+        if (autoSelect && isFullRevisionHash(searchTerm)) {
+          const completeHashMatch = results.find(
+            (rev) => rev.revision.toLowerCase() === searchTerm.toLowerCase(),
+          );
+          if (completeHashMatch) {
+            if (searchType === 'new') {
+              const alreadySelected = displayedRevisions.some(
+                (rev) => rev.id === completeHashMatch.id,
+              );
+              if (alreadySelected) {
+                setDisplayDropdown(false);
+                setInputValue('');
+                return;
+              }
+            }
+            setDisplayDropdown(false);
+            setInputValue('');
+            onSearchResultsToggle(completeHashMatch);
+            return;
+          }
+        }
+
         if (thisRequestId !== requestsCounterRef.current) return;
         // The user edited the text since the request started.
         // Let's ignore the result then.
@@ -137,7 +169,7 @@ export default function SearchInputAndResults({
         setRecentRevisions(null);
       }
     },
-    [repository],
+    [repository, onSearchResultsToggle, searchType, displayedRevisions],
   );
 
   const debouncedSearchRecentRevisions = useCallback(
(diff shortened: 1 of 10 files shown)


src/components/Search/SearchInputAndResults.tsx
```diff
@@ -66,8 +68,15 @@ export default function SearchInputAndResults({
     }
   }, []);
 
+  // Helper to check if the search term matches a full 40 character revision hash
+  const isFullRevisionHash = (searchTerm: string) =>
```
beatrice-acasandrei: Since we are permitting partial searches (by short hash), this improvement should also apply to short hashes, not solely to full revision hashes.

Note: This request improvement was added by me in bugzilla, but it is actually something that came up in our latest feedback google form. It's missing some details and it was anonymous, so we'll need to make some decisions regarding the behaviour.

@kala-moz What do you think?

kala-moz:
@moijes12 I've run out of time today so I'll review this tomorrow!

moijes12:
> @moijes12 I've run out of time today so I'll review this tomorrow! 

Sure Kala.

src/components/Search/SearchInputAndResults.tsx
kala-moz: @beatrice-acasandrei I agree with your comment on the bug. The behavior should remain the same except now the checkbox is checked and the revision is shown in the component after the hash is pasted. This should also work for short hashes.

kala-moz (changes requested):
Thanks for taking this on! Right now this somewhat works for full hashes. The problem is that after pasting the full hash, the revision dropdown doesn't appear below the search input so the user can't change their selection. And if the user manually deletes the revision from the view, the revision dropdown still doesn't appear. The user is stuck with their original selection. We'll have to investigate why this happens. And like Beatrice said, this should also work for short hashes.

moijes12:
Thanks @kala-moz and @beatrice-acasandrei 

I will work on this and add the changes to the PR

moijes12 (commented):
Thanks @kala-moz

I have added some unit tests for the functionality to select full hash.

For short hash, is there any character length that I need to look at ? From what I see in the commit selection box, the hash has a length of 12 characters. Should I use that ?

kala-moz:
> Thanks @kala-moz
> 
> I have added some unit tests for the functionality to select full hash.
> 
> For short hash, is there any character length that I need to look at ? From what I see in the commit selection box, the hash has a length of 12 characters. Should I use that ?

@moijes12 No, use the `truncateHash` helper in `src/utils/helper`

For example:
```
  const shortHashMatch = results.find(
        (rev) =>
          truncateHash(rev.revision).toLowerCase() ===
          searchTerm.toLowerCase(),
      );
```

src/components/Search/SearchInputAndResults.tsx
```diff
@@ -176,12 +213,13 @@ export default function SearchInputAndResults({
   return (
     <Box ref={containerRef}>
       <SearchInput
+        value={inputValue}
         onFocus={() => setDisplayDropdown(true)}
```
kala-moz: So, we still have the problem of the results drop-down not appearing when the user clicks on the input after pasting the full hash. Also, if the user deletes the selected revision, the results drop-down also doesn't appear when the user clicks the input. Try this to fix the problem and get the full list. 

```
onFocus={() => {
  setDisplayDropdown(true);
  if (!inputValue) {
    void searchRecentRevisions('');
  }
}}

```

src/components/Search/SearchInputAndResults.tsx
```diff
@@ -116,6 +125,29 @@ export default function SearchInputAndResults({
...
+          const completeHashMatch = results.find(
+            (rev) => rev.revision.toLowerCase() === searchTerm.toLowerCase(),
+          );
+          if (completeHashMatch) {
+            if (searchType === 'new') {
+              const alreadySelected = displayedRevisions.some(
+                (rev) => rev.id === completeHashMatch.id,
+              );
+              if (alreadySelected) {
+                setDisplayDropdown(false);
```
kala-moz: Don't set the display to false because that's not the desired behavior after a selection. So please delete this.

src/components/Search/SearchInputAndResults.tsx
```diff
@@ -116,6 +125,29 @@ export default function SearchInputAndResults({
...
+                setInputValue('');
+                return;
+              }
+            }
+            setDisplayDropdown(false);
```
kala-moz: Same here. Delete this please.

kala-moz (changes requested):
Hi Alex, thanks for the updates. We still have to work on the behavior because we don't want to hide the drop-down nor clear the input after the hash is pasted and the revision selected. We also need to be able to get the full list of revisions after a selected revision is deleted. I've added some suggestions throughout.

src/components/Search/SearchInputAndResults.tsx
```diff
@@ -116,6 +125,29 @@ export default function SearchInputAndResults({
...
+            (rev) => rev.revision.toLowerCase() === searchTerm.toLowerCase(),
+          );
+          if (completeHashMatch) {
+            if (searchType === 'new') {
+              const alreadySelected = displayedRevisions.some(
+                (rev) => rev.id === completeHashMatch.id,
+              );
+              if (alreadySelected) {
+                setDisplayDropdown(false);
+                setInputValue('');
```
kala-moz: Don't clear the input value here either. That's also not the desired behavior

src/components/Search/SearchInputAndResults.tsx
```diff
@@ -116,6 +125,29 @@ export default function SearchInputAndResults({
...
+                return;
+              }
+            }
+            setDisplayDropdown(false);
+            setInputValue('');
```
kala-moz: Same here. Please delete.

moijes12:
Thanks @kala-moz 

This feedback is great. I will work on this and get it done

moijes12:
@kala-moz @beatrice-acasandrei I have implemented some of the review comments. I am moving this PR back to draft till I complete work on this.

moijes12:
Will be covered in #1053

PR #1027, opened 2026-04-09
Added the optional param enable_silverman_kde to loaders and treeherder logic

kala-moz:
Quick PR to enable the Silverman KDE for perfcompare: 

[Deploy link](https://deploy-preview-1027--mozilla-perfcompare.netlify.app/compare-results?baseRev=215a333dede39c4849a3ccbe28c58fd827bb1f7c&baseRepo=try&newRev=3ca79c893ced2264860775a789dc51146cfcc7ca&newRepo=try&framework=13)

Screenshot 2026-04-09 at 12 14 30 PM

### DIFF
--- a/src/components/CompareResults/loader.ts
+++ b/src/components/CompareResults/loader.ts
@@ -26,6 +26,7 @@ export function checkValues({
   framework,
   replicates,
   testVersion,
+  silvermanKDEEnabled,
 }: {
   baseRev: string | null;
   baseRepo: Repository['name'] | null;
@@ -34,6 +35,7 @@ export function checkValues({
   framework: string | number | null;
   replicates: boolean;
   testVersion: TestVersion | null;
+  silvermanKDEEnabled: boolean;
 }): {
   baseRev: string;
   baseRepo: Repository['name'];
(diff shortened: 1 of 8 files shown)


src/components/CompareResults/loader.ts
```diff
@@ -83,6 +86,7 @@ export function checkValues({
 
   if (testVersion === MANN_WHITNEY_U || testVersion === null) {
     replicates = true;
+    silvermanKDEEnabled = true;
```
gmierz: We'd want this to be false since we don't want to enable it by default. Actually, we might not want to set this here since it would always force the setting into this value right?

src/components/CompareResults/overTimeLoader.ts
```diff
@@ -108,6 +111,7 @@ function checkValues({
 
   if (testVersion === MANN_WHITNEY_U || testVersion === null) {
     replicates = true;
+    silvermanKDEEnabled = true;
```
gmierz: I think the same is true here.

gmierz (approved):
Works great now, thanks for the quick patch! r+

PR #1029, opened 2026-04-17
Bug 2032537:  show count of the improvements and regressions in the subtests per platform 

kala-moz:
Solution for [Bug 2032537](https://bugzilla.mozilla.org/show_bug.cgi?id=2032537)
[JIRA](https://mozilla-hub.atlassian.net/browse/PCF-738)

Updated:

In this PR, I created a custom hook called `useSubtestRegressionCount` to find, for a given test block, every row that has subtests; fetch those subtests from the treeherderAPI; and return the counts of regressions and improvements for each platform. In `treeherder.ts`, added memoized fetches for the subtests to improve performance and avoid unnecessary refetching. `RevisionRow` supplies the hook with the props to display the number of regressions and improvements for the subtests present in each row of the block. 

Note: need to add test coverage next. So in progress. 

[Deploy link](https://deploy-preview-1029--mozilla-perfcompare.netlify.app/compare-results?baseRev=f850e62a23fb0f218c56ff26cdcf77aedf56be4f&baseRepo=mozilla-central&newRev=ec1b1f3e6d3c46f2260b43fd6ca3552498060ac4&newRepo=mozilla-central&framework=13&search=speedometer&test_version=mann-whitney-u)

Screenshot 2026-04-22 at 6 23 53 PM

### DIFF
--- a/src/__tests__/CompareResults/RevisionRow.test.tsx
+++ b/src/__tests__/CompareResults/RevisionRow.test.tsx
@@ -7,10 +7,22 @@ import { compareView } from '../../common/constants';
 import { isDistributionNormal } from '../../common/testVersions/mannWhitney';
 import { loader } from '../../components/CompareResults/loader';
 import RevisionRow from '../../components/CompareResults/RevisionRow';
+import { useSubtestRegressionCount } from '../../hooks/useSubtestRegressionCount';
 import { CompareResultsItem, MannWhitneyResultsItem } from '../../types/state';
 import { Platform } from '../../types/types';
 import getTestData from '../utils/fixtures';
 import { screen, renderWithRouter } from '../utils/test-utils';
+
+jest.mock('../../hooks/useSubtestRegressionCount');
+const mockUseSubtestRegressionCount = useSubtestRegressionCount as jest.Mock;
+
+// Default to no subtest counts so existing tests are unaffected.
+beforeEach(() => {
+  mockUseSubtestRegressionCount.mockReturnValue({
+    counts: null,
+    isLoading: false,
+  });
+});
 
 function renderWithRoute(component: ReactElement) {
   fetchMock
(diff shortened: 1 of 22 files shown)


mgaudet:
1. I like the general thrust here. 
2. I tried with [this](https://deploy-preview-1029--mozilla-perfcompare.netlify.app/compare-lando-results?baseLando=191622&newLando=191624&baseRepo=try&newRepo=try&framework=13) link; I found it took quite a long time for the SP3 results to show up
3. I do wonder about grouping all the platforms together a bit; did you try per-platform improve/regress boxes and found it too noisy?

kala-moz:
> 1. I like the general thrust here.
> 
>     2. I tried with [this](https://deploy-preview-1029--mozilla-perfcompare.netlify.app/compare-lando-results?baseLando=191622&newLando=191624&baseRepo=try&newRepo=try&framework=13) link; I found it took quite a long time for the SP3 results to show up
> 
>     3. I do wonder about grouping all the platforms together a bit; did you try per-platform improve/regress boxes and found it too noisy?

@mgaudet I'm glad you brought up #3 because I wasn't too sure if it was per platform or all the platforms together. I can update it to go by platform and see how it looks. Hmm, I'll look over #2.

gmierz (changes requested):
Great change! Some things I noticed as I was testing:
(1) It looks like the spacing between the rows on the result page are larger. Is this expected?
(2) Looking at the network tab, when I am loading the page with MWU on the results page, I see the subtest results being fetched twice: once when we're fetching the main results, and then a second time once the main results are obtained. Seems like a bug in the implementation somewhere and will need to be fixed.
(3) Sometimes it's unclear which row the tags that show the regressions/improvements are associated with (is it the row above it or below it). Is there something we can do to make it clearer that it's the row below them?

I think that if we're starting to fetch the subtest results in the background, we should get a bug filed to use those results in the subtest page too instead of doing another request for the same data when we navigate to the subtests.

kala-moz:
@gmierz Thanks for your feedback! 

1.) and 3.) The extra spacing is because of the regressions and improvement pills' placements. I've moved them inside the revision row so it's clearer which pills belongs to which row and to preserve the original spacing of the rows. 

Screenshot 2026-04-22 at 6 23 53 PM




2.) I'm not seeing the subtests call for the main results fetch? We shouldn't be fetching the subtests when the main results are being fetched. The only subtests calls I see are for the rows with unique subtests that I'm fetching to count the improvements and regressions. Since that's a lot of unique calls, I could batch them into one request? Also could you please show me exactly where you see the subtests fetch for the main results call, maybe with screenshots?

Screenshot 2026-04-22 at 3 37 06 PM

gmierz (approved):
r+ great work! The fix you made also seemed to make the caching work when switching the test versions too. :)

PR #1048, opened 2026-06-10
Add median diff percentage to blue blurb in expanded view

kala-moz:
There was request to add the median percentage to the blue median difference blurb in the expanded section. Please check that the math is correct.


kala-moz:
Closing since this PR finished the task : https://github.com/mozilla/perfcompare/pull/1050

PR #1050, opened 2026-06-10
Assorted fixes to improve the app when numbers are very large and not measuring time.

padenot:
If the numbers are large (e.g. we're measuring gigabytes), the math would just break -- this is what the first commit fixes.

The second commit slightly improves the confidence in the case which the distribution are skewed, which is typical in our case. This is lifted from scipy, cross-checked like the other bits of code.

Then the rest are cosmetic / fixes (in order of the patches):

- it's nicer to say 800MB than 800000000 bytes
- I thought that it was confusing when displaying the median diff: the test could be marked as high confidence, but the media diff bootstrap could be marked as low confidence. Conceptually this makes sense, but we can phrase it better: we want to warn when a confidence interval includes zero, e.g. "the median was improved by 0.5ms, confidence interval [0.6, -0.1], meaning this could well be a -0.1 regression. If this interval does not include 0 (e.g. 0.5ms, confidence interval [0.8, 0.3], then in all cases we're certain it improves things, there's low chance it regresses things
-  again when displaying large numbers such as amount of memory, our auto-sizing for the KDE was perfectible, I saw myself having to resize. I changed things so it looks better (leaving about 10% of padding on either side of the min/max). Also when the distribution is all over the place, I don't think the automatic bandwidth calculation makes any sense, so we add a slider much like we added one for the modal split.

This fixes BMO https://bugzilla.mozilla.org/show_bug.cgi?id=2046147

### DIFF
--- a/src/common/testVersions/mannWhitney.tsx
+++ b/src/common/testVersions/mannWhitney.tsx
@@ -415,16 +415,30 @@ export const mannWhitneyStrategy = {
       baseRuns.length > 0 && newRuns.length > 0
         ? bootstrapMedianDiffCI(baseRuns, newRuns)
         : null;
-    const rawUnit = mwResult.base_measurement_unit ?? mwResult.new_measurement_unit ?? 'ms';
+    const rawUnit =
+      mwResult.base_measurement_unit ?? mwResult.new_measurement_unit ?? 'ms';
     const { fmt, displayUnit } = ci
       ? adaptUnit([ci.medianDiff, ci.ciLow, ci.ciHigh], rawUnit)
       : adaptUnit([], rawUnit);
     const ciCrossesZero = ci && ci.ciLow < 0 && ci.ciHigh > 0;
+    const baseMedian = (() => {
+      if (!baseRuns.length) return null;
+      const s = [...baseRuns].sort((a, b) => a - b);
+      const m = Math.floor(s.length / 2);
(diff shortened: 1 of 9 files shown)


src/common/testVersions/mannWhitney.tsx
```diff
@@ -415,12 +415,19 @@ export const mannWhitneyStrategy = {
...
-    const fmt = (n: number) => (n >= 0 ? '+' : '') + n.toFixed(1);
+    const rawUnit =
+      mwResult.base_measurement_unit ?? mwResult.new_measurement_unit ?? 'ms';
+    const { fmt, displayUnit } = ci
+      ? adaptUnit([ci.medianDiff, ci.ciLow, ci.ciHigh], rawUnit)
+      : adaptUnit([], rawUnit);
+    const ciCrossesZero = ci && ci.ciLow < 0 && ci.ciHigh > 0;
     const summary = ci ? (
       <span>
-        <strong>Δ median</strong> = {fmt(ci.medianDiff)} ms 95% CI [
```
kala-moz, line 421: Ah nice catch. Do you mind calculating the median diff percentage here as well and include it in the summary please?

kala-moz (approved):
Just have one ask, if you don't mind. Otherwise, good to go after we rebase - the tests will go green too. Thank you!

PR #1051, opened 2026-06-11
Bug-1855186: Show whether lower or higher is "better" for each metric

sumairq:
https://bugzilla.mozilla.org/show_bug.cgi?id=1855186
### Problem

  Fixes **"Rows do not show if 'lower' or 'higher' metrics are better."** Results carry a `lower_is_better` flag, but it was used only for sorting and never shown — so a reader couldn't tell from the numbers whether a delta was an improvement or a regression. The old Treeherder view had a "Better: Lower/Higher" column; this restores that signal.

  ### What changed

  A muted inline label — **"Lower is better" / "Higher is better"** — now appears next to the metric name:

  - **Main results table:** once per test, in the test header (`TestHeader`).
  - **Subtests view:** once in the subtests header (`SubtestsRevisionHeader`), next to the suite name.

  New `BetterDirectionIndicator` component renders the label from `lower_is_better`, and renders nothing when it's `null`/`undefined` (Mann-Whitney can be null) instead of a misleading direction. Display-only — no loader/API/data-model changes; works in both Student's-t and Mann-Whitney U views.

### Screenshots

 | Results view (Multiple Revisions) | Results view (Single Revision) |
  | --- | --- |
  | image | image |

| Subtests view |
  | --- |
  | Screenshot 2026-06-11 094730 |

  ### Key decisions

  - **Shown once per group, never per row.** `lower_is_better` is a suite/metric property, so it's identical across rows. A per-row version was redundant and (being `nowrap`) widened the "Subtests" grid column and pushed others right. Header placement fixes both and matches `TestHeader`.
  - **No new column / no grid changes.** The label is inline in existing header markup; column widths are untouched.
  - **Status column left as-is** — it already covers "scan for regressions"; this adds only the direction signal.
  - **Typing.** `HeaderProperties` picks from `CombinedResultsItemType`, so `lower_is_better` resolves to `boolean | null` automatically (no intersection workaround); `SubtestsRevisionsHeader` is built from `CompareResultsItem`, so it stays `boolean`.
  - **Design tokens used** — `FontSizeRaw.Small` and `Spacing.xSmall` from `src/styles`.

  ### Testing

  `summarizeVisibleRows` strips the label from title parsing so the ~60 filtering/sorting expectations stay focused on row data; the label itself is covered by a dedicated assertion and snapshots.

### DIFF
--- a/src/components/CompareResults/BetterDirectionIndicator.tsx
+++ b/src/components/CompareResults/BetterDirectionIndicator.tsx
@@ -1,0 +1,40 @@
+import Box from '@mui/material/Box';
+
+import { Strings } from '../../resources/Strings';
+import { FontSizeRaw, Spacing } from '../../styles';
+
+interface BetterDirectionIndicatorProps {
+  lowerIsBetter: boolean | null | undefined;
+}
+
+// The direction comes from the result's
+// `lower_is_better` field. When that field is unknown (it can be null for
+// Mann-Whitney results), we render nothing rather than a misleading label.
+function BetterDirectionIndicator({
+  lowerIsBetter,
+}: BetterDirectionIndicatorProps) {
+  if (lowerIsBetter === null || lowerIsBetter === undefined) {
+    return null;
+  }
+
+  const label = lowerIsBetter
+    ? Strings.components.betterDirection.lower
+    : Strings.components.betterDirection.higher;
+
+  return (
+    <Box
+      component='span'
+      data-testid='better-direction-indicator'
+      sx={{
+        color: 'text.secondary',
+        fontSize: FontSizeRaw.Small.fontSize,
+        whiteSpace: 'nowrap',
+        marginInline: `${Spacing.xSmall}px`,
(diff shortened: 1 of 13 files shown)


src/components/CompareResults/BetterDirectionIndicator.tsx
```diff
@@ -0,0 +1,40 @@
...
+    ? Strings.components.betterDirection.lower
+    : Strings.components.betterDirection.higher;
+
+  return (
+    <Box
+      component='span'
+      data-testid='better-direction-indicator'
+      sx={{
+        color: 'text.secondary',
+        fontSize: FontSizeRaw.Small.fontSize,
```
kala-moz: Let's use `FontSizeRaw.Normal` instead.

kala-moz (changes requested):
Thanks, Sumair! Just one small change and we can merge.

kala-moz (approved):
Thank you!

PR #1045, opened 2026-06-01
Bug: 2044194 Port client-side KDE mode detection from kde-widget into CommonGraph

kala-moz:
[See Bug 2044194](https://bugzilla.mozilla.org/show_bug.cgi?id=2044194)

This PR adds client-side KDE mode detection by porting the pure-logic helpers (`fitModesFromKde, areaFracs, assignLetters, matchModes, splitByMode`) from the [`kde-widget `](https://github.com/padenot/perfcompare-new-stats/blob/main/standalone-stats/kde-widget.js) into `src/utils/kde.js`. See Paul's example [here](https://paul.cx/public/example-output.html).

It wires those helpers into `CommonGraph` so detected peaks render as labelled vertical markers on the KDE chart (e.g. "Base A: 12.34 (45%)"). Also adds the interactive valley-depth slider — with the threshold state lifted to `RevisionRowExpandable` so the upcoming mode-blurb panel can read the same modes without recomputing the KDE. 

Includes unit tests for the new pure functions and updated CommonGraph tests covering the slider props and mode-overlay series. 

Note: There's still more work to do like the upcoming blurb panel but didn't want to overload this PR by touching too many different files.

[
Deploy Link](https://deploy-preview-1045--mozilla-perfcompare.netlify.app/compare-results?baseRev=29a14637a9d81a23621ff9e0e555a4f958aca482&baseRepo=mozilla-central&newRev=ee0e0aa2307fae09742d269717b416b918b1e96e&newRepo=mozilla-central&newRev=59a2ff528e3bee0556fc2dfac7b3bee390367e70&newRepo=mozilla-central&newRev=cc299aa6a1cb20497aee5004231a0c0a0be1d0d1&newRepo=mozilla-central&framework=1&filter_status=improvement%2Cregression)

Screenshot 2026-06-09 at 09 42 30

PR #1052, opened 2026-06-12
Add KdeModesPanel to expanded view

kala-moz:
This PR fulfills two bugs: 

1. [Bug 2044194 Port client-side KDE mode detection from kde-widget into CommonGraph](https://bugzilla.mozilla.org/show_bug.cgi?id=2044194)
2. [Bug 2032241 Results Page: Add a statement in the expanded row on modes found and changes](https://bugzilla.mozilla.org/show_bug.cgi?id=2032241)

See #1045 for more info.  

[Deploy Link](https://deploy-preview-1052--mozilla-perfcompare.netlify.app/compare-results?baseRev=8de2dede9c34d18332b9bcff08bda4e6ff103b30&baseRepo=try&newRev=119723724e68b7d3799fa8c89216b36264e01c69&newRepo=try&newRev=26e995970eaabbc908c630aad3ab0c50f47e5b44&newRepo=try&framework=4&test_version=mann-whitney-u&filter_status=improvement%2Cregression)

Screenshot 2026-06-16 at 12 49 47


src/components/CompareResults/KdeModesPanel.tsx
```diff
@@ -0,0 +1,433 @@
...
+// Significance test ported from kde-widget.js: CI sits entirely on one side
+// AND isn't the degenerate [0, 0] interval.
+function isSignificant(ciLow: number, ciHigh: number): boolean {
+  return (
+    (ciHigh <= 0 && (ciLow < 0 || ciHigh < 0)) ||
+    (ciLow >= 0 && (ciLow > 0 || ciHigh > 0))
+  );
+}
+
+// fast / mid / slow path label by letter rank (A = fastest).
```
padenot: I think we should move away from language like "fastest" "slowest" because as we've seen in the last PR, it could be measuring e.g. "bytes" (memory consumption).

src/components/CompareResults/KdeModesPanel.tsx
```diff
@@ -0,0 +1,433 @@
...
+function pathLabel(letter: string, totalModes: number): string | null {
+  const rank = letter.charCodeAt(0) - 65;
+  if (totalModes <= 1) return null;
+  if (rank === 0) return 'fast path';
+  if (rank === totalModes - 1) return 'slow path';
```
padenot: e.g. here.

src/components/CompareResults/KdeModesPanel.tsx
```diff
@@ -0,0 +1,433 @@
...
+  if (!ci) return <Box sx={{ color: palette.subtle }}>no CI available</Box>;
+  const color = sig
+    ? ci.medianDiff < 0
+      ? palette.ok
+      : palette.bad
+    : palette.muted;
+  const arrow = sig
+    ? ci.medianDiff < 0
+      ? '▼ faster'
+      : '▲ slower'
```
padenot: Maybe just "better" and "worse" will do, but be careful: we need to make all this "lower is better" or "higher is better" aware.

src/components/CompareResults/KdeModesPanel.tsx
```diff
@@ -0,0 +1,433 @@
...
+    const newLoc = nModes.peakLocs[modeIdx];
+    const nearestBase = bModes.peakLocs.reduce((a, b) =>
+      Math.abs(b - newLoc) < Math.abs(a - newLoc) ? b : a,
+    );
+    return { modeIdx, improved: newLoc < nearestBase };
+  });
+
+  return {
+    pairs,
+    improvements: pairs.filter((p) => p.sig && p.ci && p.ci.medianDiff < 0),
```
padenot: e.g. here we're unconditionally doing `medianDiff < 0` -> improvement, which isn't quite true, e.g. if we're measuring a "Score" (such as a number of iteration per second or a score or whatever).

src/components/CompareResults/KdeModesPanel.tsx
```diff
@@ -0,0 +1,433 @@
...
+};
+
+function computeBlurb(
+  baseValues: number[],
+  newValues: number[],
+  vt: number,
+  isSubtest: boolean,
+): Blurb | null {
+  const bKde = safeKde(baseValues, bandwidthFor(baseValues, isSubtest));
+  const nKde = safeKde(newValues, bandwidthFor(newValues, isSubtest));
```
padenot: Not sure we want to use `bandwidthFor` here, because that doesn't have the bandwidth multipler. We could be in a situation where the blurb says there are 3 modes but there are only 2 vertical lines on the chart. Can we make this use the same bandwidth?

src/components/CompareResults/KdeModesPanel.tsx
```diff
@@ -0,0 +1,433 @@
...
+  text: string;
+  kind: 'improvement' | 'regression' | 'mixed' | 'neutral';
+};
+
+// Significance test ported from kde-widget.js: CI sits entirely on one side
+// AND isn't the degenerate [0, 0] interval.
+function isSignificant(ciLow: number, ciHigh: number): boolean {
+  return (
+    (ciHigh <= 0 && (ciLow < 0 || ciHigh < 0)) ||
+    (ciLow >= 0 && (ciLow > 0 || ciHigh > 0))
```
padenot, line 45: I this we need this <= and => in the unlikely case we land on 0.

src/utils/kdeAnalysis.ts
```diff
@@ -0,0 +1,139 @@
...
+ * and letter labels. Returns the empty ModeInfo (rather than null) when there's
+ * nothing to fit — callers check `peakLocs.length` to detect empty.
+ */
+export function computeModeInfo(
+  x: number[],
+  y: number[],
+  vt: number,
+): ModeInfo {
+  if (!x.length || !y.length) return EMPTY_MODE_INFO;
+  const { peakLocs, boundaries } = fitModesFromKde(x, y, vt);
```
padenot, line 131: Don't we compute the KDE twice here? This is also the cause for our dispcrepency between bandwidth, causing potential confusion in mode count.

src/components/CompareResults/KdeModesPanel.tsx
```diff
@@ -0,0 +1,433 @@
...
+function computeBlurb(
+  baseValues: number[],
+  newValues: number[],
+  vt: number,
+  isSubtest: boolean,
+): Blurb | null {
+  const bKde = safeKde(baseValues, bandwidthFor(baseValues, isSubtest));
+  const nKde = safeKde(newValues, bandwidthFor(newValues, isSubtest));
+  if (!bKde || !nKde) return null;
+  const bModes = computeModeInfo(Array.from(bKde.x), Array.from(bKde.y), vt);
```
padenot: Those are already arrays aren't they?

padenot (changes requested):
Some tweaks needed, but generally the right direction.

src/components/CompareResults/KdeModesPanel.tsx
kala-moz: Yes, good catch. I was using the language from your widget but forgot to consider the other non-time based metrics

kala-moz: Replaced them all with better/worse framing

src/components/CompareResults/KdeModesPanel.tsx
kala-moz: We need this, right? I want to make sure I understand your comment @padenot

src/components/CompareResults/KdeModesPanel.tsx
kala-moz: Hmm, true. Done.

src/components/CompareResults/KdeModesPanel.tsx
kala-moz: Ah, yes you're right. I forgot to drop them from an earlier version.

src/utils/kdeAnalysis.ts
kala-moz: There's no longer a bandwidth discrepancy due to this [commit](https://github.com/mozilla/perfcompare/pull/1052/commits/ec7368c8ae7718beadd0469a1a381f72c8272e41).

src/components/CompareResults/KdeModesPanel.tsx
padenot: sorry yes, typo, I meant: "I think we need this ..."

padenot (approved):
Thanks!

PR #1056, opened 2026-06-18
[ON HOLD] Bug 2032243: Use the confidence interval for the significance column

kala-moz:
The significance column, its filter, its sort, and the expanded-row alert now all read from a single precomputed bootstrap (BCa) CI on the difference of medians, replacing the Mann-Whitney p-value's `interpretation` string as the source of truth.

~~Added `MannWhitneyResultsItem.bootstrapCi` and a `precomputeMannWhitneyCI` helper that the four data loaders (compare, over-time, subtests, subtests-over-time) call once per row. Doing it at load time keeps sort/filter/render at property-access speed.~~

- Sig column: `matchesFunction` and `sortFunction` read `bootstrapCi`. Sort is "significant first, then |medianDiff| desc" — written in ASC semantics so `useTableSort`'s DESC swap produces that order. The cell render uses `bootstrapCi?.significant` to pick icon vs `-`
- `renderExpandedRight` reads the same precomputed CI (with an inline fallback for callers that mount the strategy without a loader).
- Drop the "Significance (p-value)" row from `PValCliffsDeltaComp`. The Δ-median alert already conveys CI-based significance, and showing the p-value contradicts the Sig column.
- Update `tooltipSignificance` and `tooltipStatusMannWhitney` to match the new definition (BCa CI excludes zero, not p-value < 0.05).
- Update tests + snapshots to reflect the new sig values and orderings; remove two RevisionRow tests that asserted text from the now-removed p-value row.

[Deploy Link](https://deploy-preview-1056--mozilla-perfcompare.netlify.app/compare-results?baseRev=8de2dede9c34d18332b9bcff08bda4e6ff103b30&baseRepo=try&newRev=119723724e68b7d3799fa8c89216b36264e01c69&newRepo=try&newRev=26e995970eaabbc908c630aad3ab0c50f47e5b44&newRepo=try&framework=4&test_version=mann-whitney-u)

[Browser time preview](https://deploy-preview-1056--mozilla-perfcompare.netlify.app/compare-results?baseRev=51308dbe0e7e0810c9f6eec28ce49d34c49d6480&baseRepo=mozilla-central&newRev=64a89bfa4912122c11d6d2d381bd4c8d538b56e6&newRepo=mozilla-central&framework=13&test_version=mann-whitney-u&search=speedometer)

### **Updates:** 

1. Now considers replicates when available. 
2. Changed from symbols to letters - S for Significant. NS for Not Significant 
[...]

PR #1059, opened 2026-06-23
[ON HOLD] Bug 2024208: Add a column to denote a magnitude of change based on the modality analysis

kala-moz:
The Median Diff (%) column has been replaced with a Mode Δ (%) column that reports the largest matched-pair peak shift between the Base and New distributions, derived from the same KDE + mode-detection + modality-matching pipeline that powers the chart and the expanded-row blurb.

~~Like `bootstrapCi`, the value is precomputed once in the four data loaders (`loader.ts`, `overTimeLoader.ts`, `subtestsLoader.ts`, `subtestsOverTimeLoader.tsx`) via `precomputeLargestPeakShift` and attached to `result.modeDeltaPct`. KDE + mode matching is expensive enough that running it on every sort comparison or render would tank interactions.~~

[Speedometer 3 preview link](https://deploy-preview-1059--mozilla-perfcompare.netlify.app/subtests-compare-results?baseRev=51308dbe0e7e0810c9f6eec28ce49d34c49d6480&baseRepo=mozilla-central&newRev=64a89bfa4912122c11d6d2d381bd4c8d538b56e6&newRepo=mozilla-central&framework=13&baseParentSignature=5854576&newParentSignature=5854576&test_version=mann-whitney-u)

I had some trouble with netlify thus all the open / reopen commands. :/ 

**Update:** The Mode Δ (%) column and the Distribution Interpretation row in `MannWhitneyCompareMetrics` were reading mode counts from two independent sources: the client-side ISJ/SJ KDE pipeline vs. the backend's wider Silverman KDE. Therefore, could could disagree in jarring ways (e.g. "Mode Δ = 4.16%" sitting directly above "No modes or data, possible oversmoothing"). Now, both views are piped through a single precomputed analysis so they can't drift.

**Update 2:** I precompute modality on replicates when available

**Update 3:** Flip the sortFunction to ASC semantics so the framework's swap produces the intuitive direction: DESC → biggest improvements first, ASC → biggest regressions first. Comment updated to spell this out (same convention the Sig column already uses).

**Update 4:** Lazy-precompute Mode Δ analysis to restore load speed. All four loaders no longer call `precomputeModalityAnalysis`

**Update 5:**  Select by peak magnitude vs largest shift. Also, simplified the language of the tooltip for the column. 

Screenshot 2026-06-24 at 16 53 25


padenot:
Similarly to https://github.com/mozilla/perfcompare/pull/1056#issuecomment-4789858578 I only see `-` ?

kala-moz:
> Similarly to [#1056 (comment)](https://github.com/mozilla/perfcompare/pull/1056#issuecomment-4789858578) I only see `-` ?

I've decided to also remove the dash since it doesn't give enough information. I've switched to `NM` which mean `No Modes`. I've updated the tooltip with more information. 

Ahhh, I didn't use replicates. That's were the bug was. [Updated subtests preview](https://deploy-preview-1059--mozilla-perfcompare.netlify.app/subtests-compare-results?baseRev=51308dbe0e7e0810c9f6eec28ce49d34c49d6480&baseRepo=mozilla-central&newRev=64a89bfa4912122c11d6d2d381bd4c8d538b56e6&newRepo=mozilla-central&framework=13&baseParentSignature=5854576&newParentSignature=5854576&test_version=mann-whitney-u)

[Mozilla central preview](https://deploy-preview-1059--mozilla-perfcompare.netlify.app/compare-results?baseRev=51308dbe0e7e0810c9f6eec28ce49d34c49d6480&baseRepo=mozilla-central&newRev=64a89bfa4912122c11d6d2d381bd4c8d538b56e6&newRepo=mozilla-central&framework=13&test_version=mann-whitney-u&search=speedometer&sort=significance%7Casc)

padenot:
There's quite a lot of room in that column, looking at your screenshot, we can say the words in full? Similarly for NM

padenot:
I was browsing around the deploy link you helpfully provided, and I found: 
Screenshot 2026-07-21 at 11 30 17

Here, we have an example of a test that's a bit noisy. We see the main peak doesn't move, but we report a few percent shift. Two thoughts:

- If we go this route, we should take the two peaks of the highest magnitude probably?
- The raw MWU more or less does the right thing already, but is susceptible to outliers

What was the intent behind this? Maybe I'm not looking at cases that greatly benefit from this?

kala-moz:
> There's quite a lot of room in that column, looking at your screenshot, we can say the words in full? Similarly for NM

1.) Say what words in full? 
2.) Okay, I see we do have space to write No Modes vs NM.

kala-moz:
> I was browsing around the deploy link you helpfully provided, and I found: Screenshot 2026-07-21 at 11 30 17
> 
> Here, we have an example of a test that's a bit noisy. We see the main peak doesn't move, but we report a few percent shift. Two thoughts:
> 
>     * If we go this route, we should take the two peaks of the highest magnitude probably?
> 
>     * The raw MWU more or less does the right thing already, but is susceptible to outliers
> 
> 
> What was the intent behind this? Maybe I'm not looking at cases that greatly benefit from this?

@padenot Hmm. I see what you mean. I can switch it to select by peak magnitude. The current selection by shift size was a suggestion from sparky in [Bug 2024208](https://bugzilla.mozilla.org/show_bug.cgi?id=2024208) so not set in stone. We wanted to have a magnitude of change that didn't rely on mean / median alone but took advantage of the modality work. But using peak magnitude makes much more sense. 

[Updated deploy link](https://deploy-preview-1059--mozilla-perfcompare.netlify.app/subtests-compare-results?baseRev=51308dbe0e7e0810c9f6eec28ce49d34c49d6480&baseRepo=mozilla-central&newRev=64a89bfa4912122c11d6d2d381bd4c8d538b56e6&newRepo=mozilla-central&framework=13&baseParentSignature=5854746&newParentSignature=5854746&test_version=mann-whitney-u)

Screenshot 2026-07-21 at 19 41 02

kala-moz:
I will be closing this PR until I've determined whether this col is needed after we've released the simplified view.

PR #1074, opened 2026-07-31
Remove canova from the codeowners

canova:
It's been a while since I needed to look at the dependency updates, and looks like @kala-moz and everyone are doing a great job at handling them! I don't think my strict ping/review request is need anymore but of course I'm more than happy to help in case you need anything from me! So please just ping me for anything!


kala-moz (approved):
Thanks, Nazim!

PR #1076, opened 2026-08-10
Bug 2058240: Filters from cookies don't make it into the URL

kala-moz:
[Bug 2058240](https://bugzilla.mozilla.org/show_bug.cgi?id=2058240)

[Deploy Link](https://deploy-preview-1076--mozilla-perfcompare.netlify.app/compare-results?baseRev=f4adac7285060421005f0c3e730b43d5fc7eaaa2&baseRepo=try&newRev=ec8c1e75812cf175d8a10b63974cf7b7c5f4588b&newRepo=try&framework=13&advanced_columns=cliffs_delta&initialized=1)

Filters and sort were persisted to both the URL and cookies, and read as `urlParam ?? cookie` per column. So opening a shared link fell back to the recipient's own cookies for any column/sort not spelled out in the URL — the same URL showed different results for different people.

Introduce an explicit "initialized" URL marker:

- On the first load of an uninitialized URL, seed the remembered filter/sort cookies into the URL and stamp it `initialized=1` (via `replaceState`, so no rerender or loader refetch).

- Once initialized, filters/sort are read only from the URL — a missing param means the default, never a cookie. Cookies are still written on change so the memory survives for the next fresh visit.

Also read the live URL params when writing filter/sort changes, instead of a memoized snapshot, so an out-of-band param (the seeded marker) isn't clobbered and dropped by a later toggle.

Centralize the query-param/cookie keys and the marker helpers in `tableStatePersistence.ts`. Add tests covering: a fresh URL seeds from cookies and gains the marker, an initialized URL ignores cookies, and the marker survives a filter toggle.

Co-Authored-By: Claude Opus 4.8

### DIFF
--- a/src/components/CompareResults/ResultsTable.tsx
+++ b/src/components/CompareResults/ResultsTable.tsx
@@ -11,6 +11,7 @@ import TableContent from './TableContent';
 import TableHeader from './TableHeader';
 import { MANN_WHITNEY_U } from '../../common/constants';
 import useAdvancedColumns from '../../hooks/useAdvancedColumns';
+import useInitializeTableStateFromCookies from '../../hooks/useInitializeTableStateFromCookies';
 import useRawSearchParams from '../../hooks/useRawSearchParams';
 import useSeedAdvancedColumnsFromUrl from '../../hooks/useSeedAdvancedColumnsFromUrl';
 import useTableFilters from '../../hooks/useTableFilters';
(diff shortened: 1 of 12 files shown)


src/components/CompareResults/ResultsTable.tsx
esanuandra, line 105: Should we consider the `framework`, `searchTerm` and `testVersion` as part of the filters and apply the changes here as well?

src/components/CompareResults/SubtestsResults/SubtestsResultsTable.tsx
esanuandra, line 1: `onSearchTermChange` from `SubtestsResultsMain.tsx` could also be updated to read the `currentUrlParams`

esanuandra (approved):
Lgtm
I didn't find other bugs, only the update in the subtests page

PR #1087, opened 2026-08-26
Bug-1995794: Disable retrigger dropdown when there are no runs

moijes12:
Disable the Base and New retrigger count selects independently when there are no retriggerable job ids for that side, and default a disabled select to 0 so it does not look like jobs will start.

@beatrice-acasandrei @kala-moz

### DIFF
--- a/src/__tests__/CompareResults/Retrigger.test.tsx
+++ b/src/__tests__/CompareResults/Retrigger.test.tsx
@@ -127,6 +127,25 @@ describe('Retrigger', () => {
     mockedGetLocationOrigin.mockImplementation(() => 'http://localhost:3000');
   });
 
+  async function openRetriggerConfigModal(
+    compareResult: typeof result = result,
+  ) {
+    setUpUserCredentials();
+    render(<RetriggerButton result={compareResult} variant='icon' />);
+
+    const openModalButton = await screen.findByRole('button', {
+      name: 'retrigger jobs',
+    });
+
+    const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime });
+    await user.click(openModalButton);
+
+    return {
+      baseSelect: await screen.findByLabelText('Base'),
+      newSelect: await screen.findByLabelText('New'),
+    };
+  }
+
   it('should display Sign In modal when there are no credentials', async () => {
     render(<RetriggerButton result={result} variant='icon' />);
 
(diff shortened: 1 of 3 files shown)


beatrice-acasandrei (commented):
In my opinion it would be helpful to gray out the dropdown text to clearly indicate that the field is disabled. @esanuandra @kala-moz What do you think?

Image

src/components/CompareResults/Retrigger/RetriggerConfigModal.tsx
beatrice-acasandrei: Consider disabling "Confirm" when both are disabled:
```suggestion
            <Button
              type='submit'
              disabled={!props.hasBaseJobs && !props.hasNewJobs}
            >
              {retriggerStrings.submitButton}
            </Button>
```

src/__tests__/CompareResults/Retrigger.test.tsx
```diff
@@ -267,6 +267,63 @@ describe('Retrigger', () => {
 
     expect(new MockedHooks().triggerHook).toHaveBeenCalled();
   });
+
+  async function openRetriggerConfigModal(
```
beatrice-acasandrei: Test helper should be at the top of the describe block before any test cases, not in the middle

src/components/CompareResults/Retrigger/RetriggerConfigModal.tsx
```diff
@@ -16,17 +16,20 @@ const retriggerStrings = Strings.components.retrigger.config;
...
 }) {
   return (
-    <FormControl sx={{ width: '100%' }}>
+    <FormControl sx={{ width: '100%' }} disabled={disabled}>
       <InputLabel id={`${prefix}-retrigger-count-label`}>{label}</InputLabel>
       <Select
         labelId={`${prefix}-retrigger-count-label`}
         name={`${prefix}-retrigger-count`}
-        defaultValue={5}
+        defaultValue={disabled ? 0 : 5}
```
beatrice-acasandrei, line 36: Could you add a brief explanatory comment here? It’s worth noting that using defaultValue is safe in this context because the modal unmounts when closed, ensuring the state re-initializes on each open.

src/components/CompareResults/Retrigger/RetriggerConfigModal.tsx
```diff
@@ -16,17 +16,20 @@ const retriggerStrings = Strings.components.retrigger.config;
...
+        disabled={disabled}
```
beatrice-acasandrei: FormControl automatically passes disabled down to its children through React context. So passing disabled explicitly to Select too is unnecessary. It's not a bug — it just creates extra noise and could confuse a reader into thinking it's needed for a reason.
```suggestion
```

moijes12:
@beatrice-acasandrei I have addressed the review comments in my latest commits

esanuandra:
I agree with @beatrice-acasandrei's suggestion to gray out the dropdown text to clearly indicate that the field is disabled.

PR #1093, opened 2026-09-01
Bug-2039589 - Add caching to ttest/mwu views

gopar:
Ticket: https://bugzilla.mozilla.org/show_bug.cgi?id=2039589


src/logic/treeherder.ts
```diff
@@ -309,7 +359,10 @@ export function memoizedFetchSubtestsCompareOverTimeResults(
   const key = JSON.stringify(params);
   const cached = subtestCompareOverTimeResultsCache.get(key);
   if (cached) return cached;
-  const promise = fetchSubtestsCompareOverTimeResults(params);
+  const promise = fetchSubtestsCompareOverTimeResults(params).catch((error) => {
+    subtestCompareOverTimeResultsCache.delete(key);
+    throw error;
+  });
```
gopar, line 365: Both of these were poisoning the cache when the network request fails, so simple fix is to remove key from cache when it does so

src/logic/treeherder.ts
```diff
@@ -278,6 +278,53 @@ export const memoizedFetchRevisionForRepository = moize(
   { isPromise: true, isShallowEqual: true, maxSize: 5 },
 ) as typeof fetchRevisionForRepository;
 
+// Memoized versions of the compare results fetch functions.
+// Going back and forth between the t-test and MWU views re-runs the route
+// loader, which would refetch the results from Treeherder every time.
+// We cache the value in memory which is keyed on the request parameters
```
beatrice-acasandrei: nit
```suggestion
// We cache the value in memory which is keyed on the request parameters.
// We intentionally do not cap the cache size: in a single session,
// the number of distinct revision combinations is small.
```

src/logic/treeherder.ts
```diff
@@ -309,11 +359,23 @@ export function memoizedFetchSubtestsCompareOverTimeResults(
...
   return promise;
 }
 
+// Clears all the memoization caches in this file. Only tests should need
+// this, to isolate the module-level caches between test cases.
+export function clearTreeherderCaches() {
+  compareResultsCache.clear();
+  compareOverTimeResultsCache.clear();
+  subtestCompareResultsCache.clear();
+  subtestCompareOverTimeResultsCache.clear();
```
beatrice-acasandrei, line 378: `clearTreeherderCaches` could also reset the moize cache
```suggestion
  subtestCompareOverTimeResultsCache.clear();
  memoizedFetchRevisionForRepository.clear();
```

kala-moz (approved):
LGTM; just rebase

